PR #1, opened 2015-03-12
MongoDB image

jhadvig:
Created the whole necessary structure for mongodb repo:
- images
- test suit
- documentation

The `make test` && `make test TARGET=rhel7` are passing so it should be safe.
@bparees @soltysh @mfojtik @mnagy PTAL :metal:

### DIFF
--- a/2.4/Dockerfile
+++ b/2.4/Dockerfile
@@ -1,0 +1,37 @@
+FROM centos:centos7
+
+# MongoDB image for OpenShift.
+#
+# Volumes:
+#  * /var/lib/mongodb/ - Datastore for MongoDB
+# Environment:
+#  * $MONGODB_USERNAME - Database user name
+#  * $MONGODB_PASSWORD - User's password
+#  * $MONGODB_DATABASE - Name of the database to create
+#  * $MONGODB_ADMIN_PASSWORD - Password of the MongoDB Admin
+
+# Image metadata
+ENV MONGODB_VERSION         2.4  
+ENV IMAGE_DESCRIPTION       MongoDB 2.4
+ENV IMAGE_TAGS              mongodb,mongodb24
+ENV IMAGE_EXPOSE_SERVICES   27017:mongodb
+
+MAINTAINER  Jakub Hadvig <jhadvig@redhat.com>
+EXPOSE 27017
+
+RUN yum install -y epel-release && \
+    rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
+    yum -y --setopt=tsflags=nodocs install \
+    https://www.softwarecollections.org/en/scls/rhscl/v8314/epel-7-x86_64/download/rhscl-v8314-epel-7-x86_64.noarch.rpm \
+    https://www.softwarecollections.org/en/scls/rhscl/mongodb24/epel-7-x86_64/download/rhscl-mongodb24-epel-7-x86_64.noarch.rpm && \
+    yum install -y --setopt=tsflags=nodocs v8314 mongodb24-mongodb mongodb24 && \
+    yum clean all
+
+COPY contrib /opt/openshift/
+ENV PATH /opt/openshift/bin:$PATH
+
+WORKDIR /opt/openshift
+
+VOLUME ["/var/lib/mongodb/"]
+
+ENTRYPOINT ["run.sh"]
--- a/2.4/contrib/bin/run.sh
+++ b/2.4/contrib/bin/run.sh
@@ -1,0 +1,91 @@
+#!/bin/bash
+
+# SCL in CentOS/RHEL 7 doesn't support --exec, we need to do it ourselves
+source scl_source enable mongodb24
+set -e
+
+MAX_ATTEMPTS=60
+SLEEP_TIME=1
+
+function usage() {
+	echo "You must specify following environment variables:"
+	echo "  \$MONGODB_USERNAME"
+	echo "  \$MONGODB_PASSWORD"
+	echo "  \$MONGODB_DB - optional (default database - 'production')"
+	echo "  \$MONGODB_ADMIN_PASSWORD - optional"
+	exit 1
+}
+
+function up_test() {
+    for i in $(seq $MAX_ATTEMPTS); do
+        echo "=> Waiting for confirmation of MongoDB service startup"
+        set +e
+        mongo admin --eval "help"
+        status=$?
+        set -e
+        if [ $status -eq 0 ]; then
+            echo "=> MongoDB service has started"
+            return 0
+        fi
+        sleep $SLEEP_TIME
+    done
+    echo "=> Giving up: Failed to start MongoDB service"
+    exit 1
+}
+
+function down_test() {
+    for i in $(seq $MAX_ATTEMPTS); do
+        echo "=> Waiting till MongoDB service is stopped"
+        set +e
+        mongo admin --eval "help"
+        status=$?
+        set -e
+        if [ $status -ne 0 ]; then
+            echo "=> MongoDB service has stopped"
+            return 0
+        fi
+        sleep $SLEEP_TIME
+    done
+    echo "=> Giving up: Failed to stop MongoDB service"
+    exit 1
+}
+
+function create_mongodb_users() {
+	# Start MongoDB service with disabled database authentication.
+	mongod -f /opt/openshift/etc/mongodb.conf &
+
+	# Check if the MongoDB daemon is up.
+	up_test
+
+	# Make sure env variables don't propagate to mongod process.
+	mongo_user="$MONGODB_USERNAME" ; unset MONGODB_USERNAME
+	mongo_pass="$MONGODB_PASSWORD" ; unset MONGODB_PASSWORD
+	mongo_db=${MONGODB_DATABASE:-"production"} ; unset MONGODB_DATABASE
+
+	# Create MongoDB database admin, if his password is specified with MONGODB_ADMIN_PASSWORD
+	if [ "$MONGODB_ADMIN_PASSWORD" ]; then
+		echo "=> Creating an admin user with a ${MONGODB_ADMIN_PASSWORD} password in MongoDB"
+		mongo admin --eval "db.addUser({user: 'admin', pwd: '$MONGODB_ADMIN_PASSWORD', roles: [ 'dbAdminAnyDatabase', 'userAdminAnyDatabase' , 'readWriteAnyDatabase' ]});"
+		unset MONGODB_ADMIN_PASSWORD
+	fi
+
+	# Create standard user with read/write permissions, in specified database ('production' by default).
+	mongo $mongo_db --eval "db.addUser({user: '${mongo_user}', pwd: '${mongo_pass}', roles: [ 'readWrite' ]});"
+	mongo admin --eval "db.shutdownServer();"
+
+	# Create a empty file which indicates that the database users were created.
+	touch /var/lib/mongodb/.mongodb_users_created
+
+	# Check if the MongoDB daemon is down.
+	down_test
+}
+
+test -z "$MONGODB_USERNAME" && usage
+test -z "$MONGODB_PASSWORD" && usage
+
+if [ ! -f /var/lib/mongodb/.mongodb_users_created ]; then
+	create_mongodb_users
+fi
+
+# Start MongoDB service with enabled database authentication.
+exec mongod -f /opt/openshift/etc/mongodb.conf --auth
--- a/2.4/contrib/etc/mongodb.conf
+++ b/2.4/contrib/etc/mongodb.conf
@@ -1,0 +1,21 @@
+# mongodb.conf
+
+port = 27017
+#fork = true
+pidfilepath = /opt/rh/mongodb24/root/var/run/mongodb/mongodb.pid
+dbpath = /var/lib/mongodb
+
+# Disable data file preallocation.
+noprealloc = true
+smallfiles = true
+
+journal = true
+
+#logpath = /var/lib/mongodb/mongodb.log
+# Verbose logging output.
+#verbose = true
+# Enables periodic logging of CPU utilization and I/O wait
+#cpu = false
+
+# Disable the HTTP interface (Defaults to localhost:28017).
+nohttpinterface = true
--- a/2.4/test/run
+++ b/2.4/test/run
@@ -1,0 +1,113 @@
+#!/bin/bash
+#
+# Test the MongoDB image.
+#
+# IMAGE_NAME specifies a name of the candidate image used for testing.
+# The image has to be available before this script is executed.
+#
+
+set -exo nounset
+shopt -s nullglob
+
+IMAGE_NAME=${IMAGE_NAME-openshift/mongodb-24-centos7-candidate}
+
+CIDFILE_DIR=$(mktemp --suffix=mongodb_test_cidfiles -d)
+
+cleanup() {
+    for cidfile in $CIDFILE_DIR/* ; do
+        CONTAINER=$(cat $cidfile)
+
+        echo "Stopping and removing container $CONTAINER..."
+        docker stop $CONTAINER
+        docker rm $CONTAINER
+        rm $cidfile
+        echo "Done."
+    done
+    rmdir $CIDFILE_DIR
+}
+trap cleanup EXIT SIGINT
+
+get_cid() {
+    local id="$1" ; shift || return 1
+    echo -n $(cat "$CIDFILE_DIR/$id")
+}
+
+get_container_ip() {
+    local id="$1" ; shift
+    docker inspect --format='{{.NetworkSettings.IPAddress}}' $(get_cid "$id")
+}
+
+mongo_cmd() {
+    docker run --rm --entrypoint=scl $IMAGE_NAME enable mongodb24 -- mongo "$DB" --host $CONTAINER_IP -u "$USER" -p "$PASS" --eval "${@}"
+}
+
+mongo_admin_cmd() {
+    docker run --rm --entrypoint=scl $IMAGE_NAME enable mongodb24 -- mongo admin --host $CONTAINER_IP -u admin -p "$ADMIN_PASS" --eval "${@}"
+}
+
+test_connection() { 
+    local name=$1 ; shift
+    ip=$(get_container_ip $name)
+    echo "  Testing MongoDB connection to $ip..."
+    local max_attempts=20
+    local sleep_time=2
+    for i in $(seq $max_attempts); do
+        echo "    Trying to connect..."
+        set +e
+        mongo_cmd "db.getSiblingDB('test_database');"
+        status=$?
+        set -e
+        if [ $status -eq 0 ]; then
+            echo "  Success!"
+            return 0
+        fi
+        sleep $sleep_time
+    done
+    echo "  Giving up: Failed to connect. Logs:"
+    docker logs $(get_cid $name)
+    return 1
+}
+
+test_mongo() {
+    echo "  Testing MongoDB"
+    if [ -v ADMIN_PASS ]; then
+        echo "  Testing Admin user privileges"
+        mongo_admin_cmd "db=db.getSiblingDB('${DB}'); db.removeUser('${USER}');"
+        mongo_admin_cmd "db=db.getSiblingDB('${DB}'); db.addUser({user: '${USER}', pwd: '${PASS}', roles: ['readWrite', 'userAdmin', 'dbAdmin']});"
+        mongo_admin_cmd "db=db.getSiblingDB('${DB}'); db.testData.insert({ x : 0 });"
+        mongo_cmd "db.addUser({user: 'test_user2', pwd: 'test_password2', roles: ['readWrite']});"
+    fi
+    echo "  Testing user privileges"
+    mongo_cmd "db.testData.insert({ y : 1 });"
+    mongo_cmd "db.testData.insert({ z : 2 });"
+    mongo_cmd "db.testData.find().forEach(printjson);"
+    mongo_cmd "db.testData.count();"
+    mongo_cmd "db.testData.drop();"
+    mongo_cmd "db.dropDatabase();"
+    echo "  Success!"
+}
+
+create_container() {
+    local name=$1 ; shift
+    cidfile="$CIDFILE_DIR/$name"
+    # create container with a cidfile in a directory for cleanup
+    docker run --cidfile $cidfile -d "$@" $IMAGE_NAME
+    echo "Created container $(cat $cidfile)"
+}
+
+run_tests() {
+    local name=$1 ; shift
+    envs="-e MONGODB_USERNAME=$USER -e MONGODB_PASSWORD=$PASS -e MONGODB_DATABASE=$DB"
+    if [ -v ADMIN_PASS ]; then
+        envs="$envs -e MONGODB_ADMIN_PASSWORD=$ADMIN_PASS"
+    fi
+    create_container $name $envs
+    CONTAINER_IP=$(get_container_ip $name)
+    test_connection $name
+    test_mongo $name
+}
+
+# Tests.
+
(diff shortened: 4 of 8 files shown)


2.4/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+RUN yum install -y epel-release && \
+    rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
+    yum -y --setopt=tsflags=nodocs install \
+    https://www.softwarecollections.org/en/scls/rhscl/v8314/epel-7-x86_64/download/rhscl-v8314-epel-7-x86_64.noarch.rpm \
+    https://www.softwarecollections.org/en/scls/rhscl/mongodb24/epel-7-x86_64/download/rhscl-mongodb24-epel-7-x86_64.noarch.rpm && \
+    yum install -y --setopt=tsflags=nodocs v8314 mongodb24-mongodb mongodb24 && \
+    yum clean all
+
+COPY contrib/etc/ /opt/openshift/etc
```
soltysh: Since @mnagy already set some conventions around it, I'd prefer that dir be called `mongo`, see https://github.com/openshift/mysql/blob/master/5.5/Dockerfile#L29

README.md
```diff
@@ -1 +1,138 @@
...
+
+|  Volume mount point    | Description            |
+| :--------------------- | ---------------------- |
+|  `/var/lib/mongodb/`   | MongoDB data directory |
+
+
+### Usage
+
+We will assume that you are using the `openshift/mongodb-24-centos7`
+image. Supposing that you want to set only mandatory required environment
```
soltysh: s/Supposing that you want/Suppose you want

README.md
```diff
@@ -1 +1,138 @@
...
+variables and store the database on in the `/home/user/database`
```
soltysh: s/database on in the/database in the

README.md
```diff
@@ -1 +1,138 @@
...
+command:
+
+```console
+docker run -d -e MONGODB_USER=<user> -e MONGODB_PASSWORD=<password> -e MONGODB_DATABASE=<database> -v /home/user/database:/var/lib/mongodb openshift/mongodb-24-centos7
+```
+
+If you are initializing the database and it's the first time you are using the
+specified shared volume, the database will be created, together with database
+administrator user and also MongoDB root user if `MONGODB_ADMIN_PASSWORD`
+environment variable is specified. after that the MongoDB daemon will be
```
soltysh: You should either uppercase after or change the dot into comma and don't forget to put a dot at the end of this statement (after started).

README.md
```diff
@@ -1 +1,138 @@
...
+started
+If you are re-attaching the volume to another container the creation of the
+database admin user and the root user will be skipped and only the mongodb
+daemon will be started
```
soltysh: Missing dot.

README.md
```diff
@@ -1 +1,138 @@
...
+
+
+### MongoDB root user
+The root user in not set by default. You can create him by setting
```
soltysh: s/in/is
s/create him/create one

README.md
```diff
@@ -1 +1,138 @@
...
+The root user in not set by default. You can create him by setting 
+`MONGODB_ROOT_PASSWORD` environment variable, in which case the root 
+user name will be set to `admin`. This process is done upon initializing 
+the database.
+
+
+## Software Collections
+We use [Software Collections](https://www.softwarecollections.org/) to
+install and launch MongoDB. If you want to execute a command inside of a
+running container (for debugging for example), you need to prefix it
```
soltysh: s/for debugging for example/eg. for debugging

README.md
```diff
@@ -1 +1,138 @@
...
+with `scl enable` command. Some examples:
```
soltysh: s/scl enable/scl enable mongodb24

README.md
```diff
@@ -1 +1,138 @@
...
+scl enable mongodb24 -- mongo <db_name> -u <username> -p <password>
+
+# Executing a command inside a running container from host
+# Note: You will be able to run mongodb commands without invoking the scl commands
+docker exec -ti <CONTAINER_ID> scl enable mongodb24 /bin/bash
+```
+
+## Test
+
+This repository also provides test framework, which launches tests to check functionality of the MongoDB image.
```
soltysh: Suggestion:
This repository also provides test framework, which checks basic functionality of the MongoDB image.

README.md
```diff
@@ -1 +1,138 @@
...
+
+*  **CentOS based image**
+
+    ```
+    $ cd mongodb
+    $ make test VERSION=2.4
+    ```
+
+**Notice: By omitting the `VERSION` parameter, the build/test action will be performed
+on all the supported versions of MongoDB. Since we are now supporting only version `2.4`
```
soltysh: s/on all the/ on all of the
s/Since we are now supporting only/Since we are only supporting

2.4/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+# MongoDB image for OpenShift.
+#
+# Volumes:
+#  * /var/lib/mongodb/ - Datastore for MongoDB
+# Environment:
+#  * $MONGODB_USER - Database user name
+#  * $MONGODB_PASSWORD - User's password
+#  * $MONGODB_DATABASE - Name of the database to create
+
```
bparees, line 12: ADMIN password?

2.4/contrib/bin/run.sh
```diff
@@ -0,0 +1,55 @@
...
+	fi
+
+	mongo $mongo_db --eval "db.addUser({user: '${mongo_user}', pwd: '${mongo_pass}', roles: [ 'readWrite', 'dbAdmin', 'userAdmin' ]});"
+	mongo admin --eval "db.shutdownServer();"
+
+	# Create a empty file which indicates that the database users were created.
+	touch /var/lib/mongodb/.mongodb_users_created
+
+	# Sleep for couple of seconds before the server daemon is started.
+	sleep 3
```
bparees: why?

2.4/contrib/etc/mongodb.conf
```diff
@@ -0,0 +1,21 @@
+# mongodb.conf
+
+port = 27017
+#fork = true
+pidfilepath = /opt/rh/mongodb24/root/var/run/mongodb/mongodb.pid
+dbpath = /var/lib/mongodb
+
+# Disable data file preallocation.
+noprealloc = true
```
bparees, line 9: should any of this be configurable via environment variable?  (seems like most of it should be)

2.4/test/run
```diff
@@ -0,0 +1,102 @@
...
+    # create container with a cidfile in a directory for cleanup
+    docker run --cidfile $cidfile -d "$@" $IMAGE_NAME
+    echo "Created container $(cat $cidfile)"
+}
+
+run_tests() {
+    local name=$1 ; shift
+    envs="-e MONGODB_USER=$USER -e MONGODB_PASSWORD=$PASS -e MONGODB_DATABASE=$DB"
+    if [ -v ROOT_PASS ]; then
+        envs="$envs -e MONGODB_ROOT_PASSWORD=$ROOT_PASS"
```
bparees: mongodb_admin_password?

also i don't see anything actually validating the admin account.

2.4/contrib/bin/run.sh
jhadvig: It take like second or two, to shutdown the mongodb daemon. Time to time I hit a problem that starting the deamon on the line 55 failed, cause the daemon was still running. After I put the sleep there I didnt hit the issue once.

2.4/contrib/etc/mongodb.conf
jhadvig, line 9: I've set the same parameter that we are using in v2 as a starting point. What I thought about is, if there is some way how could be supply apps with user custom config, so we could avoid creating and documenting tons of env vars.

2.4/test/run
jhadvig: Will update and create a separate test for the admin. :+1:

2.4/contrib/bin/run.sh
soltysh: On your machine it was 3 sec, on others it might be different. I'd suggest using some wait, where you'd check if daemon is running every couple of secs and after max timeout just fail with information about problem starting the daemon. Something similar to our `wait_for_command` function in origin [tests](https://github.com/openshift/origin/blob/master/hack/util.sh#L15).

2.4/Dockerfile
jhadvig, line 12: oops :anguished:

2.4/contrib/bin/run.sh
jhadvig: @soltysh +1 will check it out. Thanks :-)

bparees: @soltysh +1, that's where i was headed w/ my question :)

2.4/contrib/etc/mongodb.conf
bparees, line 9: since the DB images aren't STI images (coughcoughcough) i'm not sure how we'd enable apps to supply the DB config files.  Maybe some volume mount mechanism but i don't have that full picture in my head.

README.md
```diff
@@ -1 +1,138 @@
...
+	```
+
+*  **CentOS7 based image**
+
+	```console
+	git clone https://github.com/openshift/mongodb.git
+	cd mongodb
+	make build
+	```
+
```
bparees, line 49: ... or just pull it from dockerhub.....

2.4/contrib/etc/mongodb.conf
jhadvig, line 9: yeah the cough is in the right place here :P
Will think about possible solutions in here and maybe it would be a good topic for arch meeting. It would be certainly beneficial if users could supply the config files for the DBs.

bparees, line 9: good idea, let's queue it up for arch, right after @soltysh 's build notification topic.

jhadvig:
@soltysh @bparees @mfojtik  I've updated the PR. The substitution of the config files its still missing. Will deliver it  when we figure out with @mnagy the proper solution. In the meantime I dont wanna block this PR, since the substitution is not even part of the MySQL image.
Thoughts ?

2.4/contrib/bin/run.sh
```diff
@@ -0,0 +1,66 @@
...
+}
+
+function create_mongodb_users {
+	mongod -f /opt/openshift/etc/mongodb.conf &
+
+	# Check if the MongoDB daemon is up.
+	RC=1
+	while [[ RC -ne 0 ]]; do
+	    echo "=> Waiting for confirmation of MongoDB service startup"
+	    sleep 2
```
mfojtik: is 2 seconds really necessary? hw about 0.5?

2.4/test/run
```diff
@@ -0,0 +1,113 @@
...
+
+        echo "Stopping and removing container $CONTAINER..."
+        docker stop $CONTAINER
+        docker rm $CONTAINER
+        rm $cidfile
+        echo "Done."
+    done
+    rmdir $CIDFILE_DIR
+}
+trap cleanup EXIT
```
mfojtik: SIGINT

2.4/test/run
```diff
@@ -0,0 +1,113 @@
...
+
+get_cid () {
```
mfojtik: nuke space

2.4/test/run
```diff
@@ -0,0 +1,113 @@
...
+    local id="$1" ; shift || return 1
+    echo $(cat "$CIDFILE_DIR/$id")
```
mfojtik: echo -n

2.4/contrib/bin/run.sh
```diff
@@ -0,0 +1,66 @@
...
+
+	# Create standard user with read/write permissions, in specified database ('production' by default).
+	mongo $mongo_db --eval "db.addUser({user: '${mongo_user}', pwd: '${mongo_pass}', roles: [ 'readWrite' ]});"
+	mongo admin --eval "db.shutdownServer();"
+
+	# Create a empty file which indicates that the database users were created.
+	touch /var/lib/mongodb/.mongodb_users_created
+
+	# Check if the MongoDB daemon is down.
+	while [[ RC -eq 0 ]]
```
mfojtik: add retries (look at test_connection), if this break it will run forever

2.4/test/run
```diff
@@ -0,0 +1,113 @@
...
+    mongo_cmd "db.testData.drop();"
+    mongo_cmd "db.dropDatabase();"
+    echo "  Success!"
+}
+
+create_container() {
+    local name=$1 ; shift
+    cidfile="$CIDFILE_DIR/$name"
+    # create container with a cidfile in a directory for cleanup
+    docker run --cidfile $cidfile -d "$@" $IMAGE_NAME
```
mfojtik, line 94: docker run --rm

2.4/test/run
```diff
@@ -0,0 +1,113 @@
...
+    echo "Created container $(cat $cidfile)"
+}
+
+run_tests() {
+    local name=$1 ; shift
+    envs="-e MONGODB_USER=$USER -e MONGODB_PASSWORD=$PASS -e MONGODB_DATABASE=$DB"
+    if [ -v ADMIN_PASS ]; then
+        envs="$envs -e MONGODB_ADMIN_PASSWORD=$ADMIN_PASS"
```
mfojtik, line 102: env+=""

README.md
```diff
@@ -1 +1,146 @@
...
+### Usage
+
+We will assume that you are using the `openshift/mongodb-24-centos7`
+image. Suppose that you want to set only mandatory required environment
+variables and store the database in the `/home/user/database`
+directory on the host filesystem, you need to execute the following
+command:
+
+```console
+docker run -d -e MONGODB_USER=<user> -e MONGODB_PASSWORD=<password> -e MONGODB_DATABASE=<database> -v /home/user/database:/var/lib/mongodb openshift/mongodb-24-centos7
```
bparees: if you're only setting the mandatory params (as described on line 78), you don't need to set mongodb_database since you made that one optional.

README.md
```diff
@@ -1 +1,146 @@
 # MongoDB for OpenShift - Docker images
+
+This repository contains Dockerfiles for MongoDB images for OpenShift.
+Users can choose between RHEL and CentOS based images.
+
+### Versions
+
+MongoDB versions currently supported are:
```
mfojtik: Currently supported version of MongoDB

README.md
```diff
@@ -1 +1,146 @@
...
+docker run -d -e MONGODB_USER=<user> -e MONGODB_PASSWORD=<password> -e MONGODB_DATABASE=<database> -v /home/user/database:/var/lib/mongodb openshift/mongodb-24-centos7
+```
+
+If you are initializing the database and it's the first time you are using the
+specified shared volume, the database will be created, together with database
+administrator user and also MongoDB admin user if `MONGODB_ADMIN_PASSWORD`
+environment variable is specified. After that the MongoDB daemon will be
+started.
+If you are re-attaching the volume to another container the creation of the
+database admin user and the admin user will be skipped and only the mongodb
```
bparees: "database user and the admin user"

README.md
mfojtik: Same below

README.md
```diff
@@ -1 +1,146 @@
...
+scl enable mongodb24 -- mongo <db_name> -u <username> -p <password>
+
+# Executing a command inside a running container from host
+# Note: You will be able to run mongodb commands without invoking the scl commands
+docker exec -ti <CONTAINER_ID> scl enable mongodb24 /bin/bash
+```
+
+## Test
+
+This repository also provides test framework, which checks basic functionality of the MongoDB image.
```
bparees, line 119: _a_ test framework

README.md
```diff
@@ -1 +1,146 @@
...
+
+### Environment variables
+
+The image recognizes following environment variables that you can set
+during initialization, by passing `-e VAR=VALUE` to the Docker run
+command.
+
+|    Variable name          |    Description                              |   Default  |
+| :------------------------ | -----------------------------------------   | ---------- |
+|  `MONGODB_USER`           | User name for MONGODB account to be created |
```
mfojtik: Username

README.md
```diff
@@ -1 +1,146 @@
...
+|  `MONGODB_PASSWORD`       | Password for the user account               |
+|  `MONGODB_DATABASE`       | Database name (optional)                    | production |
+|  `MONGODB_ADMIN_PASSWORD` | Password for the admin user (optional)      |
+
+You can also set following mount points by passing `-v /host:/container`
+flag to docker.
```
mfojtik: Docker

README.md
```diff
@@ -1 +1,146 @@
...
+
+## Test
+
+This repository also provides test framework, which checks basic functionality of the MongoDB image.
+
+User can choose between testing MongoDB based on RHEL or CentOS image.
+
+*  **RHEL based image**
+
+    This image is not available as automated build in [Docker Index](https://index.docker.io).
```
bparees: strike this?

README.md
```diff
@@ -1 +1,146 @@
...
+	cd mongodb
+	make build TARGET=rhel7
+	```
+
+*  **CentOS7 based image**
+
+	This image is also available on DockerHub. To download it use:
+
+	```console
+	docker pull openshift/mongodb
```
mfojtik: docker pull openshift/mongodb-24-centos7

README.md
```diff
@@ -1 +1,146 @@
...
+
+
+### MongoDB admin user
+The admin user is not set by default. You can create one by setting
+`MONGODB_ADMIN_PASSWORD` environment variable, in which case the admin 
+user name will be set to `admin`. This process is done upon initializing 
+the database.
+
+
+## Software Collections
```
mfojtik: Note about Software Collections

bparees:
couple minor nits, otherwise lgtm.

README.md
```diff
@@ -1 +1,146 @@
...
+We use [Software Collections](https://www.softwarecollections.org/) to
+install and launch MongoDB. If you want to execute a command inside of a
+running container (eg. for debugging), you need to prefix it
+with `scl enable mongodb24` command. Some examples:
+
+```console
+# Running mongodb commands inside the container
+scl enable mongodb24 -- mongo <db_name> -u <username> -p <password>
```
mfojtik, line 110: again, this should be already setuped for users and ideally they should never call this SCL commands.... we can think about filing up environment with X_SCL

mfojtik, line 110: add this to .profile or .bashrc

mfojtik:
Couple nits from me as well, otherwise LGTM, nice job @jhadvig !

2.4/contrib/bin/run.sh
bparees: I debated making the same comment, but ultimately given how this is intended to run, i'm not sure it makes sense to give up.  If we give up, k8s is just going to respawn the container repeatedly.  Unless we think that is going to help/cause success, it's better to just leave the one container running in a loop than to keep spawning new containers indefinitely.

bparees: whoops, my comment was intended for the earlier loop, waiting for mongo to start (thought you were commenting on that one).

Not clear to me what the implications of the daemon not shutting down would be, if we pressed on...

mfojtik: @bparees my concern is that if mongodb get stuck this loop will run forever... having some number of retries and then just exiting with >0 seems better that having endless loop

2.4/contrib/bin/run.sh
mfojtik: also same comment as below, if mongodb gets stuck and will not accept connections for some reason (even though it will show itself as running, this will be waiting forever to start (and the container might still be reporting as 'running').

2.4/contrib/etc/mongodb.conf
mfojtik, line 9: I do remember that this preallocation should be done otherwise something really bad will happen ;D (this comes from my v2 memories)

README.md
bparees, line 110: if you docker exec into a container using /bin/sh or /bin/bash does it actually get run as a login shell to process .profile/.bashrc?  I would think not (you'd have to specify -l).  so not sure how useful that is going to be.

2.4/contrib/bin/run.sh
bparees: this is where my comment about k8s retrying really applies.  I guess I can see the argument that someone seeing their container continuously getting restarted might lead them to investigate further where as it just "running" might not, but the idea that we'd keep relaunching the same container over and over again with no real expectation that things are going to change seems like a mistake.

README.md
mfojtik, line 110: @bparees I think either .bashrc or .profile is being processed even without login shell (afaik)... my thinking is to make this more user friendly for users that don't know what is SCL...

2.4/contrib/bin/run.sh
bparees: yeah i understand the concern, i'm just not sure having k8s get stuck in a loop forever doing container creation is "better"

mfojtik: @bparees yeah, to not get into endless loop, we should exit(1) after some reasonable amount of time... so k8s can schedule another container (on another node) that might succeed.

2.4/contrib/bin/run.sh
mfojtik: @bparees isn't it possible to configure amount of restarts? Also I do believe that 'restarting' the container will produce an event that might be consumed either by the user or by the platform provider.... if certain amount of this events are reached, the platform provider might act on that.... comparing this to having container 'stuck' in launching, I think failing is better choice.

README.md
bparees, line 110: sure, i agree with doing it if indeed it's always processed.

2.4/contrib/bin/run.sh
bparees: given there are no external dependencies i'm not sure why we'd have any hope we'd get different behavior if we relaunch it...  if we do, that's bad.  

also we do have health/liveness checks which (if properly used) would catch the fact that this container is not in fact healthy though it is running.

The main argument I can see for having a timeout on these loops is because otherwise a user might not be sure how long they should reasonably wait for it to start.  By us setting a timeout we're explicilty telling them "if it hasn't started yet, it's not going to".  I might be convinced by that argument.

mfojtik: @bparees you might get different behavior... let say your container was launched on crappy node with corrupted storage or problem in kernel? Then restarting container of different node fixes the problem.

liveness problem capture this, that is true... but then aren't you just delegating shooting container to k8s itself? (and it will try to restart it ;-)

@bparees so what user will do when it gets 'stuck' mongodb container? Should user act and shoot it? For example, if I deploying production app and the mongo gets stuck to whatever livenessProbe is configured (if it is configured) then my app will be broken till I realize that mongodb failed to start and I manually fix it.

bparees: @mfojtik i think your bad node/corrupt image scenario convinced me we should limit how long we retry.  But for the last case you describe, the liveness probe would shoot the container.

jhadvig: @bparees @mfojtik thank you for adding valuable comments on this issue. Will put the `timeout` to one minute. I guess that should be a plenty of time to start mongo. Thanks :+1:

2.4/test/run
jhadvig, line 94: @mfojtik the container with the mongo daemon is removed in the `cleanup()` function + cant do `--rm` when using docker daemon

2.4/test/run
jhadvig, line 102: @mfojtik is this really necessary ? I like the original one more, cause in that case we can avoid one empty space in the beginning of the `+=` solution. Thoughts ?

jhadvig:
@mfojtik @bparees thanks for comments. PTAL
Skipped the substitution of the config files and enabling the SCL, until we figure out the proper way how to handle this topics.

PR #2, opened 2015-03-18
Making Dockerfiles more consistant with other DB images

jhadvig:
The run script will now live in the root of mongodb-24 along with Dockerfile && and copying it into `/usr/local/bin` so we can avoid creating a new dir for holding the scripts and so avoid extending the PATH env var.
@mfojtik @mnagy PTAL


mnagy:
LGTM :+1:

Issue #4, opened 2015-03-20
Database name should be required


mfojtik:
DATABASE_NAME needs to be required because users can't make assumptions about 'default' database name. If we make them to specify the database name, then we don't need them to point to documentation for the default one.

mfojtik:
@jhadvig fyi

jhadvig:
@mfojtik the default name is `production`.  But if we decide to go this path Im fine with it.

PR #5, opened 2015-03-20
Mandatory MONGODB_DATABASE env var

jhadvig:
Fixing https://github.com/openshift/mongodb/issues/4 issue.
Now specifying the DB name is mandatory.

@mfojtik @mnagy PTAL


mfojtik:
LGTM

PR #8, opened 2015-03-23
Fix mongodb daemon shutdown

jhadvig:
Shutting down mongodb daemon more clearly

### DIFF
--- a/2.4/run-mongod.sh
+++ b/2.4/run-mongod.sh
@@ -7,6 +7,8 @@ set -e
 MAX_ATTEMPTS=60
 SLEEP_TIME=1
 
+MONGODB_CONFIG_PATH="/opt/openshift/etc/mongodb.conf"
+
 function usage() {
 	echo "You must specify following environment variables:"
 	echo "  \$MONGODB_USERNAME"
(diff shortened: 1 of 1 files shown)


2.4/run-mongod.sh
```diff
@@ -71,7 +71,7 @@ function create_mongodb_users() {
 
 	# Create standard user with read/write permissions, in specified database ('production' by default).
 	mongo $mongo_db --eval "db.addUser({user: '${mongo_user}', pwd: '${mongo_pass}', roles: [ 'readWrite' ]});"
-	mongo admin --eval "db.shutdownServer();"
+	mongod -f /opt/openshift/etc/mongodb.conf --shutdown
```
mfojtik: make this env var. don't use paths directly

2.4/run-mongod.sh
```diff
@@ -7,6 +7,8 @@ set -e
 MAX_ATTEMPTS=60
 SLEEP_TIME=1
 
+MONGODB_CONFIG_PATH="/opt/openshift/etc/mongodb.conf"
```
mfojtik, line 10: `WORKDIR /opt/openshift`

this means that you can just do:

MONGODB_CONFIG_PATH="./etc/mongodb.conf"

jhadvig, line 10: @mfojtik imo that would only confuse some users. And if somebody would want to change the path of the config file, he would need to do it even in Dockerfile

mfojtik, line 10: k

PR #9, opened 2015-03-25
Adding various changes

jhadvig:
- Move DB volume into /var/lib/mogodb/data
- Add SCL enablement into /var/lib/mysql/.bashr so we can now run the
  arbitrary commands in the image without needing to override the
  entrypoint
- Making Dockerfiles more consistent with the rest of DB images

### DIFF
--- a/2.4/Dockerfile
+++ b/2.4/Dockerfile
@@ -1,9 +1,10 @@
 FROM centos:centos7
+MAINTAINER  Jakub Hadvig <jhadvig@redhat.com>
 
 # MongoDB image for OpenShift.
 #
 # Volumes:
-#  * /var/lib/mongodb/ - Datastore for MongoDB
+#  * /var/lib/mongodb/data - Datastore for MongoDB
 # Environment:
 #  * $MONGODB_USERNAME - Database user name
 #  * $MONGODB_PASSWORD - User's password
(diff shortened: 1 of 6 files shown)


jhadvig:
@mnagy, @mfojtik PTAL

2.4/Dockerfile
```diff
@@ -16,22 +17,28 @@ ENV IMAGE_DESCRIPTION       MongoDB 2.4
 ENV IMAGE_TAGS              mongodb,mongodb24
 ENV IMAGE_EXPOSE_SERVICES   27017:mongodb
 
-MAINTAINER  Jakub Hadvig <jhadvig@redhat.com>
 EXPOSE 27017
 
+# This image must forever use UID 184 for mongodb user so our volumes are
+# safe in the future. This should *never* change, the last test is there
+# to make sure of that.
```
mnagy: Maybe also mention gid 998?

"..forever use UID 184 and GID 998.."

2.4/Dockerfile
```diff
@@ -1,9 +1,10 @@
 FROM centos:centos7
+MAINTAINER  Jakub Hadvig <jhadvig@redhat.com>
```
mfojtik, line 2: :+1:

2.4/contrib/etc/mongodb.conf
```diff
@@ -3,7 +3,7 @@
 port = 27017
 #fork = true
 pidfilepath = /opt/rh/mongodb24/root/var/run/mongodb/mongodb.pid
-dbpath = /var/lib/mongodb
+dbpath = /var/lib/mongodb/data
```
mfojtik, line 6: this should be captured by the envsubst that @mnagy is doing

mfojtik:
LGTM

2.4/contrib/etc/mongodb.conf
jhadvig, line 6: And it will be, once we update the image with that functionality

2.4/run-mongod.sh
```diff
@@ -59,36 +65,30 @@ function create_mongodb_users() {
...
 	# Check if the MongoDB daemon is down.
 	down_test
 }
 
-test -z "$MONGODB_USERNAME" && usage
-test -z "$MONGODB_PASSWORD" && usage
-test -z "$MONGODB_DATABASE" && usage
-
-if [ ! -f /var/lib/mongodb/.mongodb_users_created ]; then
+if [ "$1" = "mongod" -a ! -f /var/lib/mongodb/data/.mongodb_users_created ]; then
```
mnagy: Can you please do this similarly to this? https://github.com/mnagy/postgresql/commit/4a2b7d4ca437b09325175e7d4af1b378c4e1c34a
Since you want to run mongod with parameters, you can just add following line after the db initialization if:
`set -- mongod -f $MONGODB_CONFIG_PATH --auth "$@"`

jhadvig: @mnagy imagine if user would override the the `--dbpath`, in that case the running the container would fail. Like I understand your concern but it would introduce introduce more complexity which we dont need atm. So I would rather not go this path.

2.4/run-mongod.sh
```diff
@@ -59,36 +65,33 @@ function create_mongodb_users() {
...
 fi
 
-# Start MongoDB service with enabled database authentication.
-exec mongod -f $MONGODB_CONFIG_PATH --auth
+unset_env_vars
+
+if [ "$1" = "mongod" ]; then
+    # Start MongoDB service with enabled database authentication.
+    exec mongod -f $MONGODB_CONFIG_PATH --auth
+fi
```
mnagy, line 100: You have spaces here instead of tabs :)

mnagy:
LGTM

PR #14, opened 2015-04-07
WIP: Add MongoDB replica set example

mfojtik:
Don't merge this as I'm still trying to make the scripts hacks look nicer :-) There should be no/minimal changes to standalone MongoDB.

I created this PR mainly to get your feedback. Please see the README.md (especially the 'Known Issues' section).

### DIFF
--- a/2.4/contrib/common.sh
+++ b/2.4/contrib/common.sh
@@ -1,0 +1,93 @@
+#!/bin/bash
+
+# Used for wait_for_mongo_* functions
+MAX_ATTEMPTS=60
+SLEEP_TIME=1
+
+export CONTAINER_PORT=$(echo -n $IMAGE_EXPOSE_SERVICES | cut -d ':' -f 1)
+
+function container_addr() {
+  echo -n $(cat /var/lib/mongodb/.address)
+}
+
+function mongo_addr() {
+  echo -n "$(container_addr):${CONTAINER_PORT}"
+}
+
+function cache_container_addr() {
+  ip addr | grep 'state UP' -A2 | tail -n1 | awk '{print $2}' | cut -f1  -d'/' > /var/lib/mongodb/.address
+}
+
+# wait_for_mongo_up waits until the mongo server accepts incomming connections
+function wait_for_mongo_up() {
+  for i in $(seq $MAX_ATTEMPTS); do
+    echo "=> Waiting for confirmation of MongoDB service startup"
+    set +e
+    mongo admin --eval "help" &>/dev/null
+    status=$?
+    set -e
+    if [ $status -eq 0 ]; then
+      echo "=> MongoDB service has started"
+      return 0
+    fi
+    sleep $SLEEP_TIME
+  done
+  echo "=> Giving up: Failed to start MongoDB service"
+  exit 1
+}
+
+# wait_for_mongo_up waits until the mongo server is down
+function wait_for_mongo_down() {
+  for i in $(seq $MAX_ATTEMPTS); do
+    echo "=> Waiting till MongoDB service is stopped"
+    set +e
+    mongo admin --eval "help" &>/dev/null
+    status=$?
+    set -e
+    if [ $status -ne 0 ]; then
+      echo "=> MongoDB service has stopped"
+      return 0
+    fi
+    sleep $SLEEP_TIME
+  done
+  echo "=> Giving up: Failed to stop MongoDB service"
+  exit 1
+}
+
+# endpoints returns list of IP addresses with other instances of MongoDB
+# To get list of endpoints, you need to have headless Service named 'mongodb'.
+# NOTE: This won't work with standalone Docker container.
+function endpoints() {
+  dig mongodb A +search +short 2>/dev/null
+}
+
+# no_endpoints returns true if the only endpoint is the current container itself
+# or there are no endpoints registred (running as standalone Docker container)
+function no_endpoints() {
+  [ -z "$(endpoints)" ] && return 0
+  [ "$(endpoints)" == "$(container_addr)" ]
+}
+
+# deregister removes the current MongoDB from the cluster
+function deregister() {
+  for node in $(endpoints); do
+    echo "=> Removing $(mongo_addr) from replica set"
+    mongo admin --host ${node}:${CONTAINER_PORT} --eval "rs.remove(\"$(mongo_addr)\");"
+  done
+}
+
+# mongo_initiate initiate the replica set
+function mongo_initiate() {
+  local mongo_wait="while (rs.status().startupStatus || (rs.status().hasOwnProperty(\"myState\") && rs.status().myState != 1)) { printjson( rs.status() ); sleep(1000); }; printjson( rs.status() );"
+  local mongo_config="var config={ _id: \"${MONGODB_REPLICA_NAME}\", members: [ { _id: 0, host: \"$(mongo_addr)\" } ] };"
+  mongo admin --eval "${mongo_config};rs.initiate(config);${mongo_wait}"
+}
+
+# mongo_advertise advertise the current container to other mongo replicas
+function mongo_advertise() {
+    local node="$1"
+    # Skip the current container
+    [[ "${node}" == "$(container_addr)" ]] && continue
+    echo "=> Advertising new MongoDB node to ${node}"
+    mongo admin --host ${node}:${CONTAINER_PORT} --eval "rs.add(\"$(mongo_addr)\");"
+}
--- a/2.4/contrib/common.sh
+++ b/2.4/contrib/common.sh
@@ -15,7 +15,16 @@ function mongo_addr() {
 }
 
 function cache_container_addr() {
-  ip addr | grep 'state UP' -A2 | tail -n1 | awk '{print $2}' | cut -f1  -d'/' > /var/lib/mongodb/.address
+  echo "=> Waiting for container IP address ..."
+  for i in $(seq $MAX_ATTEMPTS); do
+    result=$(ip addr | grep 'state UP' -A2 | tail -n1 | awk '{print $2}' | cut -f1 -d'/')
+    if [ ! -z "${result}" ]; then
+      echo -n $result > /var/lib/mongodb/.address
+      echo "=> Got $(mongo_addr)"
+      return 0
+    fi
+    sleep $SLEEP_TIME
+  done
 }
 
 # wait_for_mongo_up waits until the mongo server accepts incomming connections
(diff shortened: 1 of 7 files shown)


mfojtik:
@soltysh @jhadvig @bparees @danmcp PTAL

2.4/examples/replica/mongo_controller.json
```diff
@@ -0,0 +1,64 @@
...
+            "name": "mongodb-node"
+          },
+          "desiredState": {
+            "manifest": {
+              "containers":[
+                {
+                  "name":  "mongo-server",
+                  "image": "openshift/mongodb-24-centos7:replica",
+                  "env": [
+                    { "name": "MONGODB_USERNAME",       "value": "user"  },
```
mfojtik: FYI: This is sorely ignored at this moment. I'll have to rework authentication completely to use 'keyfile' instead of passwords. I need to have this here just to pass the validation ;-)

2.4/run-mongod.sh
```diff
@@ -14,108 +15,96 @@ export MONGODB_DATADIR=/var/lib/mongodb/data
...
 export MONGODB_CONFIG_PATH=/var/lib/mongodb/mongodb.conf
-export MONGODB_PID_FILE=var/run/mongodb/mongodb.pid
-
-MAX_ATTEMPTS=60
-SLEEP_TIME=1
+export MONGODB_PID_FILE=/var/lib/mongodb/mongodb.pid
 
 function usage() {
-	echo "You must specify following environment variables:"
-	echo "  \$MONGODB_USERNAME"
```
mfojtik, line 26: @jhadvig action item for you.... remove the '$' prefix... it is highly unpractical when copy&pasting the variable in terminal ;-)

2.4/run-mongod.sh
```diff
@@ -14,108 +15,96 @@ export MONGODB_DATADIR=/var/lib/mongodb/data
...
+
+  mongo_cmd_args="-f $MONGODB_CONFIG_PATH --noprealloc --smallfiles --oplogSize 64"
+
+  if [ ! -z "${MONGODB_REPLICA_NAME-}" ]; then
+    mongo_cmd_args+=" --replSet ${MONGODB_REPLICA_NAME}"
+    ( mkfifo /tmp/replica_supervisor_log && cat /tmp/replica_supervisor_log ) &
+    /var/lib/mongodb/replica_supervisor.sh &
+    trap 'cleanup' SIGINT SIGTERM
+    mongod $mongo_cmd_args & mongo_pid=$!
+    wait $mongo_pid
```
mfojtik, line 105: before you started crucifying me here, this is needed to get the 'trap' working. In replicated setup, you need to de-register the current node from the set and this must be done when the container is shutting down.

2.4/contrib/common.sh
```diff
@@ -0,0 +1,93 @@
...
+  done
+  echo "=> Giving up: Failed to stop MongoDB service"
+  exit 1
+}
+
+# endpoints returns list of IP addresses with other instances of MongoDB
+# To get list of endpoints, you need to have headless Service named 'mongodb'.
+# NOTE: This won't work with standalone Docker container.
+function endpoints() {
+  dig mongodb A +search +short 2>/dev/null
```
bparees, line 70: i imagine we'll want this to be parameterizable ultimately.

2.4/examples/replica/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+## How does it work?
+
+The MongoDB replication controller is set to **1** in this example. Once the Pod
+with MongoDB is started, the replica set is initialized (`rs.initiate()`)
+and the Pod is elected as the 'master' (as there are no other MongoDB Pods
+running, yet).
+
+If you want to add new node to MongoDB replica set, bump the value of 'replicas'
+in the Kubernetes Replication Contoller to '2'.
```
bparees, line 35: does it work if i start with 2+?

2.4/run-mongod.sh
bparees, line 105: how long does this take?  doesn't k8s/docker only give you about 10s to cleanly terminate before killing?

mfojtik, line 105: yeah, but that is enough, usually the deregister takes less than second (it just sent the request to all replicas)

2.4/examples/replica/README.md
mfojtik, line 35: Unfortunately no. I think it will create 2 MongoDB instances but none of them will initiate the replicaSet (as they will both have endpoints). In case k8s will be slow, both will not have endpoints and in that case both will initiate the replicaSet but they will not see each other.

The solution for this can be creating an initial Pod (outside ReplicationController) that will serve as the first MongoDB node and so it will initiate the replicaSet. Then the ReplicationController will be created with replicas set to 0 and when you bump the number they will serve as 'replica' nodes.

mfojtik, line 35: @danmcp is the above reasonable?

2.4/examples/replica/mongo_controller.json
mfojtik: @bparees @danmcp I think `keyfile` is reasonable here. I can use Secret to distribute it

2.4/contrib/common.sh
```diff
@@ -0,0 +1,103 @@
...
+function mongo_initiate() {
+  local mongo_wait="while (rs.status().startupStatus || (rs.status().hasOwnProperty(\"myState\") && rs.status().myState != 1)) { printjson( rs.status() ); sleep(1000); }; printjson( rs.status() );"
+  local mongo_config="var config={ _id: \"${MONGODB_REPLICA_NAME}\", members: [ { _id: 0, host: \"$(mongo_addr)\" } ] };"
+  echo "=> Initiating MongoDB replica using: ${mongo_config}"
+  mongo admin --eval "${mongo_config};rs.initiate(config);${mongo_wait}"
+}
+
+# mongo_advertise advertise the current container to other mongo replicas
+function mongo_advertise() {
+    local node="$1"
```
jhadvig: Why different sizes of indentation ? 2 vs. 4 ?

2.4/run-mongod.sh
jhadvig, line 26: @mfojtik you mean from other images ? Cause I see you already did it below :)

2.4/run-mongod.sh
```diff
@@ -14,108 +15,95 @@ export MONGODB_DATADIR=/var/lib/mongodb/data
...
 function unset_env_vars() {
-	unset MONGODB_USERNAME MONGODB_PASSWORD MONGODB_DATABASE MONGODB_ADMIN_PASSWORD
+  unset MONGODB_USERNAME MONGODB_PASSWORD MONGODB_DATABASE MONGODB_ADMIN_PASSWORD
 }
 
 function create_mongodb_users() {
-	# Start MongoDB service with disabled database authentication.
-	mongod -f $MONGODB_CONFIG_PATH &
+  # Start MongoDB service with disabled database authentication.
+  mongod -f $MONGODB_CONFIG_PATH --noprealloc --smallfiles --oplogSize 64 &
```
jhadvig, line 43: Whats the point it adding `--noprealloc --smallfiles` since they are defined in config file ?

2.4/contrib/common.sh
mfojtik: just a typo, good catch

2.4/run-mongod.sh
mfojtik, line 26: @jhadvig yeah, I did it because it was annoying :-)

2.4/run-mongod.sh
mfojtik, line 43: good point, I haven't seen them and without these options the mongo grows quite big ;-)

2.4/examples/replica/mongo_controller.json
bparees: i'd vote fore allowing both.  setting env variables is a lower barrier to entry for users than creating a keyfile+secret.  (assuming i'm understanding the options correctly)

2.4/contrib/common.sh
mfojtik, line 70: yeah, you mean the service name?

bparees, line 70: yes

2.4/examples/replica/README.md
bparees, line 35: i wonder if we can do better using the deployment hooks that dan mace is working on to just a hook in exactly one pod.

2.4/contrib/common.sh
mfojtik, line 70: @bparees it will have to be parsed from the `env` command, good catch.

2.4/examples/replica/README.md
danmcp, line 35: @mfojtik Not really :(  Or at least I would like to do better.

mfojtik, line 35: @danmcp the challenge here is that you have initialize the replicaSet on one of the Mongo nodes. If you start with 3 replicas (RC replicas=3) then all three pods will start in parallel and you don't know on what pod you should run the rs.initiate().

I'm not the mongo guru here but my theory is that you can create 3 Pods and then pick one (perhaps the first IP returned from dig mongodb A +short +search) and send the rs.initiate(). I'm not sure if this will work and if you can do rs.add() if the replicaSet is not initialized... I can do more research on this tomorrow.

UPDATE: The problem I see here is that which Pod should call the rs.initiate(). Given that all Pods are equal, then there might be a race condition as all Pods will call the rs.initiate() on first endpoint in parallel, which might break initialization of replSet.

@danmcp will the above be better?

danmcp, line 35: @mfojtik I've chatted with @bparees and @smarterclayton on this a little.  I think the consensus is we want to deploy a membership management pod to handle this.  You can either reach out to the member management from mongo or reach into mongo through its api from the member management pod.  The latter we think is possible with mongo but may not be will all frameworks.  But the idea is you run a secondary service that will let you elect (or do the election yourself) an initializer.

Ideally we would be able to make this pattern reusable.

mfojtik, line 35: @danmcp this is kind of what I'm currently doing. The Pod I launch first with 'role=seeder' acts as the management Pod. This Pod initialize the replicaSet and then the other member can register to this Pod. If this Pod goes away, other members (replicas) elect a new master.

I think, MongoDB call what you describe 'config server'. This is mentioned also in Kubernetes issue related to MongoDB replication. The 'config server' acts as the member management Pod (can also be scaled AFAIK).

danmcp, line 35: @mfojtik Config servers would be for sharding scenarios right?  Not replica sets.  We should probably talk this through at this point.

mfojtik, line 35: @danmcp I created another example that use the 'member management' Pod as you described:

https://github.com/mfojtik/mongodb/tree/scale_management/2.4/examples/replica

See the README.md there. It is working perfectly, removing the members works also. The only thing that remains opened is authentication (keyfile).

mfojtik:
Closing this, see: https://github.com/openshift/mongodb/pull/17

PR #21, opened 2015-04-22
Add iproute for 'ip' command

sdodson:
It's included in centos base image but not rhel7. Add it to both as it may be
removed from centos in the future if the two images converge. I've also added bind-utils to the rhel7 image to keep the two as similar as possible.


bparees:
@jhadvig fyi.

PR #22, opened 2015-04-22
Support random uid at runtime

csrwng:
Makes it possible to run mongodb with arbitrary user
Example:

```
docker run -e MONGODB_USER=user -e MONGODB_PASSWORD=pass \
   -e MONGODB_DATABASE=db1 -u 12345 -p 27017:27017 \
   openshift/mongodb-24-centos7
```

### DIFF
--- a/2.4/Dockerfile
+++ b/2.4/Dockerfile
@@ -12,7 +12,7 @@ MAINTAINER  Jakub Hadvig <jhadvig@redhat.com>
 #  * $MONGODB_ADMIN_PASSWORD - Password of the MongoDB Admin
 
 # Image metadata
-ENV MONGODB_VERSION         2.4  
+ENV MONGODB_VERSION         2.4
 ENV IMAGE_DESCRIPTION       MongoDB 2.4
 ENV IMAGE_TAGS              mongodb,mongodb24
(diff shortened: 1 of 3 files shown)


csrwng:
@jhadvig @mnagy pls take a look

mnagy:
Can you add a test case please?

2.4/run-mongod.sh
```diff
@@ -6,6 +6,8 @@ source /var/lib/mongodb/.bashrc
 
 set -eu
 
+export HOME=/var/lib/mongodb
```
jhadvig: So this would be right approach, unfortunately there is a bug in mongodb24 scl and the pkg is setting wrong HOME dir for the mongodb user. Check https://bugzilla.redhat.com/show_bug.cgi?id=1206151 and I doubt they will fix it any time soon.
So Im wondering if this change couldnt break us somehow.

csrwng: I made this change because I'd made a similar change to the mysql image. However, after trying it out without the change to the script above, everything worked ok. So I've removed it. Updated the commit.

csrwng:
Added tests for a random uid

jhadvig:
LGTM :+1:

PR #26, opened 2015-05-12
squash on build

bparees:


bparees:
@goldmann fyi per our conversation this morning, i'm just doing it this way now.

hack/build.sh
```diff
@@ -9,6 +9,22 @@ declare -a VERSIONS=(2.4)
...
+	easy_install --user pip
+	rm -rf docker-scripts
+	git clone --branch 0.3.0 https://github.com/goldmann/docker-scripts/
+	pushd docker-scripts
+	~/.local/bin/pip install --force --user .
+	popd
+	# find the top of the "centos" or "rhel" layer
+	layer=$(~/.local/bin/docker-scripts layers -m ${base} | tail -n 1)
+	# squash everything above it
+	$HOME/.local/bin/docker-scripts squash -f $layer ${IMAGE_NAME}
```
goldmann: Using `$base` here does not give you the expected results? I'm pretty sure the `squash` command will properly find the layer ID. Can you compare the output of the `squash` logs (the "Attempting to squash from layer" line with the ID you get from `layers -m $base`? These should be really the same since I do this checking: https://github.com/goldmann/docker-scripts/blob/master/docker_scripts/squash.py#L347

hack/build.sh
```diff
@@ -9,6 +9,22 @@ declare -a VERSIONS=(2.4)
 OS=$1
 VERSION=$2
 
+function squash {	
+	base=$(grep ^FROM Dockerfile | awk '{print $2}')
```
goldmann: Small improvement: `awk '/^FROM/{print $2}' Dockerfile` :)

hack/build.sh
```diff
@@ -9,6 +9,22 @@ declare -a VERSIONS=(2.4)
...
+	# install the docker layer squashing tool
+	easy_install --user pip
```
goldmann: You can now use pip to install docker-scripts:

```
pip install --user docker-scripts==0.3.2
```

hack/build.sh
bparees: I should have read the help more closely, I thought -f had to be a layer id, didn't realize it could be an image name.

hack/build.sh
bparees: not convinced that's an improvement (readability wise), but ok. :)

hack/build.sh
goldmann: Yah, it could be really anything, I let Docker decode it and when it fails, this means that something is wrong with it.

hack/build.sh
bparees: $ ~/.local/bin/pip install --user docker-scripts=0.3.2
Exception:
Traceback (most recent call last):
  File "/root/.local/lib/python2.7/site-packages/pip-6.1.1-py2.7.egg/pip/basecommand.py", line 246, in main
    status = self.run(options, args)
  File "/root/.local/lib/python2.7/site-packages/pip-6.1.1-py2.7.egg/pip/commands/install.py", line 308, in run
    name, None, isolated=options.isolated_mode,
  File "/root/.local/lib/python2.7/site-packages/pip-6.1.1-py2.7.egg/pip/req/req_install.py", line 220, in from_line
    isolated=isolated)
  File "/root/.local/lib/python2.7/site-packages/pip-6.1.1-py2.7.egg/pip/req/req_install.py", line 79, in **init**
    req = pkg_resources.Requirement.parse(req)
  File "/root/.local/lib/python2.7/site-packages/pip-6.1.1-py2.7.egg/pip/_vendor/pkg_resources/__init__.py", line 2960, in parse
    reqs = list(parse_requirements(s))
  File "/root/.local/lib/python2.7/site-packages/pip-6.1.1-py2.7.egg/pip/_vendor/pkg_resources/__init__.py", line 2904, in parse_requirements
    "version spec")
  File "/root/.local/lib/python2.7/site-packages/pip-6.1.1-py2.7.egg/pip/_vendor/pkg_resources/__init__.py", line 2869, in scan_list
    raise ValueError(msg, line, "at", line[p:])
ValueError: ('Expected version spec in', 'docker-scripts=0.3.2', 'at', '=0.3.2')

goldmann: You need to use equal signs, not one :)

```
pip install --user docker-scripts==0.3.2
```

vs.

```
pip install --user docker-scripts=0.3.2
```

hack/build.sh
```diff
@@ -9,6 +9,16 @@ declare -a VERSIONS=(2.4)
 OS=$1
 VERSION=$2
 
+function squash { 
+  # install the docker layer squashing tool
+  easy_install --user pip
```
goldmann: In fact you don't even need to install `pip`, just `easy_install --user docker-scripts==0.3.2` directly. Will save another line + time.

I guess it's time to go to bed now :)

PR #29, opened 2015-05-19
use uid instead of username

bparees:


bparees:
@csrwng ptal.

csrwng:
LGTM

rhcarvalho:
Trying to understand the history, @bparees @csrwng do you remember why we made this change?

php-coder:
@rhcarvalho I suppose that it was for passing OpenShift check for uid range.

PR #30, opened 2015-05-21
Update mounted dir README

jhadvig:
Based on the https://bugzilla.redhat.com/show_bug.cgi?id=1214613 I update the README, so users can avoid hitting the same issue.


jhadvig:
@bparees @mfojtik PTAL
After review I will also add this to postresql and mysql

bparees:
lgtm, thanks!

PR #31, opened 2015-05-26
Mandatory admin user

jhadvig:
Making admin user mandatory, since once the mongodb is launched in authentication mode with `--auth` flag, there is no easy way how to create admin user retrospectively, after the DB was created without admin user.

### DIFF
--- a/2.4/run-mongod.sh
+++ b/2.4/run-mongod.sh
@@ -24,9 +24,9 @@ function usage() {
   echo "You must specify following environment variables:"
   echo "  MONGODB_USER"
   echo "  MONGODB_PASSWORD"
-  echo "Optional variables:"
-  echo "  MONGODB_DATABASE (default: \$MONGODB_USER)"
+  echo "  MONGODB_DATABASE"
   echo "  MONGODB_ADMIN_PASSWORD"
+  echo "Optional variables:"
   echo "  MONGODB_REPLICA_NAME"
   echo "MongoDB settings:"
   echo "  MONGODB_NOPREALLOC (default: true)"
@@ -55,7 +55,7 @@ function cleanup() {
 }
 
 if [ "$1" == "initiate" ]; then
-  if ! [[ -v MONGODB_USER && -v MONGODB_PASSWORD ]]; then
+  if ! [[ -v MONGODB_USER && -v MONGODB_PASSWORD && -v MONGODB_DATABASE && -v MONGODB_ADMIN_PASSWORD ]]; then
     usage
   fi
   setup_keyfile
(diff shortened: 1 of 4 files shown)


jhadvig:
@mnagy @mfojtik PTAL

soltysh:
@jhadvig don't forget to update https://github.com/openshift/openshift-docs

jhadvig:
@soltysh thats the first thing will do after its merged ! :stuck_out_tongue_winking_eye:

README.md
```diff
@@ -100,18 +100,18 @@ $ docker run -d -e MONGODB_USER=<user> -e MONGODB_PASSWORD=<password> -e MONGODB
...
 
 MongoDB admin user
 ---------------------------------
-The admin user is not set by default. You can create one by setting
-`MONGODB_ADMIN_PASSWORD` environment variable, in which case the admin user name
-will be set to `admin`. This process is done upon initializing the database.
+
+The admin user is set by default. Admin user name is set to `admin` and you
+have to to specify his password by setting `MONGODB_ADMIN_PASSWORD` environment 
+variable. This process is done upon initializing the database.
```
mnagy: Can you please remove this entire section and just put one sentence somewhere else in the readme about the admin username being `admin`? I think the first sentence is now redundant and saying the user needs to set the admin password is also not necessary, since it is known from the list of env vars.

jhadvig: I dont think removing the whole section would be appropriate, since we have the 'admin' section also in other db images and it would be good to have this section cause its a pretty crucial thing. But you are right that the first sentence is redundant, will remove it.

soltysh:
One small nit about missing Usage section update with `MONGODB_ADMIN_PASSWORD` in Readme and you can get LGTM from me.

jhadvig:
@soltysh updated

README.md
mnagy: Other images are different. Both MySQL and PostgreSQL allow local connection without a password if that password is not provided, with MongoDB, that is not the case.

2.4/run-mongod.sh
```diff
@@ -24,9 +24,9 @@ function usage() {
   echo "You must specify following environment variables:"
   echo "  MONGODB_USER"
   echo "  MONGODB_PASSWORD"
-  echo "Optional variables:"
-  echo "  MONGODB_DATABASE (default: \$MONGODB_USER)"
+  echo "  MONGODB_DATABASE"
```
mfojtik: we don't have any default now?

README.md
```diff
@@ -95,23 +95,23 @@ in the `/home/user/database` directory on the host filesystem, you need to
...
 
 MongoDB admin user
 ---------------------------------
-The admin user is not set by default. You can create one by setting
-`MONGODB_ADMIN_PASSWORD` environment variable, in which case the admin user name
-will be set to `admin`. This process is done upon initializing the database.
+
+Admin user name is set to `admin` and you have to to specify his password by
+setting `MONGODB_ADMIN_PASSWORD` environment variable. This process is done
+upon initializing the database.
```
mnagy: What process? Again, we only need to state the admin account name is `admin`, IMO. And I think having a separate section just for that one sentence is overkill.

mfojtik:
@jhadvig LGTM

README.md
mfojtik: upon database initialization.

soltysh:
LGTM

PR #36, opened 2015-06-02
Print message if database directory is already initialized.

omron93:
Container user is not notified if database directory is already initialized and new database users are not created. So now environmental variables in container for linking could be invalid and container user don't know about it.

For example when you run container with reattached data directory, you can specify different env variables (for example you don't know that data directory is already initialized) and you are not notified that database users were already initialized and database users specified by you weren't created.

### DIFF
--- a/2.4/run-mongod.sh
+++ b/2.4/run-mongod.sh
@@ -84,7 +84,7 @@ if [ "$1" = "mongod" ]; then
       mongod $mongo_common_args --shutdown
       wait_for_mongo_down
     else
-      echo "=> Database directory is already initialized. Skipping creation of users..."
+      echo "=> Database directory is already initialized. Skipping creation of users ..."
     fi
     unset_env_vars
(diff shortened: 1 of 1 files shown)


2.4/run-mongod.sh
```diff
@@ -83,6 +83,8 @@ if [ "$1" = "mongod" ]; then
       # Restart the MongoDB daemon to bind on all interfaces
       mongod $mongo_common_args --shutdown
       wait_for_mongo_down
+    else
+      echo "=> Database directory is already initialized. Skipping creation of users..."
```
mfojtik: Can you please put extra space before '...' ("Foo ...")? We are trying to make the output look consistent ;-)

mfojtik:
LGTM please fix the minor nit

mfojtik:
[test]

mfojtik:
@omron93 thanks!

PR #45, opened 2015-06-19
Remove keyfile volume

jhadvig:
Getting rid of the volume with the keyfile and generating the keyfile value in template instead.
@mfojtik PTAL

### DIFF
--- a/2.4/contrib/common.sh
+++ b/2.4/contrib/common.sh
@@ -6,6 +6,7 @@ SLEEP_TIME=1
 
 export MONGODB_CONFIG_PATH=/var/lib/mongodb/mongodb.conf
 export MONGODB_PID_FILE=/var/lib/mongodb/mongodb.pid
+export MONGODB_KEYFILE_PATH=/var/lib/mongodb/keyfile
 
 export CONTAINER_PORT=$(echo -n $IMAGE_EXPOSE_SERVICES | cut -d ':' -f 1)
 
@@ -160,10 +161,10 @@ function mongo_create_users() {
 # setup_keyfile fixes the bug in mounting the Kubernetes 'Secret' volume that
 # mounts the secret files with 'too open' permissions.
 function setup_keyfile() {
-  if [ ! -f "${MONGODB_KEYFILE_SOURCE_PATH}" ]; then
-    echo "ERROR: You have to provide the 'keyfile' in ${MONGODB_KEYFILE_SOURCE_PATH}"
+  if [ -z "${MONGODB_KEYFILE_VALUE}" ]; then
+    echo "ERROR: You have to provide the 'keyfile' value in ${MONGODB_KEYFILE_VALUE}"
     exit 1
   fi
-  cp ${MONGODB_KEYFILE_SOURCE_PATH} ${MONGODB_KEYFILE_PATH}
+  echo ${MONGODB_KEYFILE_VALUE} > ${MONGODB_KEYFILE_PATH}
   chmod 0600 ${MONGODB_KEYFILE_PATH}
 }
--- a/2.4/examples/replica/mongo_controller.json
+++ b/2.4/examples/replica/mongo_controller.json
@@ -1,134 +1,226 @@
 {
-  "kind": "Config",
-  "apiVersion": "v1beta1",
-  "id": "mongo-example",
-  "items":[
+  "kind": "Template",
+  "apiVersion": "v1",
+
+  "metadata": {
+    "name": "mongodb-replica-example",
+    "creationTimestamp": null,
+    "annotations": {
+      "description": "MongoDB Replication Example",
+      "icon": "icon-database",
+      "tags": "database,mongodb,replication"
+    }
+  },
+  "parameters": [
+    {
+      "name": "MONGODB_USER",
+      "description": "The username for MongoDB account to be created",
+      "generate": "expression",
+      "from": "[a-zA-Z0-9]{3}"
+    },
+    {
+      "name": "MONGODB_PASSWORD",
+      "description": "Password for the MongoDB user account",
+      "generate": "expression",
+      "from": "[a-zA-Z0-9]{12}"
+    },
+    {
+      "name": "MONGODB_DATABASE",
+      "description": "Name for the MongoDB database",
+      "value": "userdb"
+    },
+    {
+      "name": "MONGODB_ADMIN_PASSWORD",
+      "description": "Password for the MongoDB admin user",
+      "generate": "expression",
+      "from": "[a-zA-Z0-9]{12}"
+    },
+    {
+      "name": "MONGODB_REPLICA_NAME",
+      "description": " The name of the replSet ",
+      "value": "rs0"
+    },
+    {
+      "name": "MONGODB_SERVICE_NAME",
+      "description": "The name of the MongoDB Service (used to DNS lookup, default: 'mongodb')",
+      "value": "mongodb"
+    },
+    {
+      "name": "MONGODB_KEYFILE_VALUE",
+      "description": "The value of the MongoDB Key",
+      "generate": "expression",
+      "from": "[a-zA-Z0-9]{255}"
+    }
+  ],
+  "objects":[
     {
       "kind": "Service",
-      "apiVersion": "v1beta2",
-      "id": "mongodb",
-      "portalIP": "None",
-      "labels": {
-        "name": "mongodb"
-      },
-      "port": 27017,
-      "selector": {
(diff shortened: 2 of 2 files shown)


2.4/contrib/common.sh
```diff
@@ -160,10 +161,10 @@ function mongo_create_users() {
 # setup_keyfile fixes the bug in mounting the Kubernetes 'Secret' volume that
 # mounts the secret files with 'too open' permissions.
 function setup_keyfile() {
-  if [ ! -f "${MONGODB_KEYFILE_SOURCE_PATH}" ]; then
-    echo "ERROR: You have to provide the 'keyfile' in ${MONGODB_KEYFILE_SOURCE_PATH}"
+  if [ -z "${MONGODB_KEYFILE_VALUE}" ]; then
```
bparees, line 164: you can't just add mandatory new env variables like this.  our existing DB templates don't provide them and the doc needs to be updated if you're going to do something like this.

2.4/examples/replica/mongo_controller.json
```diff
@@ -1,134 +1,224 @@
...
+  "kind": "Template",
+  "apiVersion": "v1",
+
+  "metadata": {
+    "name": "mongo-replica-example",
+    "creationTimestamp": null,
+    "annotations": {
+      "description": "MySQL Replication Example",
+      "icon": "icon-database",
+      "tags": "database,mongodb,replication"
```
mnagy: s/mongodb/mysql

bparees:
This can't be merged as is, it would break all our existing use  of mongo.

2.4/examples/replica/mongo_controller.json
```diff
@@ -1,134 +1,224 @@
...
-  "id": "mongo-example",
-  "items":[
+  "kind": "Template",
+  "apiVersion": "v1",
+
+  "metadata": {
+    "name": "mongo-replica-example",
+    "creationTimestamp": null,
+    "annotations": {
+      "description": "MySQL Replication Example",
```
mnagy: MongoDB Replication Example

2.4/contrib/common.sh
bparees, line 164: unless this function is only being called when using the replica flow?

2.4/examples/replica/mongo_controller.json
```diff
@@ -1,134 +1,224 @@
...
-      "labels": {
-        "name": "mongodb"
-      },
-      "port": 27017,
-      "selector": {
-        "name": "mongo-replica"
-      }
+      "name": "MONGODB_USER",
+      "description": "The username for MongoDB account to be created",
+      "value": "user"
```
mnagy: Shouldn't we generate this as well? e.g. "user[A-Z0-9]{3}"

2.4/contrib/common.sh
mfojtik, line 164: this env is not mandatory, only used in replicated mode... existing DB templates don't need to set it.

2.4/examples/replica/mongo_controller.json
```diff
@@ -1,134 +1,224 @@
...
-      "kind": "Secret",
-      "apiVersion": "v1beta3",
+      "name": "MONGODB_PASSWORD",
+      "description": "Password for the MongoDB user account",
+      "generate": "expression",
+      "from": "[a-zA-Z0-9]{12}"
+    },
+    {
+      "name": "MONGODB_DATABASE",
+      "description": "Password for the MongoDB user account",
```
mnagy: Copy&paste error

2.4/examples/replica/mongo_controller.json
```diff
@@ -1,134 +1,224 @@
...
+      "value": "userdb"
```
mnagy, line 30: Same point wrt generating.

2.4/examples/replica/mongo_controller.json
```diff
@@ -1,134 +1,224 @@
...
+      "value": "mongodb"
+    },
+    {
+      "name": "MONGODB_KEYFILE_VALUE",
+      "description": "The value of the MongoDB Key",
+      "value": "WGjNxLzJUVkEwd25ualhJNmxXODl0NlBFdGw4YmM1MENTTUFGaU9ncDh3U3JCWTA2YVZGamdBdHlYQwpoQlhNUXNQVmUxcFRYYjRPTnRnVE52d0prckVFWmR6Wjd2WWZQclorV3VwK2NsWUlMc3M4QmJuS1g3bjRzbjFJClpjMERKclJZVXNoeEdrcmhQVXNPTlVZM1NDeHorZnFXb1ljcjV1YmE3MDlib3FMQ0N1Vy90Z0laQUhGUkQ0QXYKZ2JZU1owYTQ5NmZMMGx5ZUZxS3R4a0w2aW1OUEM3dExsY3I0OUwzMGdZMkhiQmRHOGhYRnF0TkdPS2pIWVViagoxTEoxNE9nOG0zYWZDYy9DaUY1WGdaZWx2cDFhb3gwZzRGK05pR1lUWVVHTEZNT3k3UjFTc01TRFBIT2NDMWlyCllBN25KZEJoaWl5cTRteVFQa0ZxSmM1TnJEZml1QlA5STFKNEpuQ3dKUml5MThmSzNoNjJwMzBkeWNWK2dkMkQKRTk3Y0N3NzdRSmtTdXRsY29oMWZwQ2gwSVlhdDNNalBYSCtacDRkdzZFZ3Q3STlYRmJQeHdqaGk1NW5icXhodwpYbVl1R1RETDVBUm55OFNZWUdVa0dXWk5VZzRsUlhSV0xGSTdCTGdnQmJ3QjhCQzl2eS90YmJ6a0xoVjZRUnJlCm1XZVYxRVhDMDYwM3FqT1cvcEFkN2JodFkwM3pMYXduUElXZWc2UEpkL3l4T0lYbjFsN1pHdkxpRzhsbXlpY3EKeVZyS1c0M3V2Nk1tTVZWam5BL1B2MituclBuSWcyQVgrdVJWc1JVdFRQZXpta0FQOEJCNjhaRUJscWhMUXNpMQpQSVhuUFdHMVozN3JYVjVQZHlyMTJrWUlhQzBBM0NRNDNMakRjYjJTT3FNNnlmOEdzc2hqamZnYXpiZEcraEw3Ck9PMU9LeVhyWEFWamx4dGRydm9UQ0UwcEFYQjFJSWdqN21NZ2lPM2pNZ0lwSUxSOERqNVdy"
```
bparees: why isn't this generated?

2.4/examples/replica/mongo_controller.json
jhadvig: +1

2.4/examples/replica/mongo_controller.json
jhadvig: +1

2.4/examples/replica/mongo_controller.json
mfojtik: i thought initially that this must be some openssl valid string but I was wrong... @jhadvig you can safely generate this value.

2.4/examples/replica/mongo_controller.json
```diff
@@ -1,134 +1,224 @@
...
+        "ports": [
+          {
+            "protocol": "TCP",
+            "port": 27017,
+            "targetPort": 27017,
+            "nodePort": 0
+          }
+        ],
+        "selector": {
+          "name": "mongo-replica"
```
mnagy: mongodb-replica for consistency?

2.4/examples/replica/mongo_controller.json
jhadvig: Was told by @mfojtik that it would be better for the demo purposes to keep it as a value(also he told me about some restriction) but the only restriction I found is taht the string has to be 6-1024 chars long. Will update it, so the OS will generate it

mfojtik: @jhadvig you should never believe me

jhadvig: @mfojtik :broken_heart:

2.4/examples/replica/mongo_controller.json
jhadvig, line 30: @mnagy we are not generating the db name in any template. Probably wont do exceptions here

jhadvig:
@bparees @mnagy updated due to your comments :+1:  PTAL

mfojtik:
LGTM

mnagy:
Scaling nicely :)

Issue #50, opened 2015-06-22
Cluster init not working against GA


smarterclayton:
Did a straight up create - can you replicate?  Want to show this at summit.

```
Mon Jun 22 00:04:04.462 [rsStart] replSet info you may need to run replSetInitiate -- rs.initiate() in the shell -- if that is not already done
=> Waiting for MongoDB service startup  ...
Mon Jun 22 00:04:05.046 [conn1] note: no users configured in admin.system.users, allowing localhost access
=> MongoDB service has started
+ mongo_add
++ mongo_addr
+++ container_addr
++++ cat /var/lib/mongodb/.address
+++ echo -n 172.17.0.70
++ echo -n 172.17.0.70:
++ mongo_primary_member_addr
+++ endpoints
+++ service_name=mongodb
+++ dig mongodb A +search +short
++ local 'current_endpoints=172.17.0.69
172.17.0.70'
+++ echo -n 172.17.0.69 172.17.0.70
+++ cut -d ' ' -f 1
++ local mongo_node=172.17.0.69:
+++ mongo admin -u admin -p 7HsVOJ8NglVa --host 172.17.0.69: --quiet --eval 'print(rs.isMaster().primary);'
exception: connect failed
++ echo -n Mon Jun 22 00:04:05.160 Error: Missing port number in connection string '"172.17.0.69:/admin"' at src/mongo/shell/mongo.js:129
+ echo '=> Adding 172.17.0.70: to Mon Jun 22 00:04:05.160 Error: Missing port number in connection string "172.17.0.69:/admin" at src/mongo/shell/mongo.js:129 ...'
=> Adding 172.17.0.70: to Mon Jun 22 00:04:05.160 Error: Missing port number in connection string "172.17.0.69:/admin" at src/mongo/shell/mongo.js:129 ...
++ mongo_primary_member_addr
+++ endpoints
+++ service_name=mongodb
+++ dig mongodb A +search +short
++ local 'current_endpoints=172.17.0.69
172.17.0.70'
+++ echo -n 172.17.0.69 172.17.0.70
+++ cut -d ' ' -f 1
++ local mongo_node=172.17.0.69:
+++ mongo admin -u admin -p 7HsVOJ8NglVa --host 172.17.0.69: --quiet --eval 'print(rs.isMaster().primary);'
exception: connect failed
[...]

PR #51, opened 2015-06-22
Fix replication of MongoDB

mfojtik:
- Remove `IMAGE_EXPOSE_SERVICES` as we don't use it anymore
- Create userdb first and admin afterwards (for some reason this fixed replication....)
- The `mongodb-service` is a Pod not DeploymentConfig (runs just once to initialize the replSet)


mfojtik:
@smarterclayton this worked for me, but you will have to rebuild the image.... The create from template. I tried to scale it up to 5 and the replSet was successfully copied.

mfojtik:
@bparees @jhadvig PTAL (tested locally, waiting for tests to pass as I was changing the user creation)

mfojtik:
Fixes: https://github.com/openshift/mongodb/issues/50#issuecomment-114047856

jhadvig:
LGTM :+1:

bparees:
lgtm.

PR #53, opened 2015-06-23
Adding mongodb user, pw, db to mongo-replica

jhadvig:
@mfojtik PTAL


jhadvig:
FYI: I've tested the change and it was working

mfojtik:
LGTM

@bparees I've found that without these you won't be able to connect to mongodb cluster with your client app as the env vars will not be provided :/

PR #54, opened 2015-06-30
Fix the labels to match new schema

mfojtik:


mfojtik:
@bparees @jhadvig @mnagy PTAL

(I'm preserving the old labels here for backward compatibility)

bparees:
@mfojtik only preserving them in the rhel image?

PR #56, opened 2015-07-02
Rename replication template to 'template.json'

mfojtik:


2.4/examples/replica/README.md
```diff
@@ -88,4 +88,4 @@ For deleting all the created resources use:
 
 ```
 $ oc delete all --all
-```
```
mfojtik, line 91: no idea what changed here ;-)

mfojtik:
@adellape @bparees FYI (the word "controller" is now misleading as the template is not just a list of that defines a replication controller for mongodb, but it is a regular template. Also since the "replica" is in the URL already (the folder is called that way) I think we can just make it easy by calling this just "template.json" (it also looks good on command line ;-)

bparees:
we don't call any of our other templates "template.json", i suggest "mongo_replicated.json" or something.

bparees:
don't assume the file will always live in this folder, installers or something else might copy it to a directory full of template files, for example.

mfojtik:
@bparees valid points... how about "mongodb_replication_template.json"? too long?

bparees:
we use dashes in our other names.  we don't append "template" in our example app templates, but we do in our other DB templates, so i can take or leave that.

i still like "mongodb-replicated" or "mongodb-replicated-template"

bparees:
(ie replicated instead of replication)

bparees:
or mongodb-clustered

mfojtik:
`mongodb-clustered` sounds good to me

PR #60, opened 2015-07-27
Add mongodb-2.6 image

omron93:
Add rh-mongodb26 based image. All tests from 2.4 pass fine.

### DIFF
--- a/2.6/Dockerfile
+++ b/2.6/Dockerfile
@@ -1,0 +1,56 @@
+FROM centos:centos7
+MAINTAINER SoftwareCollections.org <sclorg@redhat.com>
+
+# MongoDB image for OpenShift.
+#
+# Volumes:
+#  * /var/lib/mongodb/data - Datastore for MongoDB
+# Environment:
+#  * $MONGODB_USER - Database user name
+#  * $MONGODB_PASSWORD - User's password
+#  * $MONGODB_DATABASE - Name of the database to create
+#  * $MONGODB_ADMIN_PASSWORD - Password of the MongoDB Admin
+
+
+ENV MONGODB_VERSION=2.6
+
+LABEL io.k8s.description="MongoDB is a scalable, high-performance, open source NoSQL database." \
+      io.k8s.display-name="MongoDB 2.6" \
+      io.openshift.expose-services="27017:mongodb" \
+      io.openshift.tags="database,mongodb,rh-mongodb26"
+
+EXPOSE 27017
+
+COPY run-mongod.sh /usr/local/bin/
+COPY contrib /var/lib/mongodb/
+
+# This image must forever use UID 184 for mongodb user and GID 998 for the
+# mongodb group, so our volumes are safe in the future. This should *never*
+# change, the last test is there to make sure of that.
+# Due to the https://bugzilla.redhat.com/show_bug.cgi?id=1206151,
+# the whole /var/lib/mongodb/ dir has to be chown-ed.
+RUN yum install -y epel-release && \
+    rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
+    yum -y --setopt=tsflags=nodocs install \
+    https://www.softwarecollections.org/en/scls/rhscl/v8314/epel-7-x86_64/download/rhscl-v8314-epel-7-x86_64.noarch.rpm \
+    https://www.softwarecollections.org/en/scls/rhscl/rh-mongodb26/epel-7-x86_64/download/rhscl-rh-mongodb26-epel-7-x86_64.noarch.rpm \
+    https://www.softwarecollections.org/en/scls/rhscl/rh-java-common/epel-7-x86_64/download/rhscl-rh-java-common-epel-7-x86_64.noarch.rpm && \
+    yum install -y --setopt=tsflags=nodocs bind-utils gettext iproute v8314 rh-mongodb26-mongodb rh-mongodb26 && \
+    yum clean all && \
+    mkdir -p /var/lib/mongodb/data && chown -R mongodb.mongodb /var/lib/mongodb/ && \
+    test "$(id mongodb)" = "uid=184(mongodb) gid=998(mongodb) groups=998(mongodb)" && \
+    chmod o+w -R /var/lib/mongodb && chmod o+w -R /var/opt/rh/rh-mongodb26/lib/mongodb
+
+# When bash is started non-interactively, to run a shell script, for example it
+# looks for this variable and source the content of this file. This will enable
+# the SCL for all scripts without need to do 'scl enable'.
+ENV BASH_ENV=/var/lib/mongodb/scl_enable \
+    ENV=/var/lib/mongodb/scl_enable \
+    PROMPT_COMMAND=". /var/lib/mongodb/scl_enable"
+
+VOLUME ["/var/lib/mongodb/data"]
+
+USER 184
+
+ENTRYPOINT ["run-mongod.sh"]
+CMD ["mongod"]
--- a/2.6/Dockerfile.rhel7
+++ b/2.6/Dockerfile.rhel7
@@ -1,0 +1,59 @@
+FROM rhel7
+MAINTAINER SoftwareCollections.org <sclorg@redhat.com>
+
+# MongoDB image for OpenShift.
+#
+# Volumes:
+#  * /var/lib/mongodb/data - Datastore for MongoDB
+# Environment:
+#  * $MONGODB_USER - Database user name
+#  * $MONGODB_PASSWORD - User's password
+#  * $MONGODB_DATABASE - Name of the database to create
+#  * $MONGODB_ADMIN_PASSWORD - Password of the MongoDB Admin
+
+ENV MONGODB_VERSION=2.6
+
+LABEL io.k8s.description="MongoDB is a scalable, high-performance, open source NoSQL database." \
+      io.k8s.display-name="MongoDB 2.6" \
+      io.openshift.expose-services="27017:mongodb" \
+      io.openshift.tags="database,mongodb,rh-mongodb26"
+
+# Labels consumed by Red Hat build service
+LABEL BZComponent="rh-mongodb26-docker" \
+      Name="rhscl/mongodb-26-rhel7" \
+      Version="2.6" \
+      Release="1" \
+      Architecture="x86_64"
+
+EXPOSE 27017
+
+COPY run-mongod.sh /usr/local/bin/
+COPY contrib /var/lib/mongodb/
+
+# This image must forever use UID 184 for mongodb user and GID 998 for the
+# mongodb group, so our volumes are safe in the future. This should *never*
+# change, the last test is there to make sure of that.
+# Due to the https://bugzilla.redhat.com/show_bug.cgi?id=1206151,
+# the whole /var/lib/mongodb/ dir has to be chown-ed.
+RUN yum install -y yum-utils && \
+    yum-config-manager --enable rhel-server-rhscl-7-rpms && \
+    yum-config-manager --enable rhel-7-server-optional-rpms && \
+    yum install -y --setopt=tsflags=nodocs bind-utils gettext iproute v8314 rh-mongodb26-mongodb rh-mongodb26 && \
+    yum clean all && \
+    mkdir -p /var/lib/mongodb/data && chown -R mongodb.mongodb /var/lib/mongodb/ && \
+    test "$(id mongodb)" = "uid=184(mongodb) gid=998(mongodb) groups=998(mongodb)" && \
+    chmod o+w -R /var/lib/mongodb && chmod o+w -R /var/opt/rh/rh-mongodb26/lib/mongodb
+
+# When bash is started non-interactively, to run a shell script, for example it
+# looks for this variable and source the content of this file. This will enable
+# the SCL for all scripts without need to do 'scl enable'.
+ENV BASH_ENV=/var/lib/mongodb/scl_enable \
+    ENV=/var/lib/mongodb/scl_enable \
+    PROMPT_COMMAND=". /var/lib/mongodb/scl_enable"
+
+VOLUME ["/var/lib/mongodb/data"]
+
+USER 184
+
+ENTRYPOINT ["run-mongod.sh"]
+CMD ["mongod"]
--- a/2.6/contrib/common.sh
+++ b/2.6/contrib/common.sh
@@ -1,0 +1,167 @@
+#!/bin/bash
+
+# Used for wait_for_mongo_* functions
+MAX_ATTEMPTS=60
+SLEEP_TIME=1
+
+export MONGODB_CONFIG_PATH=/var/lib/mongodb/mongodb.conf
+export MONGODB_PID_FILE=/var/lib/mongodb/mongodb.pid
+export MONGODB_KEYFILE_PATH=/var/lib/mongodb/keyfile
+export CONTAINER_PORT=27017
+
+# container_addr returns the current container external IP address
+function container_addr() {
+  echo -n $(cat /var/lib/mongodb/.address)
+}
+
+# mongo_addr returns the IP:PORT of the currently running MongoDB instance
+function mongo_addr() {
+  echo -n "$(container_addr):${CONTAINER_PORT}"
+}
+
+# cache_container_addr waits till the container gets the external IP address and
+# cache it to disk
+function cache_container_addr() {
+  echo -n "=> Waiting for container IP address ..."
+  for i in $(seq $MAX_ATTEMPTS); do
+    result=$(ip addr | grep 'state UP' -A2 | tail -n1 | awk '{print $2}' | cut -f1 -d'/')
+    if [ ! -z "${result}" ]; then
+      echo -n $result > /var/lib/mongodb/.address
+      echo " $(mongo_addr)"
+      return 0
+    fi
+    sleep $SLEEP_TIME
+  done
+  echo "Failed to get Docker container IP address." && exit 1
+}
+
+# wait_for_mongo_up waits until the mongo server accepts incomming connections
+function wait_for_mongo_up() {
+  local mongo_host="${1-}"
+  local mongo_cmd="mongo admin "
+
+  if [ ! -z "${mongo_host}" ]; then
+    mongo_cmd+="--host ${mongo_host}:${CONTAINER_PORT} "
+  fi
+
+  for i in $(seq $MAX_ATTEMPTS); do
+    echo "=> Waiting for MongoDB service startup ${mongo_host} ..."
+    set +e
+    $mongo_cmd --eval "help" &>/dev/null
+    status=$?
+    set -e
+    if [ $status -eq 0 ]; then
+      echo "=> MongoDB service has started"
+      return 0
+    fi
+    sleep $SLEEP_TIME
+  done
+  echo "=> Giving up: Failed to start MongoDB service!"
+  exit 1
+}
+
+# wait_for_mongo_down waits until the mongo server is down
+function wait_for_mongo_down() {
+  for i in $(seq $MAX_ATTEMPTS); do
+    echo "=> Waiting for MongoDB service shutdown ..."
+    set +e
+    mongo admin --eval "help" &>/dev/null
+    status=$?
+    set -e
+    if [ $status -ne 0 ]; then
+      echo "=> MongoDB service has stopped"
+      return 0
+    fi
+    sleep $SLEEP_TIME
+  done
+  echo "=> Giving up: Failed to stop MongoDB service"
+  exit 1
+}
+
+# endpoints returns list of IP addresses with other instances of MongoDB
+# To get list of endpoints, you need to have headless Service named 'mongodb'.
+# NOTE: This won't work with standalone Docker container.
+function endpoints() {
+  service_name=${MONGODB_SERVICE_NAME:-mongodb}
+  dig ${service_name} A +search +short 2>/dev/null
+}
+
+# no_endpoints returns true if the only endpoint is the current container itself
+# or there are no endpoints registered (running as standalone Docker container)
+function no_endpoints() {
+  [ -z "$(endpoints)" ] && return 0
+  [ "$(endpoints)" == "$(container_addr)" ]
+}
+
+# build_mongo_config builds the MongoDB replicaSet config used for the cluster
+# initialization
+function build_mongo_config() {
+  local members="{ _id: 0, host: \"$(mongo_addr)\"},"
+  local member_id=1
+  local container_addr="$(container_addr)"
+  for node in $(endpoints); do
+    [ "$node" == container_addr ] && continue
+    members+="{ _id: ${member_id}, host: \"${node}:${CONTAINER_PORT}\"},"
+    let member_id++
+  done
+  echo -n "var config={ _id: \"${MONGODB_REPLICA_NAME}\", members: [ ${members%,} ] }"
+}
+
+# mongo_initiate initiate the replica set
+function mongo_initiate() {
+  local mongo_wait="while (rs.status().startupStatus || (rs.status().hasOwnProperty(\"myState\") && rs.status().myState != 1)) { printjson( rs.status() ); sleep(1000); }; printjson( rs.status() );"
+  config=$(build_mongo_config)
+  echo "=> Initiating MongoDB replica using: ${config}"
+  mongo admin --eval "${config};rs.initiate(config);${mongo_wait}"
+}
+
+# get the address of the current primary member
+function mongo_primary_member_addr() {
+  local current_endpoints=$(endpoints)
+  local mongo_node="$(echo -n ${current_endpoints} | cut -d ' ' -f 1):${CONTAINER_PORT}"
+  echo -n $(mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --host ${mongo_node} --quiet --eval "print(rs.isMaster().primary);")
+}
+
+# mongo_remove removes the current MongoDB from the cluster
+function mongo_remove() {
+  echo "=> Removing $(mongo_addr) on $(mongo_primary_member_addr) ..."
+  mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" \
+    --host $(mongo_primary_member_addr) --eval "rs.remove('$(mongo_addr)');" &>/dev/null || true
+}
+
+# mongo_add advertise the current container to other mongo replicas
+function mongo_add() {
+  echo "=> Adding $(mongo_addr) to $(mongo_primary_member_addr) ..."
+  mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" \
+    --host $(mongo_primary_member_addr) --eval "rs.add('$(mongo_addr)');"
+}
+
+# run_mongod_supervisor runs the MongoDB replica supervisor that manages
+# registration of the new members to the MongoDB replica cluster
+function run_mongod_supervisor() {
+  local log_pipe_path=$(mktemp --suffix=log)
+  rm -rf ${log_pipe_path}
+  ( mkfifo ${log_pipe_path} && cat ${log_pipe_path} ) &
+  /var/lib/mongodb/replica_supervisor.sh ${log_pipe_path} &
+}
+
+# mongo_create_users creates the MongoDB admin user and the database user
+# configured by MONGO_USER
+function mongo_create_users() {
+  mongo admin --eval "db.createUser({user: 'admin', pwd: '${MONGODB_ADMIN_PASSWORD}', roles: ['dbAdminAnyDatabase', 'userAdminAnyDatabase' , 'readWriteAnyDatabase','clusterAdmin' ]})"
+
+  mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --eval "db.getSiblingDB('${MONGODB_DATABASE}').createUser({user: '${MONGODB_USER}', pwd: '${MONGODB_PASSWORD}', roles: [ 'readWrite' ]});"
+
+  touch /var/lib/mongodb/data/.mongodb_datadir_initialized
+}
+
+# setup_keyfile fixes the bug in mounting the Kubernetes 'Secret' volume that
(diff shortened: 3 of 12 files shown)


jhadvig:
[test]

2.6/examples/replica/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+A cluster of MongoDB servers that implements master-slave replication and automated failover.
+MongoDB’s recommended replication strategy.
+
+## Deployment:
+
+To create MongoDB replica set in OpenShift v3, you can use the example template
+included in this repository:
+
+* `oc create -f ./2.4/examples/replica/mongodb-clustered.json`
```
jhadvig: 2.4 -> 2.6

hack/build.sh
```diff
@@ -50,7 +50,11 @@ function squash {
 dirs=${VERSION:-$VERSIONS}
 
 for dir in ${dirs}; do
-  IMAGE_NAME="${BASE_IMAGE_NAME}-${dir//./}-${OS}"
+  if [ "$dir" == "2.6" ]; then
+    IMAGE_NAME=rhscl/rh-mongodb${dir//./}-${OS}
```
mfojtik: Why is this needed?

hack/build.sh
```diff
@@ -50,7 +50,11 @@ function squash {
 dirs=${VERSION:-$VERSIONS}
 
 for dir in ${dirs}; do
-  IMAGE_NAME="${BASE_IMAGE_NAME}-${dir//./}-${OS}"
+  if [ "$dir" == "2.6" ]; then
```
jhadvig, line 52: Is the if statement necessary here ?

2.6/test/run
```diff
@@ -0,0 +1,179 @@
...
+    local run_cmd="$2"
+    local expected="$3"
+
+    echo "  Testing the image SCL enable"
+    out=$(docker run --rm ${IMAGE_NAME} /bin/bash -c "${run_cmd}")
+    if ! echo "${out}" | grep -q "${expected}"; then
+        echo "ERROR[/bin/bash -c "${run_cmd}"] Expected '${expected}', got '${out}'"
+        return 1
+    fi
+    #out=$(docker run --rm -i ${IMAGE_NAME} /bin/bash -i <<< "${run_cmd}")
```
jhadvig: Why all the comments ?

2.6/Dockerfile.rhel7
```diff
@@ -0,0 +1,63 @@
...
+# change, the last test is there to make sure of that.
+# Due to the https://bugzilla.redhat.com/show_bug.cgi?id=1206151,
+# the whole /var/lib/mongodb/ dir has to be chown-ed.
+RUN yum install -y yum-utils && \
+    yum-config-manager --enable rhel-server-rhscl-7-rpms && \
+    yum-config-manager --enable rhel-7-server-optional-rpms && \
+    yum install -y --setopt=tsflags=nodocs bind-utils gettext iproute v8314 rh-mongodb26-mongodb rh-mongodb26 && \
+    yum clean all && \
+    mkdir -p /var/lib/mongodb/data && chown -R mongodb.mongodb /var/lib/mongodb/ && \
+    test "$(id mongodb)" = "uid=184(mongodb) gid=998(mongodb) groups=998(mongodb)" && \
```
mfojtik, line 44: @bparees I think this check is no longer needed as we run the image under random UID. cc @mnagy

2.6/test/run
```diff
@@ -0,0 +1,179 @@
...
+# Test the MongoDB image.
+#
+# IMAGE_NAME specifies a name of the candidate image used for testing.
+# The image has to be available before this script is executed.
+#
+
+set -exo nounset
+shopt -s nullglob
+
+IMAGE_NAME=${IMAGE_NAME-rhscl/rh-mongodb26-centos7-candidate}
```
jhadvig: why `rhscl` ???

2.6/Dockerfile
```diff
@@ -0,0 +1,56 @@
+FROM centos:centos7
+MAINTAINER  Jakub Hadvig <jhadvig@redhat.com>
```
jhadvig: Please check how the `MAINTAINER`s are handled in https://github.com/openshift/sti-ruby/tree/master/2.2 and do it pls accordingly

2.6/Dockerfile.rhel7
```diff
@@ -0,0 +1,63 @@
+FROM rhel7
+MAINTAINER  Jakub Hadvig <jhadvig@redhat.com>
```
jhadvig: same as previous comment, regarding `MAINTEINER`

jhadvig:
@omron93 done with first line of review, have you tested this change? 
Also could you pls squash into one commit after you update the PR.

2.6/test/run
omron93: This is the copy of 2.4/test/run. So why are the comments in 2.4/test/run?

2.6/examples/replica/README.md
omron93: OK. I overlooked this...

hack/build.sh
omron93: Name of the rh-mongodb26 image should be named for example rhscl/rh-mongodb26-rhel7 - see "Docker Images Naming and Versioning Policy"
And it is quite different to openshift/mongodb-24-rhel7.

hack/build.sh
omron93, line 52: Same above - https://github.com/openshift/mongodb/pull/60#discussion_r36095751

2.6/test/run
omron93: This also relates to naming - https://github.com/openshift/mongodb/pull/60#discussion_r36095751

2.6/Dockerfile
omron93: OK. Changed to SoftwareCollections.org sclorg@redhat.com

2.6/Dockerfile.rhel7
omron93: OK. Changed to SoftwareCollections.org sclorg@redhat.com

omron93:
@jhadvig Thanks. Yes I tested it. Also by "make test". Are there any other tests I could perform?

hack/build.sh
jhadvig: Can you please point me out to that doc ?

hack/build.sh
jhadvig, line 52: Still don't get why mongo26 has to be build with different tag.

omron93, line 52: It it a part of releasing containers as a part of RHSCL. I don't know if even mongodb24 should follow "RHSCL" naming policy...
But I don't know much about this naming. @hhorak could explain more...

hack/build.sh
jhadvig: so we already raised an objection for that naming convention, since this naming scheme will break down if we ever want to support a repo with a name that ends in a number.

2.6/Dockerfile.rhel7
bparees, line 44: if so the same applies to the centos dockerfile above.  and i think you're probably right, the chown o+w right below this presumably takes care of that.

mnagy, line 44: yes, we can remove the check and also the comment above the RUN instruction about UID.

2.6/examples/replica/mongodb-clustered.json
```diff
@@ -0,0 +1,218 @@
...
+          "name": "mongodb-service"
+        }
+      },
+      "spec": {
+        "restartPolicy": "Never",
+        "dnsPolicy": "ClusterFirst",
+        "containers": [
+          {
+            "name": "initiate",
+            "image": "openshift/mongodb-24-centos7",
```
bparees: mongodb-26-centos7

2.6/examples/replica/mongodb-clustered.json
```diff
@@ -0,0 +1,218 @@
...
+            "creationTimestamp": null,
+            "labels": {
+              "name": "mongodb-replica"
+            }
+          },
+          "spec": {
+            "containers": [
+              {
+                "name":  "member",
+                "image": "openshift/mongodb-24-centos7",
```
bparees: mongodb-26-centos7

hack/build.sh
bparees: at a minimum it can still be "rh-${BASE_IMAGE_NAME}${dir//./}-${OS}"

omron93: OK. It could be this:

```
if [ "$dir" == "2.6" ]; then
    IMAGE_NAME=rhscl/rh-${BASE_DIR_NAME}${dir//./}-${OS}
else
    IMAGE_NAME="${BASE_IMAGE_NAME}-${dir//./}-${OS}"
fi
```

But `if` still have to be there.
Any idea?

2.6/Dockerfile.rhel7
omron93, line 44: OK. But again this file is a "copy" of Dockerfile from 2.4. So should I change this also in 2.4 directory in this PR?

2.6/examples/replica/mongodb-clustered.json
omron93: "image": "rhscl/rh-mongodb26-centos7"

2.6/examples/replica/mongodb-clustered.json
omron93: "image": "rhscl/rh-mongodb26-centos7"

2.6/test/run
jhadvig: so this has been take care of in https://github.com/openshift/mongodb/pull/64/files. Pls remove it as well

hhorak:
@jhadvig Marek is correct, for mongo 2.6, the image will be available as rhscl/rh-mongodb26-rhel7 in internal rh registry. Whether this must be mirrored also in build.sh is a question, where is this script actually run?

jhadvig:
@omron93 fine if the `make test` passed, but did you also tested the clustered use-case (that will require manual testing and following the README that is in the `examples/replica` dir )

hack/build.sh
jhadvig, line 52: @bparees are we suppose to build images with this naming ? I would guess not.

bparees, line 52: @jhadvig if it's the SCL conventions i don't have a huge problem with it.  if we need to go back and change the existing ones it's more of a problem (need to not break people) but we can definitely use the new scheme going forward, at a minimum.

that said, sure my preference would be to stick with the scheme we have, so I guess I would ask @hhorak if this is absolutely necessary?

bparees:
@hhorak build.sh is what is run in our jenkins system to produce the images that get pushed to dockerhub (centos version) and to our internal private registry for QE (rhel + centos versions)

it would be best to keep the naming scheme consistent between what is built by our jenkins job and what SCL ultimately produces, so yes we should probably update the build.sh if the OSBS images you're producing will be named rh-*

however we've already expressed concerns about the lack of a separator between "mongodb" and the version since it will break once you have a framework that ends in a numeric value.  I'd like to get that settled before introducing an image using this pattern.

omron93:
>  did you also tested the clustered use-case (that will require manual testing and following the README that is in the examples/replica dir )

@jhadvig So please provide information about how to test it. What are the use cases (resizing,...) ? How the oc command works? Where to test it?

jhadvig:
@omron93 here is the README for the mongodb-24 https://github.com/openshift/mongodb/tree/master/2.4/examples/replica
for mongodb-26 it should be similar,  but I guess there will need be to some changes in the scripts, sice its different version of mongodb.

Regarding the `oc` command we have docs at https://docs.openshift.org/latest/cli_reference/overview.html

omron93:
@jhadvig So replication in MongoDB should support creating/deleting replicaset and its scaling? Nothing more. Yes?

And where to test it? Are there some server running OpenShift3 where could be possible to test this? Or is it possible to run OpenShift easily in a container?

jhadvig:
@omron93 we are testing it in a vagrant machine, as most of our stuff.
Here is how to setup the vagrant env:
https://github.com/openshift/origin/blob/master/CONTRIBUTING.adoc#develop-on-virtual-machine-using-vagrant
And here is and example how to start and OS server and create and sample app in vagrant:
https://github.com/openshift/origin/tree/master/examples/sample-app

omron93:
@jhadvig Thanks. Replication now should be working. Please test...

2.6/test/run
omron93: OK. Removed.

jhadvig:
[test]

2.6/Dockerfile.rhel7
jhadvig, line 44: @omron93 no, please just get rid of it here. we create a separate PR for the mongo-24 since its not and issue of this PR

omron93, line 44: OK, I will remove it. But for now I am not sure what all to remove. What is the result of #67?

jhadvig, line 44: ok, so that taken care of.

rhcarvalho:
@omron93 the replica set example in 2.4 has [problems that I'm currently working on](https://github.com/openshift/mongodb/issues/70), and this involves relatively big changes to how it works and how it is documented. Before copy-pasting-adjusting the example for 2.6, I'd improve the example we currently have and therefore do not include the example for 2.6.

Unfortunately we don't have good automated tests to cover the replication example, and without that it would be pain to maintain two PoCs.

jhadvig:
@omron93 i've tested the replica for the and the standalone mongodb-26. Now we just have to figure out as @rhcarvalho suggested if we will incorporate the scaled mongodb example in this version. @bparees thoughts ?
Anyway I'm fine to merge this PR. LGTM

2.6/run-mongod.sh
```diff
@@ -0,0 +1,110 @@
...
+  echo "  MONGODB_USER"
+  echo "  MONGODB_PASSWORD"
+  echo "  MONGODB_DATABASE"
+  echo "  MONGODB_ADMIN_PASSWORD"
+  echo "Optional variables:"
+  echo "  MONGODB_REPLICA_NAME"
+  echo "MongoDB settings:"
+  echo "  MONGODB_NOPREALLOC (default: true)"
+  echo "  MONGODB_SMALLFILES (default: true)"
+  echo "  MONGODB_QUIET (default: false)"
```
rhcarvalho: This is copy-pasting a mistake from 2.4, [line 16](https://github.com/openshift/mongodb/pull/60/files#diff-4dc45b8cc3069c3d1b885dd850849a39R16) makes `MONGODB_QUIET` default to `true`...

2.6/run-mongod.sh
```diff
@@ -0,0 +1,110 @@
+#!/bin/bash
+
+# For SCL enablement
```
rhcarvalho: this comment is inaccurate

2.6/run-mongod.sh
```diff
@@ -0,0 +1,110 @@
...
+source /var/lib/mongodb/common.sh
+
+set -eu
+
+# Data directory where MongoDB database files live. The data subdirectory is here
+# because .bashrc and mongodb.conf both live in /var/lib/mongodb/ and we don't want a
```
rhcarvalho: there's no `.bashrc` file in this repo nor in `/var/lib/mongodb/`, this is a historical leftover.

2.6/Dockerfile.rhel7
bparees, line 44: yeah i guess we're leaving it :)

bparees:
address @rhcarvalho 's comments and squash the commits and i give this a lgtm.

as for the replica example, i do think we ultimately want it here, but we can come back to that.

omron93:
@bparees OK. All should be added and squashed into one commit...

rhcarvalho:
@omron93 the replication example is still a copy-paste of old code. It fails in the scenario reported in #70. I've also recently made changes to other files that are outdated in this PR.
Files changed recently:

```
 2.4/contrib/common.sh                       
 2.4/examples/replica/README.md              
 2.4/examples/replica/mongodb-clustered.json 
 2.4/run-mongod.sh                           
```

2.6/Dockerfile.rhel7
```diff
@@ -0,0 +1,59 @@
...
+# When bash is started non-interactively, to run a shell script, for example it
+# looks for this variable and source the content of this file. This will enable
+# the SCL for all scripts without need to do 'scl enable'.
+ENV BASH_ENV=/var/lib/mongodb/scl_enable \
+    ENV=/var/lib/mongodb/scl_enable \
+    PROMPT_COMMAND=". /var/lib/mongodb/scl_enable"
+
+VOLUME ["/var/lib/mongodb/data"]
+
+USER 184
```
mfojtik, line 56: @bparees should we replace 184 with 1001? I thought we already did that for some other database image.

hack/build.sh
mfojtik, line 52: this will make sharing `build.sh` between multiple repos more complex. The refactor I did few weeks(?) ago attempted to unify the build.sh and Makefile so we don't have to manually patch each one in every repo when we do changes.

2.6/Dockerfile.rhel7
bparees, line 56: no, i ended up backing down on that change.

omron93:
All recently made changes should be added to this commit.

omron93:
Changed image name to `rhscl/mongodb-26-$OS`

bparees:
lgtm, @mfojtik or @jhadvig you have any final comments?

[test]

2.6/contrib/common.sh
```diff
@@ -0,0 +1,167 @@
...
+  ( mkfifo ${log_pipe_path} && cat ${log_pipe_path} ) &
+  /var/lib/mongodb/replica_supervisor.sh ${log_pipe_path} &
+}
+
+# mongo_create_users creates the MongoDB admin user and the database user
+# configured by MONGO_USER
+function mongo_create_users() {
+  mongo admin --eval "db.createUser({user: 'admin', pwd: '${MONGODB_ADMIN_PASSWORD}', roles: ['dbAdminAnyDatabase', 'userAdminAnyDatabase' , 'readWriteAnyDatabase','clusterAdmin' ]})"
+
+  mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --eval "db.getSiblingDB('${MONGODB_DATABASE}').createUser({user: '${MONGODB_USER}', pwd: '${MONGODB_PASSWORD}', roles: [ 'readWrite' ]});"
```
rhcarvalho, line 153: Why are we connecting to the `admin` database just to, as the first command to eval,  connect to another one?

We can easily mirror what we have in place for 2.4, just replacing `db.addUser` with `db.createUser`:

``` bash
mongo ${MONGODB_DATABASE} --eval "db.createUser({user: '${MONGODB_USER}', pwd: '${MONGODB_PASSWORD}', roles: [ 'readWrite' ]});"

mongo admin --eval "db.createUser({user: 'admin', pwd: '${MONGODB_ADMIN_PASSWORD}', roles: ['dbAdminAnyDatabase', 'userAdminAnyDatabase' , 'readWriteAnyDatabase','clusterAdmin' ]})"
```

2.6/examples/replica/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+## How does it work?
+
+### Service 'mongodb'
+
+This resource defines a [headless service](http://kubernetes.io/v1.0/docs/user-guide/services.html#headless-services)
+that serves as an entry point to the replica set. The service endpoints are
+the pods created by the 'mongodb' DeploymentConfig.
+
+
```
rhcarvalho: This is almost a copy-paste from `2.4/examples/replica/README.md`... except for 3 extra empty lines...
Here is one.

2.6/examples/replica/README.md
```diff
@@ -0,0 +1,84 @@
...
+### DeploymentConfig 'mongodb'
+
+This resource defines a [deployment configuration](https://docs.openshift.org/latest/architecture/core_concepts/deployments.html#deployments-and-deployment-configurations) to manage replica set members.
+Each member starts the MongoDB server without replication data. Once it is
+ready, it advertises itself to the current replica set PRIMARY, which will then
+add it to the replica set.
+When a member pod is destroyed, it is automatically removed from the replica set.
+
+
```
rhcarvalho: Here's the second one.

2.6/examples/replica/README.md
```diff
@@ -0,0 +1,84 @@
...
+To add/remove members to the replica set you can use the `oc scale` command.
+This will scale the deployment up to 5 pods:
+
```
rhcarvalho: And the last one.

2.6/examples/replica/mongodb-clustered.json
```diff
@@ -0,0 +1,167 @@
+{
+  "kind": "Template",
+  "apiVersion": "v1",
+
```
rhcarvalho: This would be identical to `2.4/examples/replica/mongodb-clustered.json`, if it wasn't for this lonely empty line...

2.6/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+    rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
+    yum -y --setopt=tsflags=nodocs install \
+    https://www.softwarecollections.org/en/scls/rhscl/v8314/epel-7-x86_64/download/rhscl-v8314-epel-7-x86_64.noarch.rpm \
+    https://www.softwarecollections.org/en/scls/rhscl/rh-mongodb26/epel-7-x86_64/download/rhscl-rh-mongodb26-epel-7-x86_64.noarch.rpm \
+    https://www.softwarecollections.org/en/scls/rhscl/rh-java-common/epel-7-x86_64/download/rhscl-rh-java-common-epel-7-x86_64.noarch.rpm && \
+    yum install -y --setopt=tsflags=nodocs bind-utils gettext iproute v8314 rh-mongodb26-mongodb rh-mongodb26 && \
+    yum clean all && \
+    mkdir -p /var/lib/mongodb/data && chown -R mongodb.mongodb /var/lib/mongodb/ && \
+    test "$(id mongodb)" = "uid=184(mongodb) gid=998(mongodb) groups=998(mongodb)" && \
+    chmod o+w -R /var/lib/mongodb && chmod o+w -R /var/opt/rh/rh-mongodb26/lib/mongodb
```
rhcarvalho, line 42: Just a note, the path changed from `/opt/rh/mongodb24/root/var/lib/mongodb` to `/var/opt/rh/rh-mongodb26/lib/mongodb`. I believe the new version of SCL changed the path structure?

rhcarvalho:
@bparees we don't have automated tests for the replication scenarios yet. I'm still skeptical about adding the replication example for 2.6 that is copy-paste from 2.4. If everything is all the same, how good is it to maintain two copies of those files?

I would have one single example, and mention that you can replace /4/6/ and it should work the same.

2.6/examples/replica/mongodb-clustered.json
```diff
@@ -0,0 +1,167 @@
...
+          "metadata": {
+            "labels": {
+              "name": "mongodb-replica"
+            }
+          },
+          "spec": {
+            "containers": [
+              {
+                "name":  "mongodb",
+                "image": "rhscl/mongodb-26-centos7",
```
rhcarvalho: I might be wrong, but I think this example won't work out of the box. I can't find [rhscl/mongodb-26-centos7](https://hub.docker.com/u/rhscl/mongodb-26-centos7/) nor even [rhscl](https://hub.docker.com/u/rhscl/) in Docker Hub... both 404.

2.6/examples/replica/README.md
omron93: OK. All fixed.

2.6/Dockerfile
omron93, line 42: Right, in RHSCL 2.0 path is changed.

2.6/contrib/common.sh
omron93, line 153: Simple replacing is not possible. It is because of a change in "localhost exeption". In mongodb 2.6 it is only active till first user is created (in 2.4 it was active all the time):

> Localhost Exception
> 
> The localhost exception allows you to enable authorization before creating the first user in the system. When active, the localhost exception allows all connections from the localhost interface to have full access to that instance. The exception applies only when there are no users created in the MongoDB instance.

https://docs.mongodb.org/v2.6/core/authentication/

This is a problem during creating users in replication. In mongodb 2.6 setting keyfile implies authorization (http://docs.mongodb.org/v2.6/reference/configuration-options/#security.keyFile). So auth is on and "normal user" it not allowed to create new users. So admin user have to be created first...

omron93:
@bparees @rhcarvalho I don't have anything against removing replication example for 2.6. But I have to know what is the final decision?

2.6/contrib/common.sh
rhcarvalho, line 153: @omron93 what you wrote above is correct, but only applies to when you run `mongod` with the `--auth` flag. During user creation [we start the server with authentication disabled](https://github.com/openshift/mongodb/pull/60/files#diff-4dc45b8cc3069c3d1b885dd850849a39R80).

I went ahead and ran the tests using the two lines I proposed above (mirroring 2.4), and it works as expected.

omron93, line 153: > I went ahead and ran the tests using the two lines I proposed above (mirroring 2.4), and it works as expected.

@rhcarvalho  And do you test also replication?
Yes, standalone mongod works. But the problem is when you run replication!

> During user creation we start the server with authentication disabled.
> When you run standalone mongod server auth is disabled during user creation. So there works both versions - mirroring 2.4 and also the actual in this PR (authenticating to admin database and then creating normal user)

But when replication is running and users are created (https://github.com/openshift/mongodb/blob/master/2.4/contrib/initiate_replica.sh#L28). run-mongod.sh is running normally and the mongod instance where users are created is running with `--keyFile` (https://github.com/openshift/mongodb/blob/master/2.4/run-mongod.sh#L103). And in MongoDB 2.6 `--keyFile` implies `--auth` so when users are created the authentication is enable. MONGODB_USER is created successfully (https://github.com/openshift/mongodb/blob/master/2.4/contrib/common.sh#L151), but then "localhost exception" is disabled (in 2.6) so connection creating admin user is not authenticated so it fails to create admin user!

rhcarvalho, line 153: @omron93 you are right :)
That actually exposes a problem in the way how we initialize the replica set.
I haven't had the chance to rework that script yet -- one more reason why I'm insisting not to include two examples before we get one right.

For what it's worth, the official docs describe setting up the replica set by starting with a `mongod` with auth disabled. True for both [2.6](https://docs.mongodb.org/v2.6/tutorial/deploy-replica-set-with-auth/) and [2.4](https://docs.mongodb.org/v2.4/tutorial/deploy-replica-set-with-auth/)

We're working on [this Trello card](https://trello.com/c/ZNMI7j81) to make it possible to change passwords. If it would be required to authenticate upfront with a password, we won't be able to implement password updates the way it is proposed.

omron93, line 153: > For what it's worth, the official docs describe setting up the replica set by starting with a mongod with auth disabled. True for both 2.6 and 2.4

Auth is disabled because `--keyFile` is not specified. It is added in point 6 (https://docs.mongodb.org/v2.6/tutorial/deploy-replica-set-with-auth/#start-each-member-of-the-replica-set-with-the-appropriate-options)

rhcarvalho, line 153: Yes, @omron93 you are right. My point is that the current script we have for setting up the replica set don't seem to follow those steps from the official docs. When we create users and set/reset passwords, we should be running with auth disabled, both in standalone and replica set mode.

If we want to include the replica set example for 2.6 the way it is, then this line will have to go the way you wrote it.

But my remarks that this will make implementing the change password flow impossible is still valid.
If you are resetting a password, you don't have the previous password to login to a auth-enabled server...

omron93, line 153: > Yes, @omron93 you are right. My point is that the current script we have for setting up the replica set don't seem to follow those steps from the official docs. When we create users and set/reset passwords, we should be running with auth disabled, both in standalone and replica set mode.

Yes, you are right. But in containers you should exec into `mongod` and have there only one process and it wouldn't be possible to run steps 7-12

> But my remarks that this will make implementing the change password flow impossible is still valid.
> If you are resetting a password, you don't have the previous password to login to a auth-enabled server...

I think it could be possible something like this: Stop master. Start mongod server without `--replSet` and `--keyFile` specified with the same dbpath and then it could be possible to change the password. Then restart mongod with `--replSet` and `--keyFiel` and reconfigure replica set. But I don't try it so I don't know if it works...

bparees:
@omron93 let's go ahead and remove it so we can get this landed.  after @rhcarvalho finishes reworking the 2.4 replica example, we can revisit migrating it to 2.6.

hack/build.sh
```diff
@@ -49,6 +49,9 @@ dirs=${VERSION:-$VERSIONS}
 
 for dir in ${dirs}; do
   IMAGE_NAME="${BASE_IMAGE_NAME}-${dir//./}-${OS}"
+  if [ "$dir" == "2.6" ]; then
+    IMAGE_NAME=rhscl/${BASE_DIR_NAME}-${dir//./}-${OS}
```
bparees, line 53: @rhcarvalho this is ultimately where the rhcl/mongodb-26-centos7 image is going to be coming from..so it's not there today, but it will be there once this gets merged and we build/push it.

however I don't think our push job is currently setup to handle this sort of naming, so that's going to require some changes if we're going to use the existing push job to push to dockerhub under an rhcl namespace instead of openshift.  i'll deal with that when i'm updating the push job to handle mongo 2.6 after this merges, though.

rhcarvalho:
Just a note, if we merge #78 before this, it would be nice to have changes to Dockerfile applied here too. I'll be happy to make the changes myself as part of #78 if this gets merged earlier.

omron93:
Replication example removed. So please merge...

bparees:
[test]

bparees:
@omron93 i'm going to merge this, assuming it passes the test.  can you do a follow on PR to align with the changes made in https://github.com/openshift/mongodb/pull/78 ?

PR #61, opened 2015-07-29
Refactoring scripts and making the image easily extensible 

omron93:
This is mostly a redesign of the scripts. Most of the changes are transparent for the image users, while they are just re-organizing the scripts. The reason for the changes is to allow to extend images easier for user and also to make future maintenance simpler, since one complex bash script would be hard to change in another layer.

These are the changes made:
-  available commands (that are expected to be used as docker run argument) are now called run-mongod, run-mongod-master and those binaries are available under /usr/bin, so users may start them on demand, e.g. when debugging the image
- entrypoint script is now called container-entrypoint
- the entrypoint now serves just for changing environment, nothing more
- configuration files are located in /etc (e.g. /etc/mongod.conf and /etc/mongos.conf)
- images are now more easily extensible, by dropping scripts with further initialization into expected directories (e.g. /usr/share/cont-layer/mongodb/post-init.d)
- container-usage returns help for the container


mfojtik:
[test]

jhadvig:
@omron93 why this PR contains also changes from https://github.com/openshift/mongodb/pull/60 ?

omron93:
@jhadvig It is because that #60 adds new version directory and only changes installed packages and some necessary things. This PR changes code inside 2.4/ and 2.6/ directories. And to be testable it needs to have things from #60 changed too.
I think this is the problem of more commits per one PR. Should I squash them into one commit?

2.4/root/usr/share/cont-docs/10-mongodb.txt
```diff
@@ -0,0 +1,124 @@
...
+**Notice: When mounting directory from host into the container, ensure that the mounted
+directory has the appropriate permissions and that the owner and group of the directory
+matches the user UID or name which is running inside the container.**
+
+
+
+Configuration
+-------------
+
+To configure the image mount MongoDB configuration file into container. So you can mount config file `/etc/mongod.conf` into container by `-e /etc/mongod.conf:/etc/mongod.conf`.
```
bparees: -e?  do you mean -v?  and as an argument to what?

2.4/root/usr/share/cont-docs/10-mongodb.txt
```diff
@@ -0,0 +1,124 @@
...
+| :-------------------- | ------------------------------------------------------------------------- | --------- |
+|  `MONGODB_NOPREALLOC` | Disable data file preallocation.                                          |  true     |
+|  `MONGODB_SMALLFILES` | Set MongoDB to use a smaller default data file size.                      |  true     |
+|  `MONGODB_QUIET`      | Runs MongoDB in a quiet mode that attempts to limit the amount of output. |  true     |
+
+You can also set following mount points by passing `-v /host:/container` flag to docker.
+
+|  Volume mount point      | Description            |
+| :----------------------- | ---------------------- |
+|  `/var/lib/mongodb/data` | MongoDB data directory |
```
bparees, line 101: should /etc/mongodb.conf be listed here?

2.4/root/usr/share/cont-docs/10-mongodb.txt
omron93: You are right. Thanks.

2.4/root/usr/share/cont-docs/10-mongodb.txt
omron93, line 101: `/etc/mongod.conf` is not specified as a volume in Dockerfile. So it is different to `/var/lib/mongodb/data`.
The mongod.conf is mentioned in "configuration" section.

mfojtik:
@omron93 needs rebase

omron93:
@mfojtik Rebased.

omron93:
Changed image name to `rhscl/mongodb-26-$OS`

omron93:
Closing in favour of #112

PR #64, opened 2015-08-03
update maintainer and remove dead code

bparees:


bparees:
@mfojtik ptal pending maintainer

bparees:
argh. merged the wrong tab.  will revert if necessary.

jhadvig:
@bparees the cleanup looks good, but not sure if the `MAINTAINER` addresses are correct. The address for centos image should be `SoftwareCollections.org <sclorg@redhat.com>` and for rhel there should be none (at least thats what I see in the ruby-2.2). And whats with the `dev@lists.openshift.redhat.com` @smarterclayton suggested ?

hhorak:
Please, don't use dev@lists.openshift.redhat.com, the images are supposed to be not only for OpenShift.

bparees:
fixing here: 
https://github.com/openshift/mongodb/pull/65

PR #67, opened 2015-08-07
switch to an arbitrary USER

bparees:


bparees:
@mnagy ptal

mnagy:
Didn't we agree to leave USER as is, and only remove and add comments?

bparees:
I don't recall that but that's not saying much... what would be the reason to do so?

mnagy:
Well, my reasoning was that leaving the user id as is would make it easier and more intuitive to use the image with Docker. I know that isn't our primary use-case, but it won't hurt the OS use-case at all. IIRC, you were arguing that we should make it clear that the images can be used with an arbitrary user, is that right? I think (and was arguing it to Michal) that documenting this fact (in README, OS docs and Dockerfile) would be sufficient.

bparees:
@mnagy setting the user to 1001 doesn't hurt the non-openshift use case either, right?

and yes there were two reasons i wanted to do this:
1) to make it clear the image can be used with an arbitrary user
2) to help prove/test that fact by ensuring the image is never run as a "special" user except by explicit preference.  So by default any testing we do with the image will be using an arbitrary user.  It ensures we're never accidentally relying on any privileges the mongodb user has.

mnagy:
@bparees yes, but it is hurting the Docker use case. I think 2) could be addressed by the tests themselves, but I have to admit that would complicate things and would probably be error prone. /me is convinced

2.4/Dockerfile
```diff
@@ -37,7 +34,6 @@ RUN yum install -y epel-release && \
     yum install -y --setopt=tsflags=nodocs bind-utils gettext iproute v8314 mongodb24-mongodb mongodb24 && \
     yum clean all && \
     mkdir -p /var/lib/mongodb/data && chown -R mongodb.mongodb /var/lib/mongodb/ && \
```
mnagy, line 36: I think we should change chown to use 1001.1001.

mnagy, line 36: This applies to other Dockerfile and other images as well. Also, please remove the `-R`, it is not needed.

bparees, line 36: chowning to 1001.1001 would defeat the purpose of saying "user 1001 isn't special"

mnagy, line 36: Will the image still work without chowning? I'm not sure about it. I think we should be able to run the image without privileges (so no nss switching necessary) and without volume and it should still work. I don't think it will, because it will not be able to write to /var/lib/mongodb.

Also, I'm taking back my comment about `-R`, it clearly is required.

bparees, line 36: @mnagy line 37 below covers write permission to /var/lib/mongodb for everyone else which is why this works for any user.

hhorak:
I'm not convinced. Let's say somebody doesn't specify UID when starting the container (just because not everybody reads doc). Then the data will be readable by 1001 user which may be available on the host. I don't think this is something admins expect nor what we should do. mongodb.mongodb permissions are sane defaults, from my PoV.

bparees:
@hhorak you sure about that?  the /var/lib/volumes directory on the host is root owned with 700 permissions.  I don't think a host user can access files within that directory even if they "own" them on the host.

hhorak:
@bparees I've just did this (96121e765fe1 is image with 1001 as default user):

```
# docker run -ti -v /tmp/datadirmongo:/var/lib/mongodb/data 96121e765fe1 bash
bash-4.2$ touch /var/lib/mongodb/data/somefile
bash-4.2$ exit
```

And found `/tmp/datadirmongo/somefile` with owner 1001:0.

mnagy:
@hhorak I tried the same (without creating the datadir on host) and this happened:

```
[vagrant@openshiftdev mongodb]$ docker run -ti -v /tmp/datadirmongo:/var/lib/mongodb/data openshift/mongodb-24-centos7 bash
bash-4.2$ touch /var/lib/mongodb/data/lala
touch: cannot touch '/var/lib/mongodb/data/lala': Permission denied
```

I've talked to @bparees yesterday and he pointed out that it will only be readable by 1001 user if you chown it to 1001 on host in the first place. Hence this isn't really an issue.

hhorak:
@mnagy when you use 0777 for the dir `/tmp/datadirmongo`, there will be no "Permission denied".

mnagy:
@hhorak yes, but if you chmod to 0777, you know that the directory will be accessible to anyone.

hhorak:
@mnagy the directory itself will be readable for anybody, but not writable by 1001 user. And since files within `/var/lib/mongodb/data` directory use 0600 attrs, they won't be readable for anybody actually. So, what I want to retain is that without specifying any user explicitly, the files will be readable/writeable by mongodb user. That's a standard that we've had in all databases.

bparees:
@hhorak if you chmod the host dir to 0777 then it's certainly writable by uid 1001 inside the container. (and the files will be owned by 1001 on the host, but again you've made an explicit choice to expose those files to everyone since you chmod'd the dir to 777)

In addition there's no particular reason to even assume the host system has a mongodb user.  that's one of the points of containers (that i get my software from the image, not by installing things on my host).  

What's the use case for having the files be owned by "uid 184" on the host?

hhorak:
@bparees when you do `chmod 0777` to main directory (which you pass to -v), you only make explicit that this particular directory is writable for anybody. But nobody can read the data. You may erase a file from that directory, but that's all. And of course wise user won't change directory to 0777, but will rather change owner to mongodb user.

In case we'd go with 1001, the admin would either have to specify `-u` every time (which is weird requirement, since we simply must have sane defaults) or chown the directory to 1001 -- in that case everybody would be wondering why chown directory to such UID? That easily might be my grandmother's UID. Or some less-powerful-admin's.

1001 user was used in cases we don't have better option. For databases we do have better option, because the non-root user already exists.

Also, this case is not that different from the current non-container case. Admin might want to store database files into some arbitrary directory. The only change to do that is preparing the directory (`mkdir data ; chown mongodb.mongodb /data`). I expect most of the users will expect the same steps to be done for mongodb container: `mkdir data ; chown mongodb.mongodb /data` and running mongodb container with `-v /data:/var/lib/mongodb/data`.

bparees:
> 1001 user was used in cases we don't have better option. For databases we do have better option, because the non-root user already exists.

why do you assume that user exists?  there is no reason for me to have a mongodb user on my host system.  In fact i don't have one.  so the files on my host are owned by a mysterious user "184".   And what if at some point the packaging is changed so that the mongodb user is no longer 184?

> But nobody can read the data. You may erase a file from that directory, but that's all. And of course wise user won't change directory to 0777, but will rather change owner to mongodb user.

anyone can read it on the host, assuming you chmod'd the volume dir to 777.  mongodb creates its files with 755 permissions.

> 1001 user was used in cases we don't have better option. For databases we do have better option, because the non-root user already exists.

not on the host it doesn't.

hhorak:
If there is mongodb user, it will always be 184. Nobody will ever change this, it would break too many things. If the user is not existing, that is the same case as if default user with UID 1001 does not exist. Yes, if mongodb user does not exist, it won't be called nice, but it is still better case.

bparees:
@hhorak ok, i'm going to concede on this one.

PR #69, opened 2015-08-12
Several small improvements to the replica README

rhcarvalho:
- Do not reference inexistent .json file
- Update docs of env vars

@bparees @jhadvig PTAL


2.4/examples/replica/README.md
```diff
@@ -2,20 +2,28 @@
...
 
-To create MongoDB replica set in OpenShift v3, you can use the example template
-included in this repository:
+To create a MongoDB replica set in OpenShift v3, you can use the example template
+included in this repository and:
 
-* `oc create -f ./2.4/examples/replica/mongodb-clustered.json`
+* add a new template to the current project:
+```
+$ oc create -f 2.4/examples/replica/mongodb-clustered.json
```
bparees: if they follow these instructions they'll still need to take another action to instantiate the template.  that's not very clear from this.

maybe just delete it and only include the new-app instruction.

rhcarvalho: Ok, that's why I added the other one ;)

How about we change the order, `oc new-app` first, `oc create` last? I see the value of `oc create` in that it allows you to work from the Web Console later.

rhcarvalho: @bparees eventually I went your way, kept only `oc new-app`.
Would get more verbose to highlight that you can use the Web Console, but not only, as after the template is added you can also use `oc new-app --template`.

bparees:
yeah we have docs to cover "all the ways you can do stuff and why", for our examples providing one good way is sufficient.

[merge]

mnagy:
[merge]

PR #75, opened 2015-08-17
[WIP] Replication example with persistent storage

### LINKED ISSUE #114: Replication example with persistent storage
The replication example for MongoDB should use persistent storage ([Trello card](https://trello.com/c/jPGXhxlJ/694-3-mongodb-replica-set-example-with-persistent-storage)).

This is an attempt to bring the conversation from https://github.com/openshift/origin/issues/4211#issuecomment-158940707 closer to where it is to be implemented.

rhcarvalho:
Closes #114.

### DIFF
--- a/2.4/examples/replica/mongodb-clustered.json
+++ b/2.4/examples/replica/mongodb-clustered.json
@@ -55,6 +55,23 @@
   ],
   "objects":[
     {
+      "kind": "PersistentVolumeClaim",
+      "apiVersion": "v1",
+      "metadata": {
+        "name": "mongodb-data-claim"
+      },
+      "spec": {
+        "accessModes": [
+          "ReadWriteOnce"
+        ],
+        "resources": {
+          "requests": {
+            "storage": "2Gi"
+          }
+        }
+      }
+    },
+    {
       "kind": "Service",
       "apiVersion": "v1",
       "metadata": {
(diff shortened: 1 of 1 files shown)


rhcarvalho:
This doesn't work right now because the replicated pods share the same `PersistentVolumeClaim`, but should have independent persistent storage.

2.4/examples/replica/mongodb-clustered.json
```diff
@@ -55,6 +55,23 @@
...
+      "metadata": {
+        "name": "mongodb-data-claim"
+      },
+      "spec": {
+        "accessModes": [
+          "ReadWriteOnce"
+        ],
+        "resources": {
+          "requests": {
+            "storage": "2Gi"
```
mfojtik, line 69: is 2Gi necessary for an example? I mean each Pod will get this right? If I scale this to 10 pods I cut 20Gi from my storage poll.

rhcarvalho, line 69: No. The problem is that PVCs are not "multiplied".  All pods get the same claim, backed by the same storage. Because of that, the current template in this PR does not work.

The value "2Gi" is not the most relevant bit here right now, and we can change it later. I just needed to push this to bring up the discussion on "how to scale PVCs".

rhcarvalho:
Closing this in favor of #184.

PR #78, opened 2015-08-26
Reorder Dockerfile for better use of cache

rhcarvalho:
Every change to the scripts was causing yum install to run over and over.

@bparees PTAL

### DIFF
--- a/2.4/Dockerfile
+++ b/2.4/Dockerfile
@@ -11,7 +11,6 @@ MAINTAINER SoftwareCollections.org <sclorg@redhat.com>
 #  * $MONGODB_DATABASE - Name of the database to create
 #  * $MONGODB_ADMIN_PASSWORD - Password of the MongoDB Admin
 
-
 ENV MONGODB_VERSION=2.4
 
 LABEL io.k8s.description="MongoDB is a scalable, high-performance, open source NoSQL database." \
@@ -20,9 +19,6 @@ LABEL io.k8s.description="MongoDB is a scalable, high-performance, open source N
       io.openshift.tags="database,mongodb,mongodb24"
 
 EXPOSE 27017
-
-COPY run-mongod.sh /usr/local/bin/
-COPY contrib /var/lib/mongodb/
 
 # This image must forever use UID 184 for mongodb user and GID 998 for the
 # mongodb group, so our volumes are safe in the future. This should *never*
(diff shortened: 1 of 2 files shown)


2.4/Dockerfile
```diff
@@ -40,6 +36,9 @@ RUN yum install -y epel-release && \
     test "$(id mongodb)" = "uid=184(mongodb) gid=998(mongodb) groups=998(mongodb)" && \
     chmod o+w -R /var/lib/mongodb && chmod o+w -R /opt/rh/mongodb24/root/var/lib/mongodb
 
+COPY run-mongod.sh /usr/local/bin/
+COPY contrib /var/lib/mongodb/
```
bparees, line 41: same comment as on the other 2 prs

2.4/Dockerfile
```diff
@@ -38,7 +34,14 @@ RUN yum install -y epel-release && \
...
     test "$(id mongodb)" = "uid=184(mongodb) gid=998(mongodb) groups=998(mongodb)" && \
-    chmod o+w -R /var/lib/mongodb && chmod o+w -R /opt/rh/mongodb24/root/var/lib/mongodb
+    # Loosen permission bits to avoid problems running container with arbitrary UID
+    chmod -R a+rwx /opt/rh/mongodb24/root/var/lib/mongodb
+
+COPY run-mongod.sh /usr/local/bin/
+COPY contrib /var/lib/mongodb/
+
+# Loosen permission bits to avoid problems running container with arbitrary UID
+RUN chmod -R a+rwx /var/lib/mongodb
```
rhcarvalho, line 44: @bparees added this extra RUN line like in the PR for PostgreSQL.
In hindsight, I'm unsure about `a+rwx`, in particular setting the executable bit in every file...
Maybe `a+rw` would be better, leaving the executable bit only for files that where already executable...? But then, would be good to make them executable by all (if they were not already).

bparees, line 44: yeah i'm not sure the best approach here either.  i don't love making non executable files executable, but it'd be worse to leave a file that should be executable, not able to be executed (though in theory a user could make a copy of it and make the copy executable).

i think for now a+rwx is the safer course of action, but long term we need a better solution for permissions in general.

bparees:
[test]

2.4/Dockerfile
rhcarvalho, line 44: Note that `run-mongod.sh`'s permissions are inherited from what is stored in git. That's one file that we don't set permissions explicitly, and it sure need to be executable.

bparees:
[test]

PR #87, opened 2015-09-17
Less loose permission bits for /var/lib/mongodb inside the container

hhorak:
This is similar to https://github.com/openshift/mysql/commit/f44b12dfb67bbda8a5e14341313c716c8a303e09


mfojtik:
LGTM

PR #89, opened 2015-09-24
Fix description of user creation in README

omron93:
In README there is:

> the database will be created together with the database administrator user and MongoDB admin user

I think "admin user" and "database administrator user" is the same.


mfojtik:
LGTM

PR #91, opened 2015-10-08
Specify HOME variable

omron93:
Add correct HOME variable.

It fixes bugs like this -  https://bugzilla.redhat.com/show_bug.cgi?id=1265976

### DIFF
--- a/2.4/Dockerfile
+++ b/2.4/Dockerfile
@@ -11,7 +11,8 @@ MAINTAINER SoftwareCollections.org <sclorg@redhat.com>
 #  * $MONGODB_DATABASE - Name of the database to create
 #  * $MONGODB_ADMIN_PASSWORD - Password of the MongoDB Admin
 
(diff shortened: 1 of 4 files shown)


mfojtik:
LGTM [test]

jhadvig:
@omron93 wants there a bug in mongo24 where the HOME for the mongo user was wrong ?

jhadvig:
https://bugzilla.redhat.com/show_bug.cgi?id=1206151

omron93:
@jhadvig What is wrong about HOME=/var/lib/mongodb?
Even in the mentioned bug you wants this change. And this mainly solves that when you run with "random" UID (docker run -u ...) the HOME is set to / and user can't write there.

jhadvig:
@omron93 nothings wrong, the change itself is good :+1: 
Just wanted to point out the reason, so when we look back, we know why the change was done(partially)

PR #98, opened 2015-10-20
Re-ensure that hosts are ready before rs.initiate

grdryn:
We've noticed an intermittent issue where sometimes the rs.initiate
doesn't work successfully. This seems to be because not all of the hosts
discovered just beforehand in the call to build_mongo_config, are ready.

This _should_ be handled by the waiting that occurred in
initiate_replica.sh, but it seems that sometimes not all of the hosts
are up at that point (not all IPs are discoverable). The later
call (endpoints function) that happens in build_mongo_config then
sometimes finds more, but they're not checked to be ready.

This change performs the wait in two places: the original place (in case
there was some reason that it needed to happen before something else
there), and right before the call to rs.initiate. This should ensure
that any new hosts discovered in the build_mongo_config are ready for
replication.

### DIFF
--- a/2.4/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.4/root/usr/share/container-scripts/mongodb/common.sh
@@ -93,6 +93,17 @@ function no_endpoints() {
   [ "$(endpoints)" == "$(container_addr)" ]
 }
 
+# wait_for_all_hosts waits until all hosts found using endpoints function are
+# ready.
+function wait_for_all_hosts() {
+    local current_endpoints=$(endpoints)
+
+    echo "=> Waiting for all endpoints to accept connections..."
+    for node in ${current_endpoints}; do
+        wait_for_mongo_up ${node} &>/dev/null
+    done
+}
+
 # build_mongo_config builds the MongoDB replicaSet config used for the cluster
 # initialization
 function build_mongo_config() {
@@ -112,6 +123,7 @@ function mongo_initiate() {
   local mongo_wait="while (rs.status().startupStatus || (rs.status().hasOwnProperty(\"myState\") && rs.status().myState != 1)) { printjson( rs.status() ); sleep(1000); }; printjson( rs.status() );"
   config=$(build_mongo_config)
   echo "=> Initiating MongoDB replica using: ${config}"
+  wait_for_all_hosts
   mongo admin --eval "${config};rs.initiate(config);${mongo_wait}"
 }
 
--- a/2.4/root/usr/share/container-scripts/mongodb/initiate_replica.sh
+++ b/2.4/root/usr/share/container-scripts/mongodb/initiate_replica.sh
@@ -15,10 +15,7 @@ done
 current_endpoints=$(endpoints)
 mongo_node="$(echo -n ${current_endpoints} | cut -d ' ' -f 1):${CONTAINER_PORT}"
 
-echo "=> Waiting for all endpoints to accept connections..."
-for node in ${current_endpoints}; do
-  wait_for_mongo_up ${node} &>/dev/null
-done
+wait_for_all_hosts
 
 echo "=> Initiating the replSet ${MONGODB_REPLICA_NAME} ..."
 # Start the MongoDB without authentication to initialize and kick-off the cluster:
--- a/2.6/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/common.sh
@@ -93,6 +93,17 @@ function no_endpoints() {
   [ "$(endpoints)" == "$(container_addr)" ]
 }
 
+# wait_for_all_hosts waits until all hosts found using endpoints function are
+# ready.
+function wait_for_all_hosts() {
+    local current_endpoints=$(endpoints)
+
+    echo "=> Waiting for all endpoints to accept connections..."
+    for node in ${current_endpoints}; do
+        wait_for_mongo_up ${node} &>/dev/null
+    done
+}
+
 # build_mongo_config builds the MongoDB replicaSet config used for the cluster
 # initialization
 function build_mongo_config() {
@@ -112,6 +123,7 @@ function mongo_initiate() {
   local mongo_wait="while (rs.status().startupStatus || (rs.status().hasOwnProperty(\"myState\") && rs.status().myState != 1)) { printjson( rs.status() ); sleep(1000); }; printjson( rs.status() );"
   config=$(build_mongo_config)
   echo "=> Initiating MongoDB replica using: ${config}"
+  wait_for_all_hosts
   mongo admin --eval "${config};rs.initiate(config);${mongo_wait}"
 }
 
--- a/2.6/root/usr/share/container-scripts/mongodb/initiate_replica.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/initiate_replica.sh
@@ -15,10 +15,7 @@ done
 current_endpoints=$(endpoints)
 mongo_node="$(echo -n ${current_endpoints} | cut -d ' ' -f 1):${CONTAINER_PORT}"
 
-echo "=> Waiting for all endpoints to accept connections..."
-for node in ${current_endpoints}; do
-  wait_for_mongo_up ${node} &>/dev/null
-done
+wait_for_all_hosts
 
 echo "=> Initiating the replSet ${MONGODB_REPLICA_NAME} ..."
 # Start the MongoDB without authentication to initialize and kick-off the cluster:


grdryn:
There's still a case where this can "fail" that I'm looking into (and I've seen twice now), it seems to be the following: sometimes even at the second check that I've added here, there's a possibility that not all of the pods in the service have an IP address, so don't get 'discovered'.
In this case, for the ones that are discovered, it will wait until mongo us up and listening on those & then start initiation of the replica set. When that's successful, and the initiator pod drops out, the remaining pods are left to pick a new master. Sometimes, they haven't got a majority, so can't vote for a new master.

E.g. if we want 3 mongo instances in our replica set, but only 2 (or fewer) are there at initiation - the first iteration of the replica set will be those two as secondaries and the initiator as primary. Then when the posthook completes and that pod goes away, there's another one ready (but not yet configured for replication) - that's 4 potential members, only two of which are voting members, so no majority.

I might be wrong, but this is my current understanding of what's happening. I can provide logs if anybody would like to see! :)

So I think what needs to happen is that the posthook pod waits until there are more than at least half of the final intended members + 1 of the proper pods (i.e. not counting itself) available before starting initiation.

mfojtik:
@omron93 can you please review this PR?

mfojtik:
LGTM

bparees:
needs a rebase and equivalent changes to the 2.6 image.

2.4/contrib/initiate_replica.sh
```diff
@@ -15,10 +15,7 @@ done
 current_endpoints=$(endpoints)
 mongo_node="$(echo -n ${current_endpoints} | cut -d ' ' -f 1):${CONTAINER_PORT}"
 
-echo "=> Waiting for all endpoints to accept connections..."
-for node in ${current_endpoints}; do
-  wait_for_mongo_up ${node} &>/dev/null
-done
+wait_for_all_hosts
```
omron93: I think it is redundant to wait_for_all_hosts twice. Several lines after this there is called mongo_initiate and in this function there is also called wait_for_all_hosts.
So I think wait_for_all_hosts in 2.4/contrib/initiate_replica.sh is not necessary.

omron93:
@mfojtik @bparees I know that it might take different time to start mongod daemon (time from running `mongod $mongo_common_args ...` to time when this mongod accepts connections), is it also possible that some containers are started late in kubernetes? Does `dig ${service_name} A +search +short 2>/dev/null` depend on this - does this needs to have the container running or does it take info from Deployment config,...? Is it somehow possible get information that all planned pods are running?

mfojtik:
@omron93 it must have the Pod created AFAIK, not container. It might take some time especially when you have a lot of replicas defined. The `dig` is OK, but it is not bullet-proof, I think for managing MongoDB replication in production, you will need to have some 'etcd' sidecar container that will actually manage the replicas.

grdryn:
Howdy! I've just started looking at this again (sorry for taking so long to get back to it).

I've rebased and mirrored the changes to the config for 2.6, as @bparees asked.

@omron93 I've left in the two checks for now. I agree that they may be redundant, but if one is to be removed, I think it should be the other one that's removed, leaving the one that happens latest. This will give slightly more time for things to become ready. Yes, this is admittedly racy!

@mfojtik Do you mean something like https://github.com/leportlabs/mongo-k8s-sidecar?

I've also been trying to think about how to make this less racy, and cover the other scenario that I mentioned that can still happen. As I said, I think we need at least half+1 of the intended pods to be available, but how do we know the number of intended pods? I think one way could be for the [posthook definition in the template](https://github.com/openshift/mongodb/blob/master/2.4/examples/replica/mongodb-clustered.json#L91..L98) to pass an additional parameter with the number of replicas (pull the number out to a parameter and reuse it here and in the [replicas count](https://github.com/openshift/mongodb/blob/master/2.4/examples/replica/mongodb-clustered.json#L107) or maybe). Any thoughts on that?

Thanks,
Gerard.

bparees:
@rhcarvalho since you suddenly have some spare time, can you take a look at this and if it looks good, run it through the mongo replication extended tests (assuming those are in a runnable state)?

thanks

mfojtik:
@bparees this should kick all "mongo" extended tests :-)

[test][openshift]

bparees:
@mfojtik I don't think the mongo replication extended test is merged yet.

rhcarvalho:
> run it through the mongo replication extended tests (assuming those are in a runnable state)

There are no extended tests yet. Open PR: https://github.com/openshift/origin/pull/6197.
I can check this patch manually, though.

bparees:
@rhcarvalho bump

rhcarvalho:
This might be related / fix https://bugzilla.redhat.com/show_bug.cgi?id=1264048

grdryn:
Hi @rhcarvalho,

This hack would at least help with that, but that could still happen. What we're thinking for us in production is having an env var passed to the initiator to tell it how many replicas there should be, so that it can wait for all of them. This will also allow us to have a separate DC for each of the replicas so that we can have a distinct PVC per replica (kinda like #75) (we already have split to separate DCs and have persistent storage, but now we have the initiator separate (not a lifecycle hook) and we see issues more often).

rhcarvalho:
@grdryn would it still help to get this patch in?

grdryn:
@rhcarvalho I would say yes, as an intermediate quick fix until there's something better done. It 'helps', but doesn't fully 'solve' the problem. Does the kubernetes proposal that you link to in https://github.com/openshift/origin/issues/4211#issuecomment-171368947 would have any impact on a final solution here (disclaimer: I haven't ready the proposal so I don't fully understand what it's about)?

rhcarvalho:
> Does the kubernetes proposal that you link to in openshift/origin#4211 (comment) would have any impact on a final solution here

Yes it would, it would allow for a single DC with independent storage per replica.

The Bugzilla linked above is a bug when running the replication example in a multi node setup. I think it is similar to this because apparently the underlying problem was trying to setup the cluster without knowing all of the members, likely caused by slow DNS updates.

But if the problem would be indeed the DNS not reporting all IPs, this patch here wouldn't actually help that scenario, as we'd still be waiting for just part of the members.

I'm gonna do some manual tests with your changes, code-wise looks alright.

grdryn:
I've never seen an issue with DNS (even on multi-node). From [the paste](http://ur1.ca/ntyjd) that Wang gave in that BZ, I would rather suggest that the extraneous IP that isn't resolvable was the IP from the initiator (exited), and that only 2/3 replica pods were ready at initiation. This could be verified if the logs from the initiator were available (likely not, that bug is old). I have seen that sometimes if a member is not contactable for any reason, that it doesn't always get cleared from the replica set member list.

rhcarvalho:
In reading your PR and following up with reading the surrounding code, it turns out that we're calling `dig ...` an absurd amount of times, most of them unnecessarily.

I'm short on time now, but later today or tomorrow I'll open a PR to remove those unnecessary calls, will tag you there. I've got the code locally, but still need to test it.

My changes would guarantee that the list of endpoints passed to the config is the same as the list of endpoints we wait for before we initiate the setup. WDYT?

grdryn:
> My changes would guarantee that the list of endpoints passed to the config is the same as the list of endpoints we wait for before we initiate the setup. WDYT?

I don't fully understand which config you're referring to here. Is that some config that the initiator uses? Is that env vars the template or something else?

rhcarvalho:
I meant the config in `build_mongo_config`. In fact, I believe there should be a single call to discover endpoints, and later we should just reuse the value in function calls. Messy bash :/

bparees:
@grdryn:  @rhcarvalho and I chatted a bit on IRC and I believe what he's suggesting is that the actual root cause of this issue is that we are invoking dig multiple times, resulting in multiple different results which are in conflict.

If instead we retrieved the list of a hosts exactly once and used that same list consistently, we would not have to do a final re-check at the end.

rhcarvalho:
@grdryn could you have a look at recent merged PRs, I believe they might be helpful:
- #124 Dig only once -- there will be a single list of IPs to wait for (this combined with readiness and liveness probes can be valuable)
- #130 and #133 Fix getting of primary member -- fixes logic to determine the primary
- #132 Fix race condition

With those in, I believe we don't need to wait for all nodes multiple times, wdyt?

grdryn:
> With those in, I believe we don't need to wait for all nodes multiple times, wdyt?

Waiting multiple times does seem redundant, and as I hopefully mentioned before, this was a hack to try to make it work a little more often. Unfortunately I don't have the time right now to look through and understand the PRs that you mentioned, but I know you're far more familiar with all of this than I am, so I'll trust your judgement that this can be closed. :+1: 

I did take a quick look at the current master though:
Where we [get the current_endpoints at initiation](https://github.com/openshift/mongodb/blob/master/2.4/root/usr/share/container-scripts/mongodb/initiate_replica.sh#L11), if some of the pods don't yet have IPs, will they get added to the replica set automatically after (or even during) initiation? Or do we otherwise ensure that we have all (or 'enough') pods available before continuing? (the latter is the latest hack that my team are currently using -- we've added an env var from our Openshift template with the number of MongoDB replica pods that we know we're going to have at creation time, so that initiation waits until that many endpoints are discovered).

Another scenario (if members do get added dynamically when they're ready for replication): what if a minority were available when initiation starts, but all are ready when [the initiator steps down as PRIMARY](https://github.com/openshift/mongodb/blob/master/2.4/root/usr/share/container-scripts/mongodb/initiate_replica.sh#L50) -- will it be possible to elect a new PRIMARY? If I'm correct, then only the minority that were initiated can vote, but they're a minority of the total set, so may not be able to elect a new PRIMARY.

I'm not sure if all of the above still makes sense, given that there are lots of new changes in the scripts :)

rhcarvalho:
> if some of the pods don't yet have IPs, will they get added to the replica set automatically after (or even during) initiation?

Theoretically, we should have all IPs before we begin. If in practice that's not happening, that's a point to be fixed.

The hack of waiting for a fixed number might help to "give the server more time to fix itself", but ideally I would like us to find a proper solution that doesn't depend on increasing sleep times and retries at multiple places :)

> will it be possible to elect a new PRIMARY?

That's an interesting scenario. Discussing with @php-coder today, he brought the idea that we don't really need the pod running the hook to ever be part of the replica set, and then we won't need a re-election. That's something we should investigate (https://github.com/openshift/mongodb/issues/140).

@grdryn thanks very much for bringing this up. I think we can close this PR (I don't have rights to do it), it will stay as a reference point for the future.

grdryn:
Cool, I'll close it now so. I'm glad it was useful! :)

> Theoretically, we should have all IPs before we begin. If in practice that's not happening, that's a point to be fixed.

That may indeed be fixed at this stage. It was an issue for us a while back, but given that we've got our hack in place now, of course we wouldn't see it.

PR #100, opened 2015-11-05
sync container principles with mysql

omron93:


mfojtik:
[test]

hhorak:
what a weird error.. `Failed to open "/proc/4069/ns": No such file or directory` is it possible that docker was updated?

hhorak:
Hm, I've reproduced it, so it will be problem in test probably..

hhorak:
We've found the problem probably, fix is coming.

mfojtik:
[test] again

hhorak:
@mfojtik the fix is not there yet, so it will fail again :) the problem is with unbound variable, since there is `$1` referenced when not being set..

omron93:
@mfojtik It should be fixed now. Sorry for a delay, there were more things to fix than I've expected.

hhorak:
[test]

hhorak:
I don't see anything wrong here, LGTM.

2.4/Dockerfile
```diff
@@ -38,22 +38,26 @@ RUN yum install -y epel-release && \
     # Loosen permission bits to avoid problems running container with arbitrary UID
     chmod -R a+rwx /opt/rh/mongodb24/root/var/lib/mongodb
 
-COPY run-mongod.sh /usr/local/bin/
-COPY contrib /var/lib/mongodb/
-
 # Loosen permission bits to avoid problems running container with arbitrary UID
-RUN chmod -R g+rwx /var/lib/mongodb
+RUN chmod -R g+rwx /var/lib/mongodb && touch /etc/mongod.conf && chown mongodb.root /etc/mongod.conf && chmod og+rw /etc/mongod.conf
```
mfojtik: use the fix-permissions.sh script here. also does this mean the /etc/mongod.conf is now writeable by the user that runs the container?

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -4,14 +4,14 @@
 MAX_ATTEMPTS=60
 SLEEP_TIME=1
 
-export MONGODB_CONFIG_PATH=/var/lib/mongodb/mongodb.conf
-export MONGODB_PID_FILE=/var/lib/mongodb/mongodb.pid
-export MONGODB_KEYFILE_PATH=/var/lib/mongodb/keyfile
+export MONGODB_CONFIG_PATH=/etc/mongod.conf
+export MONGODB_PID_FILE=${HOME}/mongodb.pid
```
mfojtik: we were avoiding ${HOME} and rather use the absolute path in case you use different user and the HOME can be set to something unpredictable.

2.4/Dockerfile
hhorak: fix-permissions.sh is defined in s2i-base which is not what this image is using, do you propose to copy that script just to avoid those two commands?

mfojtik: @hhorak I think it would be safe to import that script here (I did that for jenkins image). That script can be re-used in future and provide consistent way to set permissions (think that in future we might want to do some chcon there, etc..)

2.4/root/usr/share/container-scripts/mongodb/common.sh
hhorak: yeah, makes sense.

2.4/Dockerfile
hhorak: @mfojtik I've just checked how fix-pemissions.sh works and now it is designed to work only on directories. I'd rather extend the script to work on files (including touch call probably) and integrate it here later.

hhorak: (just for sake of not do more harm than necessary :) )

omron93: @mfojtik /etc/mongod.conf have to be writeable because run-mongod after substitution of env. variables have to write actual config there. In openshift version it is writable too, but it is in HOME directory...

Now using fix-permission script. Please check current status...

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: fixed

hhorak:
[test]

mfojtik:
LGTM ( @bparees can you review this pls?)

bparees:
lgtm.

hhorak:
@bparees @mfojtik @omron93 Thanks guys!

bparees:
@omron93 @hhorak do we need the run-mongodb.sh symlink for any particular reason?

hhorak:
@bparees I thought it was needed to keep compatibility, but then I learned it should be enough to change it in the replica example, so I'm fine with removing the symlink in the end.

bparees:
>  @bparees I thought it was needed to keep compatibility, but then I learned it should be enough to change it in the replica example, so I'm fine with removing the symlink in the end.

cool, i'll do a pull and remove it.

bparees:
@hhorak @omron93 @mfojtik fyi something about this PR is causing the image to not successfully get pushed in our push job.  I'm still investigating since the tests seem to pass, but it seems like maybe a spurious "exit" is getting invoked that is screwing up our vagrant plugin.  I'm guessing it has to do w/ the changes to the test/run script.  Anyway just a heads up that:
1) the updates are not pushed to dockerhub
2) you may or may not have problems w/ this image if you try to build the current content in OSBS (if the issue is just w/ the test script, obviously you won't have problems)

bparees:
@hhorak @omron93 @mfojtik I spent an inordinate amount of time on this this weekend but the net is something about the new docker exec introduced in this PR that's being done in the test/run script is leaving things in a bad state such that our vagrant steps fail out rather than push the images, after the test appears to complete successfully.

hhorak:
@bparees As for the `docker exec` call in the test, it uses `-i` that might be needless here, wouldn't removing it help by any chance?

mfojtik:
@hhorak the `-i` is indeed not needed, it is used only when you are piping in something to container... and it might screw our tests if Jenkins is using some weirdo tty/terminal setup. Can you try to remove that option?

omron93:
@mfojtik I've created this PR - #107

bparees:
@mfojtik I got to the point of trying removing the "-i" shortly before i went to bed last night, but i think my environment was highly suspect by then.  But yes I was wondering the same thing.  Basically something appears to be prematurely closing the ssh session that vagrant started, after "make test" completes successfully, but before the push commands start running (both the make and push are commands sent to the same shared ssh session)

2.4/Dockerfile
```diff
@@ -38,22 +38,29 @@ RUN yum install -y epel-release && \
...
 
-COPY run-mongod.sh /usr/local/bin/
-COPY contrib /var/lib/mongodb/
-
 # Loosen permission bits to avoid problems running container with arbitrary UID
 RUN chmod -R g+rwx /var/lib/mongodb
 
+# Get prefix path and path to scripts rather than hard-code them in scripts
+ENV CONTAINER_SCRIPTS_PATH=/usr/share/container-scripts/mongodb \
+    MONGODB_PREFIX=/opt/rh/mongodb24/root/usr \
```
rhcarvalho, line 46: @omron93 do you remember why do we have this env var and what it does? I see it was added in 3.2, then replicated here, but it is not referenced anywhere outside of the Dockerfiles.

omron93: @rhcarvalho It was added to provide similar env variables as mysql-container (to make evn "same" in all docker images). Other users does not have it, so I think it could be removed now.

rhcarvalho: Okay, I see that in mysql MYSQL_PREFIX is used in scripts and config. Thanks!

PR #107, opened 2015-11-09
Fixed test/run and syncing 2.4 and 2.6 test/run

omron93:


mfojtik:
[test]

2.6/test/run
```diff
@@ -14,6 +14,7 @@ IMAGE_NAME=${IMAGE_NAME-centos/mongodb-26-centos7-candidate}
 CIDFILE_DIR=$(mktemp --suffix=mongodb_test_cidfiles -d)
 
 function cleanup() {
+    set +e
```
mfojtik: why you need this for cleanup? This means that when we fail remove one cidfile, we exit and keep the other cidfiles around

2.6/test/run
```diff
@@ -143,17 +145,30 @@ function run_configuration_tests() {
...
+
 function create_container() {
     local name=$1 ; shift
     local cargs=${CONTAINER_ARGS:-}
-    cidfile="$CIDFILE_DIR/$name"
+    cid_file="$CIDFILE_DIR/$name"
     # create container with a cidfile in a directory for cleanup
-    docker run --cidfile $cidfile -d $cargs "$@" $IMAGE_NAME
-    echo "Created container $(cat $cidfile)"
+    docker run --cidfile $cid_file -d $cargs "$@" $IMAGE_NAME
```
mfojtik, line 167: i wish for the future that we will give containers names so when the test fail and we are capturing the logs from all containers, we can query the container name and name the log file against the container name. It will help identifying failed containers :-)

omron93:
@mfojtik I hope @bparees can better explain your questions - this PR only sync changes which are in 2.4 only to 2.6 - this PR: #103 #104 #105 #106

2.6/test/run
bparees: this was something i put in over the weekend in an attempt to debug what was going on, before i got the point of running the push job against my own fork.  agree it is probably not needed, should not be in 2.4 either.

2.6/test/run
```diff
@@ -24,6 +25,7 @@ function cleanup() {
         echo "Done."
     done
     rmdir $CIDFILE_DIR
+    set -e
```
bparees: same with this of course.

omron93: Removed from 2.4 and 2.6 also

2.6/test/run
omron93: Removed from 2.4 and 2.6 also

bparees:
[test]

PR #108, opened 2015-11-09
Fix test suite and limit resources consumed

hhorak:
I've looked closely at the test suite and saw some potential issues. This PR should make it more reliable to pass. Using smallfiles then prevents to fail just because of little disc space.

### DIFF
--- a/2.4/test/run
+++ b/2.4/test/run
@@ -47,13 +47,13 @@ function mongo_admin_cmd() {
 
 function mongo_cmd_local() {
     local name="$1" ; shift
-    docker exec -i $(get_cid $name) bash -c mongo "$@"
+    docker exec $(get_cid $name) bash -c mongo "$@"
 }
 
 function test_connection() {
     local name=$1 ; shift
-    ip=$(get_container_ip $name)
-    echo "  Testing MongoDB connection to $ip..."
+    CONTAINER_IP=$(get_container_ip $name)
+    echo "  Testing MongoDB connection to $CONTAINER_IP..."
     local max_attempts=20
     local sleep_time=2
     for i in $(seq $max_attempts); do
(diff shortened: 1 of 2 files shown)


2.4/test/run
```diff
@@ -128,10 +128,19 @@ function test_config_option() {
 -e MONGODB_USER=user
 -e MONGODB_PASSWORD=password
 -e MONGODB_ADMIN_PASSWORD=adminPassword
+-e MONGODB_SMALLFILES=1
```
omron93: MONGODB_SMALLFILES should default to true in run-mongod. Doesn't this work?

2.4/test/run
```diff
@@ -128,10 +128,19 @@ function test_config_option() {
...
 -e $env_var=${value//\'/}
 "
     create_container ${name} ${DOCKER_ARGS}
 
+    # need to set these because `mongo_cmd` relies on global variables
+    USER=user
+    PASS=password
+    DB=db
+
+    CONTAINER_IP=$(get_container_ip $name)
```
omron93: In 'test_connection' there is also used function 'get_container_ip', so I suggest to remove remove this call from 'test_connection' and to use there 'CONTAINER_IP' instead of 'ip'. Or to define 'CONTAINER_IP' in 'test_connection' (change 'ip' to 'CONTAINER_IP') and not to define 'CONTAINER_IP' before calling 'test_connection'.

hhorak:
Thanks @omron93, I've reverted the unnecessary changes of SMALLFILES and did the proposed change with CONTAINER_IP.

omron93:
Thanks @hhorak 

LGTM

mfojtik:
LGTM

PR #110, opened 2015-11-12
remove test for gid value

### LINKED ISSUE #109: GID 998 will be unusable in rhel7.2
Per https://github.com/openshift/mongodb/blob/master/2.4/Dockerfile.rhel7#L28-L32 mongodb user must always be uid=184 and gid=998   however in rhel7.2 images gid 998 is already in use by systemd packages.

```
docker run ${INTERNAL_REGISTRY}/rhel7.2 cat /etc/group | grep 998
systemd-bus-proxy:x:998:
```

When installing on rhel7.2 mongodb is uid=184 gid=995
So what to do? 

@jhadvig as you wrote the comment referenced above

bparees:
fixes https://github.com/openshift/mongodb/issues/109

### DIFF
--- a/2.4/Dockerfile
+++ b/2.4/Dockerfile
@@ -21,9 +21,6 @@ LABEL io.k8s.description="MongoDB is a scalable, high-performance, open source N
 
 EXPOSE 27017
 
-# This image must forever use UID 184 for mongodb user and GID 998 for the
-# mongodb group, so our volumes are safe in the future. This should *never*
-# change, the last test is there to make sure of that.
 # Due to the https://bugzilla.redhat.com/show_bug.cgi?id=1206151,
 # the whole /var/lib/mongodb/ dir has to be chown-ed.
 RUN yum install -y epel-release && \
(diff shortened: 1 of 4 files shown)


bparees:
@mnagy @rhcarvalho ptal.

bparees:
[test]

bparees:
@sdodson fyi.

rhcarvalho:
@bparees RHEL build failed.
Could we do the changes to the chmod in a separate PR or at least a separate commit?
That change has a different reason to be, and IMHO would be better off on its own (e.g. in case we need to revert it, or for clarity when somebody is reading the commit message and trying to understand why those lines changed).
Other than that LGTM.

bparees:
[test]

bparees:
@rhcarvalho I don't see any reason for that chmod to be its own RUN layer and i don't see a need for a separate commit or PR for a trivial change like that.

the failure was because of a cut+paste issue in the 2.6 images, fixed.

mnagy:
LGTM

PR #111, opened 2015-11-12
Remove UID test

### LINKED ISSUE #109: GID 998 will be unusable in rhel7.2

mnagy:
Fixes #109

### DIFF
--- a/2.4/Dockerfile
+++ b/2.4/Dockerfile
@@ -21,9 +21,6 @@ LABEL io.k8s.description="MongoDB is a scalable, high-performance, open source N
 
 EXPOSE 27017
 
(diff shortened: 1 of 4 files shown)


mnagy:
Closing in favour of #110

PR #121, opened 2016-01-12
Verify all installed packages using rpm -V

mfojtik:


mfojtik:
[test]

php-coder:
Just curious: what problem it will solve? Why do we need to verify packages that just have installed?

mfojtik:
@php-coder yes, to make sure we install them all. yum will not return error when one package fail to install

php-coder:
> yum will not return error when one package fail to install

In this case we could use simpler command: `rpm -q foo bar >/dev/null`.
In any case I would rather add comment about why do we need this command. IMHO.

mfojtik:
since you're third guy who is asking this, I agree :-)

On Thu, Feb 4, 2016 at 5:46 PM, Vyacheslav Semushin <
notifications@github.com> wrote:

> yum will not return error when one package fail to install
> 
> In this case we could use simpler command: rpm -q foo bar >/dev/null.
> In any case I would rather add comment about why do we need this command.
> IMHO.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/openshift/mongodb/pull/121#issuecomment-179939653.

## 

---

Michal Fojtik <mi@mifo.sk <mfojtik@mifo.sk>>
Red Hat OpenShift, Engineering

rhcarvalho:
@mfojtik bump ;)

PR #120, opened 2016-01-08
add readinessProbe

wanghaoran1988:
Add a readinessProbe for this template

PR #124, opened 2016-02-10
Dig only once

rhcarvalho:
Reducing the number of calls to `dig`.

In particular, making sure we use the same list of endpoints when configuring and initializing the replica set.

Since we call `wait_for_mongo_up` for all endpoints in `initiate_replica.sh`, we ensure we have waited for all known endpoints before a call to `rs.initiate`, so this should fix what #98 was trying to solve.

### DIFF
--- a/2.4/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.4/root/usr/share/container-scripts/mongodb/common.sh
@@ -92,15 +92,18 @@ function endpoints() {
 }
 
 # build_mongo_config builds the MongoDB replicaSet config used for the cluster
-# initialization
+# initialization.
+# Takes a list of space-separated member IPs as the first argument.
 function build_mongo_config() {
+  local current_endpoints
+  current_endpoints="$1"
   local members
   members="{ _id: 0, host: \"$(mongo_addr)\"},"
   local member_id
   member_id=1
   local container_addr
   container_addr="$(container_addr)"
-  for node in $(endpoints); do
+  for node in ${current_endpoints}; do
     [ "$node" == container_addr ] && continue
     members+="{ _id: ${member_id}, host: \"${node}:${CONTAINER_PORT}\"},"
     let member_id++
@@ -108,11 +111,12 @@ function build_mongo_config() {
   echo -n "var config={ _id: \"${MONGODB_REPLICA_NAME}\", members: [ ${members%,} ] }"
 }
 
-# mongo_initiate initiate the replica set
+# mongo_initiate initiates the replica set.
+# Takes a list of space-separated member IPs as the first argument.
 function mongo_initiate() {
   local mongo_wait
   mongo_wait="while (rs.status().startupStatus || (rs.status().hasOwnProperty(\"myState\") && rs.status().myState != 1)) { printjson( rs.status() ); sleep(1000); }; printjson( rs.status() );"
-  config=$(build_mongo_config)
+  config=$(build_mongo_config "$1")
   echo "=> Initiating MongoDB replica using: ${config}"
   mongo admin --eval "${config};rs.initiate(config);${mongo_wait}"
 }
@@ -153,9 +157,9 @@ function mongo_remove() {
   local mongo_addr
   mongo_addr="$(mongo_addr)"
 
-  echo "=> Removing $mongo_addr on $primary_addr ..."
+  echo "=> Removing ${mongo_addr} on ${primary_addr} ..."
   mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" \
-    --host "$primary_addr" --eval "rs.remove('$mongo_addr');" || true
+    --host "${primary_addr}" --eval "rs.remove('${mongo_addr}');" || true
 }
 
 # mongo_add advertise the current container to other mongo replicas
@@ -166,9 +170,9 @@ function mongo_add() {
   local mongo_addr
   mongo_addr="$(mongo_addr)"
 
-  echo "=> Adding $mongo_addr to $primary_addr ..."
+  echo "=> Adding ${mongo_addr} to ${primary_addr} ..."
   mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" \
-    --host "$primary_addr" --eval "rs.add('$mongo_addr');"
+    --host "${primary_addr}" --eval "rs.add('${mongo_addr}');"
 }
 
 # run_mongod_supervisor runs the MongoDB replica supervisor that manages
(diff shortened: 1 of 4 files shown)


rhcarvalho:
@bparees @mfojtik @omron93 @grdryn please take a look

rhcarvalho:
I've included one commit where I'm also reducing the number of steps/queries to connect to the primary to add / remove members. Mongo has a special syntax to connect to a replica set that automatically connect to the primary.

rhcarvalho:
While testing this I ran into #125. If we're already broken, I'm not sure if my changes carry any harm.

bparees:
I don't know if this is a new problem introduced by your change or not, but:  Is there any guarantee that before the post-deploy hook runs, all the replicas are added as endpoints to the service?

if not, we're not going to get the right membership list.

And I don't think it is guaranteed, I think the deployment is considered success once 90% of the replicas are ready, not 100%.

Otherwise the changes look sane and beneficial to me.

rhcarvalho:
> I don't know if this is a new problem introduced by your change or not,

The bug I reported was running the image from Docker Hub.
I deleted all locally built images, docker pulled, and setup the cluster example.

> but: Is there any guarantee that before the post-deploy hook runs, all the replicas are added as endpoints to the service?

This is the right question. If it is not, then the subsequent calls getting more members is just "accidental", and not _correct by design_. If our implementation would rely on multiple DNS queries to "make sure we found all members", that's room for bugs and flakes.

I'll investigate the answer.

rhcarvalho:
https://docs.openshift.org/latest/dev_guide/deployments.html#rolling-strategy

> The Rolling strategy will:
> 1. Execute any pre lifecycle hook.
> 2. Scale up the new deployment based on the surge configuration.
> 3. Scale down the old deployment based on the max unavailable configuration.
> 4. Repeat this scaling until the new deployment has reached the desired replica count and the old deployment has been scaled to zero.
> 5. Execute any post lifecycle hook.

_hoping the docs are up-to-date_ :pray: 

Looking at the code for the Rolling deployer, which replicates the logic for Recreate as well, corroborates to my prayer:
[pkg/deploy/strategy/rolling/rolling.go#L97-L226](https://github.com/openshift/origin/blob/4eeded73703bde1de063763b2bd7a06817143061/pkg/deploy/strategy/rolling/rolling.go#L97-L226)

This will be even better if we have readiness probes -- we just need to verify https://github.com/openshift/mongodb/pull/120.

bparees:
@rhcarvalho 

```
The maxUnavailable parameter is the maximum number of pods that can be unavailable during the update. The maxSurge parameter is the maximum number of pods that can be scheduled above the original number of pods. Both parameters can be set to either a percentage (e.g., 10%) or an absolute value (e.g., 2). The default value for both is 25%.
```

https://docs.openshift.org/latest/dev_guide/deployments.html#strategies

bparees:
hm.  That may not mean what I thought, but i'm still pretty sure @smarterclayton just told me that a deployment can reach "success" even if not all replicas are available.

So the question is, what is going to happen if we run the post-deployment hook, and then a couple more replicas show up?

rhcarvalho:
> So the question is, what is going to happen if we run the post-deployment hook, and then a couple more replicas show up?

Replicas that are not in the replica set should broadcast themselves to be added to the replica set.

I think that's the same situation when we do `oc scale dc/mongo --replicas=4` -- there is no post-deployment hook after scaling, and the replica set should be able to reconfigure itself.

bparees:
> Replicas that are not in the replica set should broadcast themselves to be added to the replica set.
> I think that's the same situation when we do oc scale dc/mongo --replicas=4 -- there is no post-deployment hook after scaling, and the replica set should be able to reconfigure itself.

yeah that was my assumption (except as you noted, scaling is currently broken :) ), but i figured it was worth calling out explicitly.  Because if other replicas will broadcast to join, why don't we just add a single member to the replica set, and let the others join automatically, regardless of how many are "available" when the post-deploy hook runs?

that way we'd have a single codepath for adding members to the replicaset.

rhcarvalho:
> why don't we just add a single member to the replica set,

The problem is _which one_?

Using the post-deployment hook we know for sure which one, and we also make sure it is not part of the replica set itself (at least not as a primary).

If we would put this code in the pod template command, every replica would try to create a replica set by itself.

grdryn:
> > why don't we just add a single member to the replica set,
> 
> The problem is which one?
> 
> Using the post-deployment hook we know for sure which one, and we also make sure it is not part of the replica set itself (at least not as a primary).

If you change the template to use a DeploymentConfig per Pod, then you can have the post-hook run for just one. Having separate DeploymentConfigs for each replica would mean that scaling would be odd, but it would also give a workaround for #114 (we're going to go with separate DCs in production, and just not use the scaling mechanism, until the better solution is ready).

bparees:
> The problem is which one?

any one?  which one are we adding today?  does it matter if we just add one and ignore the rest?

> Using the post-deployment hook we know for sure which one, and we also make sure it is not part of the replica set itself (at least not as a primary).

how do we know this?

> If we would put this code in the pod template command, every replica would try to create a replica set by itself.

i'm definitely not suggesting that.

i'm not as familiar w/ the clustering flow as you are, so perhaps that's where my confusion is coming from, let's talk through the flow on scrum tomorrow.

rhcarvalho:
> i'm not as familiar w/ the clustering flow as you are, so perhaps that's where my confusion is coming from, let's talk through the flow on scrum tomorrow.

Let's do that.

rhcarvalho:
https://ci.openshift.redhat.com/jenkins/job/mongodb/212/console
Got a yum error saying "This system is not registered to Red Hat Subscription Management. You can use subscription-manager to register." -- this is weird...

[test] again

bparees:
I think that's a normal message.

Ben Parees | OpenShift

rhcarvalho:
@php-coder PTAL :eyes:

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -92,15 +92,18 @@ function endpoints() {
 }
 
 # build_mongo_config builds the MongoDB replicaSet config used for the cluster
-# initialization
+# initialization.
+# Takes a list of space-separated member IPs as the first argument.
```
php-coder, line 96: s/as the first argument/as the argument/ ?

php-coder:
LGTM

mfojtik:
LGTM

rhcarvalho:
[test] again

PR #126, opened 2016-02-15
add rsync and tar to image

bparees:


mnagy:
LGTM

omron93:
Why rsync and tar should be there?

bparees:
@omron93 it's needed to support:
https://docs.openshift.org/latest/dev_guide/copy_files_to_container.html

sorry i should have made the commit text clearer.

PR #127, opened 2016-02-16
Add MongoDB 3.2 container image.

omron93:
The latest stable MongoDB release is 3.2.

This PR adds a container image for this release. 3.2 implementation/directory is really similar to 2.6 (only necessary installation changes, db.addUser to db.createUser substitution in test/run).

**Image from this PR is not buildable!**
The rh-mongodb32 SCL is not build in CentOS yet. I am creating this PR to discuss some thing to be able to make a quicker merge after builds in cbs will be ready.

One big change in MongoDB 3.2 is a new default storage engine - WiredTiger. This storage engine bring new feature and better performance to MongoDB (https://www.mongodb.com/blog/post/whats-new-mongodb-30-part-3-performance-efficiency-gains-new-storage-architecture), but it is x86_64 only!
For WiredTiger MONGODB_NOPREALLOC and MONGODB_SMALLFILES has no effect. WT behaves similar by default.

@mfojtik Is it a problem that WT is 64bit only? Could I remove MONGODB_NOPREALLOC and MONGODB_SMALLFILES options from README, usage() and config file?

### DIFF
--- a/3.2/README.md
+++ b/3.2/README.md
@@ -1,0 +1,79 @@
+MongoDB Docker image
+====================
+
+This repository contains Dockerfiles for MongoDB images for general usage and OpenShift.
+Users can choose between RHEL and CentOS based images.
+
+Environment variables
+---------------------------------
+
+The image recognizes the following environment variables that you can set during
+initialization by passing `-e VAR=VALUE` to the Docker run command.
+
+|    Variable name          |    Description                              |
+| :------------------------ | -----------------------------------------   |
+|  `MONGODB_USER`       | User name for MONGODB account to be created |
+|  `MONGODB_PASSWORD`       | Password for the user account               |
+|  `MONGODB_DATABASE`       | Database name                               |
+|  `MONGODB_ADMIN_PASSWORD` | Password for the admin user                 |
+
+
+The following environment variables influence the MongoDB configuration file. They are all optional.
+
+|    Variable name      |    Description                                                            |    Default
+| :-------------------- | ------------------------------------------------------------------------- | ----------------
+|  `MONGODB_NOPREALLOC` | Disable data file preallocation (only for mounted data directory from older MongoDB server).  |  true
+|  `MONGODB_SMALLFILES` | Set MongoDB to use a smaller default data file size (only for mounted data directory from older MongoDB server).  |  true
+|  `MONGODB_QUIET`      | Runs MongoDB in a quiet mode that attempts to limit the amount of output. |  true
+
+
+You can also set the following mount points by passing the `-v /host:/container` flag to Docker.
+
+|  Volume mount point         | Description            |
+| :-------------------------- | ---------------------- |
+|  `/var/lib/mongodb/data`   | MongoDB data directory |
+
+**Notice: When mouting a directory from the host into the container, ensure that the mounted
+directory has the appropriate permissions and that the owner and group of the directory
+matches the user UID or name which is running inside the container.**
+
+
+Usage
+---------------------------------
+
+For this, we will assume that you are using the `centos/mongodb-32-centos7` image.
+If you want to set only the mandatory environment variables and store the database
+in the `/home/user/database` directory on the host filesystem, execute the following command:
+
+```
+$ docker run -d -e MONGODB_USER=<user> -e MONGODB_PASSWORD=<password> -e MONGODB_DATABASE=<database> -e MONGODB_ADMIN_PASSWORD=<admin_password> -v /home/user/database:/var/lib/mongodb/data centos/mongodb-32-centos7
+```
+
+If you are initializing the database and it's the first time you are using the
+specified shared volume, the database will be created with two users: `admin` and `MONGODB_USER`. After that the MongoDB daemon
+will be started. If you are re-attaching the volume to another container, the
+creation of the database user and admin user will be skipped and only the
+MongoDB daemon will be started.
+
+
+MongoDB admin user
+---------------------------------
+
+The admin user name is set to `admin` and you have to to specify the password by
+setting the `MONGODB_ADMIN_PASSWORD` environment variable. This process is done
+upon database initialization.
+
+
+Changing passwords
+------------------
+
+Since passwords are part of the image configuration, the only supported method
+to change passwords for the database user (`MONGODB_USER`) and admin user is by
+changing the environment variables `MONGODB_PASSWORD` and
+`MONGODB_ADMIN_PASSWORD`, respectively.
+
+Changing database passwords directly in MongoDB will cause a mismatch between
+the values stored in the variables and the actual passwords. Whenever a database
+container starts it will reset the passwords to the values stored in the
+environment variables.
+
--- a/3.2/Dockerfile
+++ b/3.2/Dockerfile
@@ -1,0 +1,59 @@
+FROM centos:centos7
+MAINTAINER SoftwareCollections.org <sclorg@redhat.com>
+
+# MongoDB image for OpenShift.
+#
+# Volumes:
+#  * /var/lib/mongodb/data - Datastore for MongoDB
+# Environment:
+#  * $MONGODB_USER - Database user name
+#  * $MONGODB_PASSWORD - User's password
+#  * $MONGODB_DATABASE - Name of the database to create
+#  * $MONGODB_ADMIN_PASSWORD - Password of the MongoDB Admin
+
+ENV MONGODB_VERSION=3.2 \
+    HOME=/var/lib/mongodb
+
+LABEL io.k8s.description="MongoDB is a scalable, high-performance, open source NoSQL database." \
+      io.k8s.display-name="MongoDB 3.2" \
+      io.openshift.expose-services="27017:mongodb" \
+      io.openshift.tags="database,mongodb,rh-mongodb32"
+
+EXPOSE 27017
+
+# Due to the https://bugzilla.redhat.com/show_bug.cgi?id=1206151,
+# the whole /var/lib/mongodb/ dir has to be chown-ed.
+RUN yum install -y centos-release-scl-rh && \
+    yum-config-manager --enable centos-sclo-rh-testing && \
+    INSTALL_PKGS="bind-utils gettext iproute rsync tar rh-mongodb32-mongodb rh-mongodb32" && \
+    yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
+    rpm -V $INSTALL_PKGS && \
+    yum clean all && \
+    mkdir -p /var/lib/mongodb/data && chown -R mongodb.0 /var/lib/mongodb/ /var/opt/rh/rh-mongodb32/lib/mongodb && \
+    # Loosen permission bits to avoid problems running container with arbitrary UID
+    chmod g+w -R /var/opt/rh/rh-mongodb32/lib/mongodb && \
+    chmod -R g+rwx /var/lib/mongodb
+
+# Get prefix path and path to scripts rather than hard-code them in scripts
+ENV CONTAINER_SCRIPTS_PATH=/usr/share/container-scripts/mongodb \
+    MONGODB_PREFIX=/opt/rh/rh-mongodb32/root/usr \
+    ENABLED_COLLECTIONS=rh-mongodb32
+
+# When bash is started non-interactively, to run a shell script, for example it
+# looks for this variable and source the content of this file. This will enable
+# the SCL for all scripts without need to do 'scl enable'.
+ENV BASH_ENV=${CONTAINER_SCRIPTS_PATH}/scl_enable \
+    ENV=${CONTAINER_SCRIPTS_PATH}/scl_enable \
+    PROMPT_COMMAND=". ${CONTAINER_SCRIPTS_PATH}/scl_enable"
+
+ADD root /
+
+# Container setup
+RUN touch /etc/mongod.conf && chown mongodb:0 /etc/mongod.conf && /usr/libexec/fix-permissions /etc/mongod.conf
+
+VOLUME ["/var/lib/mongodb/data"]
+
+USER 184
+
+ENTRYPOINT ["container-entrypoint"]
+CMD ["run-mongod"]
--- a/3.2/Dockerfile.rhel7
+++ b/3.2/Dockerfile.rhel7
@@ -1,0 +1,65 @@
+FROM rhel7.2
+# MongoDB image for OpenShift.
+#
+# Volumes:
+#  * /var/lib/mongodb/data - Datastore for MongoDB
+# Environment:
+#  * $MONGODB_USER - Database user name
+#  * $MONGODB_PASSWORD - User's password
+#  * $MONGODB_DATABASE - Name of the database to create
+#  * $MONGODB_ADMIN_PASSWORD - Password of the MongoDB Admin
+
+ENV MONGODB_VERSION=3.2 \
+    HOME=/var/lib/mongodb
+
+LABEL io.k8s.description="MongoDB is a scalable, high-performance, open source NoSQL database." \
+      io.k8s.display-name="MongoDB 3.2" \
+      io.openshift.expose-services="27017:mongodb" \
+      io.openshift.tags="database,mongodb,rh-mongodb32"
+
+# Labels consumed by Red Hat build service
+LABEL BZComponent="rh-mongodb32-docker" \
+      Name="rhscl/mongodb-32-rhel7" \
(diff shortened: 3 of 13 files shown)


omron93:
@mfojtik @bparees ping

mfojtik:
@omron93 I think our strategy is to always keep rhel7 + centos7 image together so I would wait merging this till the centos7 SCL is ready.

Is there any delta in scripts between this version and the existing versions? Also have you tested replication?

[test]

omron93:
> I think our strategy is to always keep rhel7 + centos7 image together so I would wait merging this till the centos7 SCL is ready.

I understand. Now I only want to know OpenShift POV on storage engine changes and if it is possible to remove config options... [see my comment](https://github.com/openshift/mongodb/pull/127#issue-134040712)

> Is there any delta in scripts between this version and the existing versions?

[rh-mongodb-32-patch.txt](https://github.com/openshift/mongodb/files/136268/rh-mongodb-32-patch.txt)

> Also have you tested replication?

No. I thought it is supported only in mongodb-24.
Replication scripts are same as for mongodb-26.

omron93:
Changed installation in CentOS Dockerfile. It is now using centos-sclo-rh-testing repository... (see https://github.com/openshift/postgresql/pull/99#issuecomment-185416629)

However it is not working now. It depends on this bug https://bugs.centos.org/view.php?id=10302

bparees:
@omron93 i think 64-bit only is ok, and you can remove the ENV variables that have will no longer have any effect.

@mfojtik let's not hold this up on replication.  Once @rhcarvalho fixes the replication for 2.4 and 2.6, and @php-coder writes replication tests, we can come back and test this with that.

bparees:
[test]

mfojtik:
LGTM (when it builds ;-)

rhcarvalho:
I feel really bad about our approach of copy-pasting files when adding support for new versions, for me it brings unnecessary maintenance burden.

@bparees if we're merging with the second commit in, we need an issue open to track moving away from a testing repository to an official release repository.

rhcarvalho:
@omron93 why don't we need the `v8314` package for 3.2, is it also because of changes in MongoDB?

rhcarvalho:
The diff between `2.6` and `3.2` looks reasonable, only three files differ:
- `test/run`
- `Dockerfile`
- `Dockerfile.rhel`

The changes in `test/run` are `removeUser -> dropUser` and `addUser -> createUser`. Perhaps those should be changed in the replication scripts as well, to save us time figuring out why the scripts don't work when we will be able to support replication for 3.2.

bparees:
> I feel really bad about our approach of copy-pasting files when adding support for new versions, for me it brings unnecessary maintenance burden.

@rhcarvalho i'm open to suggestions, but trying to make this stuff common and then finding parts of it are not common, is also a large maintenance burden.  We are also not likely to have that many versions of any given image, especially with older versions rolling EOL.

bparees:
@rhcarvalho one option i'd consider, if all the files are really identical, would be creating symlinks for now, and breaking them if we have to introduce deltas.

bparees:
> @bparees if we're merging with the second commit in, we need an issue open to track moving away from a testing repository to an official release repository.

yes.  @omron93 do you want to open that issue?

mfojtik:
@bparees I don't think symlinks with work with Docker build (just guessing)

mfojtik:
@bparees but `make build` can copy the files if needed.

omron93:
> why don't we need the v8314 package for 3.2, is it also because of changes in MongoDB

MongoDB 3.2 switched from v8 to mozjs38 JS engine

> The changes in test/run are removeUser -> dropUser and addUser -> createUser. Perhaps those should be changed in the replication scripts as well, to save us time figuring out why the scripts don't work when we will be able to support replication for 3.2.

This deprecated functions are replaced even in 2.6 scripts -> so there is no change in it.

> yes. @omron93 do you want to open that issue?

New issue - https://github.com/openshift/mongodb/issues/129

bparees:
@omron93 remove the references to the unused env variables, squash, and i'll merge (if it passes tests)

omron93:
> remove the references to the unused env variables, squash, and i'll merge (if it passes tests)

DONE

New diff from 2.6 - [rh-mongodb-32-patch.txt](https://github.com/openshift/mongodb/files/137817/rh-mongodb-32-patch.txt)

However this PR still have to wait for CentOS builds availability :-(

bparees:
@omron93 you don't want it published using the content of the test repo?

omron93:
> you don't want it published using the content of the test repo?

Yes, I want to. However it was not ready even in the testing repo.
Now hhorak is building rh-mongodb32 in cbs, so hopefully it will be available in the testing repo soon.

bparees:
@omron93 ok, re-ping me when the centos and rhel repos are both ready and kick off a test for this.

3.2/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+**Notice: When mouting a directory from the host into the container, ensure that the mounted
+directory has the appropriate permissions and that the owner and group of the directory
+matches the user UID or name which is running inside the container.**
+
+
+Usage
+---------------------------------
+
+For this, we will assume that you are using the `centos/mongodb-26-centos7` image.
```
elyscape: This should probably be changed to `centos/mongodb-32-centos7`.

rhcarvalho: good catch... we shouldn't really be copy-pasting files to add new versions :cry:

zipzit:
Hmmm.  I'm trying to link from OpenShift server --> mlab.com mongodb.  Keep getting auth errors.  [Reference. ](http://stackoverflow.com/questions/35935531/failed-database-connection-between-openshift-server-and-mlab-com)  Folks at mlabs claim:

> You'll need to upgrade your mongo shell version to 3.0+ in order to connect and authenticate to an mLab Sandbox database server. It looks like version 2.4.9 is being used. Can you please upgrade and try again?

Will fixing this issue have the effect of upgrading the mongo tools accessible on OpenShift?  Many thanks.

rhcarvalho:
> Will fixing this issue have the effect of upgrading the mongo tools accessible on OpenShift?

@zipzit this repository is for the MongoDB image for OpenShift v3. The problem you are facing is with OpenShift v2, so this PR is unrelated.

Please allow some time for somebody to answer your question on StackOverflow, or try other support channels.

rhcarvalho:
@omron93 any updates on the MongoDB 3.2 package availability?

omron93:
@mfojtik @rhcarvalho Packages seems to be available. Please test and merge.

Differences from 2.6 are only very small. See [26-to-32-diff.txt](https://github.com/openshift/mongodb/files/213368/26-to-32-diff.txt)

3.2/README.md
bparees: @omron93 yes this still needs to be updated.

bparees:
@omron93 I think you need to rebase, i'm seeing a lot more differences in the /root/usr/share/container-scripts/mongodb directory, where those scripts have been significantly improved in the 2.4 and 2.6 trees.

bparees:
also mongodb.conf.template now includes 2 new config settings (prealloc and smallfiles)

3.2/README.md
omron93: I missed this. Sry.

omron93:
@bparees Rebased.

Diff from 2.6 version to 3.2 version - [26-to-32-diff.txt](https://github.com/openshift/mongodb/files/214990/26-to-32-diff.txt)

Config file fixed.

bparees:
@omron93 do the prealloc and smallfiles config options not apply to 3.2?

bparees:
[test]

bparees:
```
The command '/bin/sh -c yum install -y yum-utils &&     yum-config-manager --enable rhel-server-rhscl-7-rpms &&     yum-config-manager --enable rhel-7-server-optional-rpms &&     INSTALL_PKGS="bind-utils gettext iproute rsync tar rh-mongodb32-mongodb rh-mongodb32" &&     yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS &&     rpm -V $INSTALL_PKGS &&     yum clean all &&     mkdir -p /var/lib/mongodb/data && chown -R mongodb.0 /var/lib/mongodb/ /var/opt/rh/rh-mongodb32/lib/mongodb &&     chmod g+w -R /var/opt/rh/rh-mongodb32/lib/mongodb &&     chmod -R g+rwx /var/lib/mongodb' returned a non-zero code: 2
```

guessing we are still missing the mongodb32 packages for rhel (maybe missing in rhui, @tdawson ? )

omron93:
> do the prealloc and smallfiles config options not apply to 3.2?

Yes.
MongoDB ships two storage engines:
- mmapv1 - used in previous versions as default. It uses prealloc and smallfiles options.
- wiredTiger - default storage engine since 3.2. It brings performance improvements. It does not use prealloc and smallfiles options.

> guessing we are still missing the mongodb32 packages for rhel

RHSCL is not released yet :-(

omron93:
> do the prealloc and smallfiles config options not apply to 3.2?

@bparees I was wrong, they apply to 3.2 with mounted volume created by 2.6 container. MongoDB will detect that the database was created by mmapv1 and use it. (To note: wiredTiger and mmapv1 are incompatible - it is necessary to run `mongodump` and `mongorestore` procedure)

Now these two options are marked only as compatible options for mounted volumes.

[26-to-32-diff.txt](https://github.com/openshift/mongodb/files/223703/26-to-32-diff.txt)

bparees:
@omron93 needs a squash, otherwise lgtm pending the availability of the 3.2 packages.

rhcarvalho:
@omron93 will there be an image / SCL release including MongoDB 3.0?

Looks like to upgrade from 2.6 to 3.2, one needs to first upgrade to 3.0:
https://docs.mongodb.org/manual/release-notes/3.2-upgrade/#upgrade-standalone-mongod-instance-to-mongodb-3-2

omron93:
>  will there be an image / SCL release including MongoDB 3.0?
> 
> Looks like to upgrade from 2.6 to 3.2, one needs to first upgrade to 3.0:
> https://docs.mongodb.org/manual/release-notes/3.2-upgrade/#upgrade-standalone-mongod-instance-to-mongodb-3-2

In RHSCL there will be 3.0 package exactly for this purpose. This collection does not include things not necessary for the upgrade and it won't be supported for normal use.

It there some support for upgrades in OpenShift images?

bparees:
@omron93 we haven't faced an upgrade scenario yet.  There are a number of ways a user might tackle it, but certainly one option would be to change the imagestreamtag that a particular deployment config was using, so the container would get re-deployed w/ the new image.  So in this case someone who had a running mongo2.6 container would change their tag to mongo:3.0, let it redeploy/migrate, then change the tag to mongo3.2.  (this assumes their data is stored on a persistent volume of course).

but to do that we'd need to provide an explicit 3.0 image as well.

omron93:
> but to do that we'd need to provide an explicit 3.0 image as well.

@bparees I understand. It shouldn't be a problem. Scripts for rh-mongodb30upg should be the same as for rh-mongodb32. So minimal changes.

How to specify that mongodb-30-centos7 should be used for upgrade only. Probably bugs which does not affect upgrade procedure to 3.2 won't be fixed. How does OpenShift handles thinks like this? One line in README is enought?

omron93:
Squashed.

bparees:
> How to specify that mongodb-30-centos7 should be used for upgrade only. Probably bugs which does not affect upgrade procedure to 3.2 won't be fixed. How does OpenShift handles thinks like this? One line in README is enought?

going to have to think about that, and maybe pull in @brenton from productization.  We haven't had this situation before.

brenton:
I agree this is a someone unique situation.  Does the upgrade from 2.6->3.0->3.2 require manual intervention or downtime?

If it was possible to ship an image that was only usable for upgrade we could release it.  When I say "only usable for upgrade" I'm meaning if you tried to use the 3.0 image it would report a warning or something of that nature saying that it's only supported as a migration path to 3.2.

omron93:
No manul interventions required. There are few compatibility changes between these versions, but user could be able to fix them (if needed) with 2.6 and then upgrade. To note in case 2.4->2.6 upgrade upgrade situation was little bit more complicated :-)

The upgrade collection is called rh-mongodb30upg. How the upgrade image should be called? Normally centos/mongodb-30-centos7 or centos/mongodb-30upg-centos7?

hhorak:
I'd actually suggest `centos/mongodb-30-upg-centos7`, which is similar to what we already use for variants in thermostat and devtoolset images. Anyway, the RHEL-based images are now available as Beta, in `rhel-server-rhscl-7-beta-rpms` channel, so using this channel this PR can be finished.

rhcarvalho:
@omron93 before we merge this, please rebase and compare to 2.6, there are probably changes since  last update here.

omron93:
@rhcarvalho @bparees @mfojtik Are you fine with `centos/mongodb-30-upg-centos7`? It looks good to me.

Merged changes from master. RHEL7 variant now use RHSCL Beta packages. Also squashed.
New diff - [26-to-32-diff.txt](https://github.com/openshift/mongodb/files/262957/26-to-32-diff.txt)

Please test.

bparees:
@omron93 yeah that name sounds fine. Unfortunately the PR tester is broken at the moment due to issues w/ the github PR tester, but as soon as we get that worked out i'll get back to this.

bparees:
PR tester is back online.
[test]

omron93:
@hhorak @bparees Do you have any idea where could be a problem with installation of rh-mongodb32?

`14: curl#77 - "Problem with the SSL CA cert (path? access rights?)"` ?

hhorak:
When I try manually the problematic URL, I see browser reporting "invalid security certificate". Probably something wrong in OpenShift repos?

bparees:
guessing the rhel base image we use just doesn't have the right beta repos enabled?

the failure i see is just a yum install failure:
The command '/bin/sh -c yum install -y yum-utils &&     yum-config-manager --enable rhel-server-rhscl-7-beta-rpms &&     yum-config-manager --enable rhel-7-server-optional-rpms &&     INSTALL_PKGS="bind-utils gettext iproute rsync tar rh-mongodb32-mongodb rh-mongodb32" &&     yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS &&     rpm -V $INSTALL_PKGS &&     yum clean all &&     mkdir -p /var/lib/mongodb/data && chown -R mongodb.0 /var/lib/mongodb/ /var/opt/rh/rh-mongodb32/lib/mongodb &&     chmod g+w -R /var/opt/rh/rh-mongodb32/lib/mongodb &&     chmod -R g+rwx /var/lib/mongodb' returned a non-zero code: 2

bparees:
and i should note that the ruby2.3 PR passed tests using the same repos, so are we sure the necessary packages are in the test repos?

bparees:
nm, not sure why the ruby2.3 PR passed at some point, but it's currently not building successfully either.

hhorak:
@bparees I still don't think we can handle this from withing images. From what I know about building on RHEL, it depends on how the host is configured and subscribed. You might want to check how the host is configured and what is inside `/usr/share/rhel/secrets/` on host, because that is used inside of images, IIUIC. The images only enable repositories, IMO.

bparees:
> @bparees I still don't think we can handle this from withing images. From what I know about building on RHEL, it depends on how the host is configured and subscribed. You might want to check how the host is configured and what is inside /usr/share/rhel/secrets/ on host, because that is used inside of images, IIUIC. The images only enable repositories, IMO.

we actually bake the right repos into our base rhel image that we build on top of and it does not currently include the scl beta repos.  i'm not totally sure i want it to, either, at risk of picking up things we don't want.  Is that an unfounded fear?

hhorak:
From my PoV, there is nothing to be afraid of. Any Dockerfile is responsible for enabling only repositories it needs -- so, if the base image included _disabled_ RHSCL Beta repos, upper layer can't use them unless explicitly enabling them. So, it seems safe to me to have such disabled RHSCL Beta repos baked even in the base image.

hhorak:
@omron93 we can use proper RHSCL repos now.

omron93:
It is using RHCSL 2.2 packages now.
@bparees @mfojtik Please test?

Diff is same (only RHEL repo differ) 
[26-to-32-diff.txt](https://github.com/openshift/mongodb/files/302981/26-to-32-diff.txt)

bparees:
[test]

omron93:
Sorry, I've forgotten to rebase to master.

See, #164

PR #128, opened 2016-02-18
Fix version typo mongo26 to mongo24

nekop:
Fix `mongo26` in Mongo 2.4 script


mfojtik:
LGTM, good catch! I wonder how could this work before...

@bparees FYI

rhcarvalho:
Looks like MONGODB_PREFIX is not referenced in the scripts :)

PR #133, opened 2016-03-01
Fix getting of primary member

### LINKED ISSUE #125: Replication: after scale up, new pod cannot connect to replica set
## Steps to reproduce
1. Create new app:
   
   ``` console
   $ oc new-app https://raw.githubusercontent.com/openshift/mongodb/master/2.4/examples/replica/mongodb-clustered.json
   ```
2. Stream logs until the deployment is finished. Replica set initialization logs are streamed as well:
   
   ``` console
   $ oc logs -f mongodb-1-deploy
   ```
3. Connect to the primary member via any pod, check that replica set configuration is ok:
   
   ``` console
   $ oc get pods
   NAME              READY     STATUS    RESTARTS   AGE
   mongodb-1-4thdo   1/1       Running   0          11m
   mongodb-1-7l8j1   1/1       Running   0          11m
   mongodb-1-qcoob   1/1       Running   1          12m
   $ oc exec mongodb-1-4thdo -it -- bash -c 'mongo admin -u admin -p $MONGODB_ADMIN_PASSWORD --host $MONGODB_REPLICA_NAME/localhost'
   rs0:PRIMARY> rs.isMaster()
   {
       "setName" : "rs0",
       "ismaster" : true,
       "secondary" : false,
       "hosts" : [
               "172.17.0.4:27017",
               "172.17.0.5:27017",
               "172.17.0.3:27017"
       ],
       "primary" : "172.17.0.4:27017",
       "me" : "172.17.0.4:27017",
       "maxBsonObjectSize" : 16777216,
       "maxMessageSizeBytes" : 48000000,
       "localTime" : ISODate("2016-02-10T21:16:02.683Z"),
       "ok" : 1
   }
   ```
4. Scale up, look at the logs of the new pod:
   
   ``` console
   $ oc scale dc/mongodb --replicas=4
   deploymentconfig "mongodb" scaled
   $ oc get pods
   NAME              READY     STATUS    RESTARTS   AGE
   mongodb-1-4thdo   1/1       Running   0          2m
   mongodb-1-7l8j1   1/1       Running   0          2m
   mongodb-1-qcoob   1/1       Running   1          2m
   mongodb-1-sj5hq   1/1       Running   0          18s
   $ oc logs -f mongodb-1-sj5hq
   ...
   Wed Feb 10 21:15:16.566 [rsStart] replSet can't get local.system.replset config from self or any seed (EMPTYCONFIG)
   ```
   
   The last message repeats indefinitely.
## Expected result

The new pod should be added to the replica set, and querying the primary should show the 4 members.

php-coder:
Modified logic for getting address of primary member: try to ask all neighbour nodes. Also mongo may print error message to stdout and I've added fix for that too. Plus minor improvements (like adding quotes).

PTAL @bparees @rhcarvalho

Fixes #125

### DIFF
--- a/2.4/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.4/root/usr/share/container-scripts/mongodb/common.sh
@@ -119,25 +119,56 @@ function mongo_initiate() {
 
 # get the address of the current primary member
 function mongo_primary_member_addr() {
-  local current_endpoints
-  current_endpoints=$(endpoints)
-  local mongo_node
-  mongo_node="$(echo "${current_endpoints}" | grep -v "$(container_addr)" | head -1):${CONTAINER_PORT}"
-  echo -n $(mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --host ${mongo_node} --quiet --eval "print(rs.isMaster().primary);")
+  local rc=0
+
+  endpoints | grep -v "$(container_addr)" |
+  (
+    while read mongo_node; do
+      cmd_output="$(mongo admin -u admin -p "$MONGODB_ADMIN_PASSWORD" --host "$mongo_node:$CONTAINER_PORT" --eval 'print(rs.isMaster().primary)' --quiet || true)"
+
+      # Trying to find IP:PORT in output and filter out error message because mongo prints it to stdout
+      ip_and_port_regexp='[0-9]\+\.[0-9]\+\.[0-9]\+\.[0-9]\+:[0-9]\+'
+      if addr="$(echo "$cmd_output" | grep -x "$ip_and_port_regexp")"; then
+        echo -n "$addr"
+        exit 0
+      fi
+
+      echo >&2 "Cannot get address of primary from $mongo_node node: $cmd_output"
+    done
+
+    exit 1
+  ) || rc=$?
+
+  if [ $rc -ne 0 ]; then
+    echo >&2 "Cannot get address of primary node: after checking all nodes we don't have the address"
+    return 1
+  fi
 }
 
 # mongo_remove removes the current MongoDB from the cluster
 function mongo_remove() {
-  echo "=> Removing $(mongo_addr) on $(mongo_primary_member_addr) ..."
+  local primary_addr
+  primary_addr="$(mongo_primary_member_addr)"
+
+  local mongo_addr
+  mongo_addr="$(mongo_addr)"
+
+  echo "=> Removing $mongo_addr on $primary_addr ..."
   mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" \
-    --host $(mongo_primary_member_addr) --eval "rs.remove('$(mongo_addr)');" &>/dev/null || true
+    --host "$primary_addr" --eval "rs.remove('$mongo_addr');" || true
 }
 
 # mongo_add advertise the current container to other mongo replicas
 function mongo_add() {
-  echo "=> Adding $(mongo_addr) to $(mongo_primary_member_addr) ..."
+  local primary_addr
+  primary_addr="$(mongo_primary_member_addr)"
+
+  local mongo_addr
+  mongo_addr="$(mongo_addr)"
+
+  echo "=> Adding $mongo_addr to $primary_addr ..."
   mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" \
-    --host $(mongo_primary_member_addr) --eval "rs.add('$(mongo_addr)');"
+    --host "$primary_addr" --eval "rs.add('$mongo_addr');"
 }
 
 # run_mongod_supervisor runs the MongoDB replica supervisor that manages
--- a/2.6/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/common.sh
@@ -119,25 +119,56 @@ function mongo_initiate() {
 
 # get the address of the current primary member
 function mongo_primary_member_addr() {
-  local current_endpoints
-  current_endpoints=$(endpoints)
-  local mongo_node
-  mongo_node="$(echo "${current_endpoints}" | grep -v "$(container_addr)" | head -1):${CONTAINER_PORT}"
-  echo -n $(mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --host ${mongo_node} --quiet --eval "print(rs.isMaster().primary);")
+  local rc=0
+
+  endpoints | grep -v "$(container_addr)" |
+  (
+    while read mongo_node; do
+      cmd_output="$(mongo admin -u admin -p "$MONGODB_ADMIN_PASSWORD" --host "$mongo_node:$CONTAINER_PORT" --eval 'print(rs.isMaster().primary)' --quiet || true)"
+
+      # Trying to find IP:PORT in output and filter out error message because mongo prints it to stdout
+      ip_and_port_regexp='[0-9]\+\.[0-9]\+\.[0-9]\+\.[0-9]\+:[0-9]\+'
+      if addr="$(echo "$cmd_output" | grep -x "$ip_and_port_regexp")"; then
+        echo -n "$addr"
+        exit 0
+      fi
+
+      echo >&2 "Cannot get address of primary from $mongo_node node: $cmd_output"
+    done
+
+    exit 1
+  ) || rc=$?
+
+  if [ $rc -ne 0 ]; then
+    echo >&2 "Cannot get address of primary node: after checking all nodes we don't have the address"
+    return 1
+  fi
 }
 
 # mongo_remove removes the current MongoDB from the cluster
 function mongo_remove() {
-  echo "=> Removing $(mongo_addr) on $(mongo_primary_member_addr) ..."
+  local primary_addr
+  primary_addr="$(mongo_primary_member_addr)"
+
+  local mongo_addr
+  mongo_addr="$(mongo_addr)"
+
+  echo "=> Removing $mongo_addr on $primary_addr ..."
   mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" \
-    --host $(mongo_primary_member_addr) --eval "rs.remove('$(mongo_addr)');" &>/dev/null || true
+    --host "$primary_addr" --eval "rs.remove('$mongo_addr');" || true
 }
 
 # mongo_add advertise the current container to other mongo replicas
 function mongo_add() {
-  echo "=> Adding $(mongo_addr) to $(mongo_primary_member_addr) ..."
+  local primary_addr
+  primary_addr="$(mongo_primary_member_addr)"
+
+  local mongo_addr
+  mongo_addr="$(mongo_addr)"
+
+  echo "=> Adding $mongo_addr to $primary_addr ..."
   mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" \
-    --host $(mongo_primary_member_addr) --eval "rs.add('$(mongo_addr)');"
+    --host "$primary_addr" --eval "rs.add('$mongo_addr');"
 }
 
 # run_mongod_supervisor runs the MongoDB replica supervisor that manages


bparees:
lgtm, will wait for @rhcarvalho though. he's a better bash guru than me.

bparees:
also needs squashing

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -117,23 +117,49 @@ function mongo_initiate() {
...
+    fi
+
+    echo >&2 "Cannot get address of primary from $mongo_node node: $cmdOutput"
+    exit 1
+  done && return
+
+  echo >&2 "Cannot get address of primary node: after checking all nodes we don't have the address"
+
+  # return empty string
+  echo
```
rhcarvalho: Notes after conversation:
- need to be "pipefail-compatible"
- exit code of `mongo_primary_member_addr` should be according to success or not
- cmdOutput -> cmd_output
- printf -> echo
- remove `local cmdOutput`
- move IP pattern to a variable

php-coder: > exit code of mongo_primary_member_addr should be according to success or not

@rhcarvalho @legionus I not agree with that I don't see benefit of this. In caller's method we always analyze the output of the method and adding exit code will means that I also add code for ignoring exit code.

php-coder: @rhcarvalho thank you for the feedback! I've fixed most of the notes. PTAL again.

> need to be "pipefail-compatible"

After testing I've found that it was already "pipefail-compatible".

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -117,23 +117,49 @@ function mongo_initiate() {
...
 # get the address of the current primary member
 function mongo_primary_member_addr() {
-  local current_endpoints=$(endpoints)
-  local mongo_node="$(echo "${current_endpoints}" | grep -v "$(container_addr)" | head -1):${CONTAINER_PORT}"
-  echo -n $(mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --host ${mongo_node} --quiet --eval "print(rs.isMaster().primary);")
+  endpoints | grep -v "$(container_addr)" | while read mongo_node; do
+    cmd_output="$(mongo admin -u admin -p "$MONGODB_ADMIN_PASSWORD" --host "$mongo_node:$CONTAINER_PORT" --eval 'print(rs.isMaster().primary)' --quiet || true)"
+
+    # Trying to find IP:PORT in output and filter out error message because mongo prints it to stdout
+    ipAndPortRegexp='[0-9]\+\.[0-9]\+\.[0-9]\+\.[0-9]\+:[0-9]\+'
```
rhcarvalho: Please use the underscore naming convention :)

php-coder: Fixed. Thanks!

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -117,23 +117,60 @@ function mongo_initiate() {
...
+    cmd_output="$(mongo admin -u admin -p "$MONGODB_ADMIN_PASSWORD" --host "$mongo_node:$CONTAINER_PORT" --eval 'print(rs.isMaster().primary)' --quiet || true)"
+
+    # Trying to find IP:PORT in output and filter out error message because mongo prints it to stdout
+    ip_and_port_regexp='[0-9]\+\.[0-9]\+\.[0-9]\+\.[0-9]\+:[0-9]\+'
+    if addr="$(echo "$cmd_output" | grep -x "$ip_and_port_regexp")"; then
+      echo "$addr"
+      exit 0
+    fi
+
+    echo >&2 "Cannot get address of primary from $mongo_node node: $cmd_output"
```
rhcarvalho: `cmd_output` will only contain the stdout of the mongo shell execution, or be empty if `true` was evaluated, right?

So I'm imagining in most cases it would be empty when we print this message, which is not very helpful, what do you think?

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -117,23 +117,60 @@ function mongo_initiate() {
 
 # get the address of the current primary member
 function mongo_primary_member_addr() {
-  local current_endpoints=$(endpoints)
-  local mongo_node="$(echo "${current_endpoints}" | grep -v "$(container_addr)" | head -1):${CONTAINER_PORT}"
-  echo -n $(mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --host ${mongo_node} --quiet --eval "print(rs.isMaster().primary);")
+  endpoints | grep -v "$(container_addr)" | while read mongo_node; do
+    cmd_output="$(mongo admin -u admin -p "$MONGODB_ADMIN_PASSWORD" --host "$mongo_node:$CONTAINER_PORT" --eval 'print(rs.isMaster().primary)' --quiet || true)"
```
rhcarvalho: I would prefer this to be part of the pipeline in the `if` instead of having `|| true` to ignore the exit code.

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -117,23 +117,60 @@ function mongo_initiate() {
...
 }
 
 # mongo_remove removes the current MongoDB from the cluster
 function mongo_remove() {
-  echo "=> Removing $(mongo_addr) on $(mongo_primary_member_addr) ..."
+  local primary_addr
+  primary_addr="$(mongo_primary_member_addr)"
+
+  if [ -z "$primary_addr" ]; then
+    echo >&2 "Cannot remove node from the replica set: primary node was not found"
```
rhcarvalho: Instead of checking for empty return value from `mongo_primary_member_addr` in every caller, the script should simply fail (when we'll have `set -o errexit`). I think that would make us a favor of adding less code :)

2.4/root/usr/share/container-scripts/mongodb/common.sh
php-coder: > cmd_output will only contain the stdout of the mongo shell execution, or be empty if true was evaluated, right?

None, the command output and exit code is different. `true` evaluates only when exit code != 0 and in this case we have an error message in stdout. Otherwise we have IP-address in stdout.

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -117,23 +117,60 @@ function mongo_initiate() {
...
+    return
+  fi
+
+  local mongo_addr
+  mongo_addr="$(mongo_addr)"
+
+  echo "=> Removing $mongo_addr on $primary_addr ..."
   mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" \
-    --host $(mongo_primary_member_addr) --eval "rs.remove('$(mongo_addr)');" &>/dev/null || true
+    --host "$primary_addr" --eval "rs.remove('$mongo_addr');" || true
```
rhcarvalho, line 158: Why are we removing the output redirection?

2.4/root/usr/share/container-scripts/mongodb/common.sh
rhcarvalho: If `cmd_output` is only the IP, why are we grepping in line 125?

php-coder: > If cmd_output is only the IP...

It's IP or an error message. That's why we are `grep`ing.

2.4/root/usr/share/container-scripts/mongodb/common.sh
php-coder, line 158: I think that if we fail to remove node from the cluster the error message should be present in logs. Probably, it could be useful in debugging.

2.4/root/usr/share/container-scripts/mongodb/common.sh
php-coder: I like to write less code but debugging in this case becomes a nightmare :-( Especially because stdout and stderr mixed together and not always reflects real execution flow. So, if it just fails you will see output from other commands and it will be hard to understand what was going on.

rhcarvalho: I agree with you... maybe there's a way we can get best of both worlds?

Wrapping every error with a message is probably over kill, and not having `set -e` and ignoring errors silently is not a great idea.

In this particular case here, can't we simply make `mongo_primary_member_addr` print an error message (it already does)? And then it would print an error message and exit in case of errors, not requiring callers to handle empty strings.

2.4/root/usr/share/container-scripts/mongodb/common.sh
php-coder: > I would prefer this to be part of the pipeline in the if instead of having || true to ignore the exit code

I don't see a way of doing this. If I put it in the pipeline then it will fail with `failpipeline` option.

2.4/root/usr/share/container-scripts/mongodb/common.sh
php-coder: > And then it would print an error message and exit in case of errors, not requiring callers to handle empty strings.

How would you port this code?

``` bash
echo "=> The new PRIMARY member is $(mongo_primary_member_addr), shutting down current member ..."
```

?

2.4/root/usr/share/container-scripts/mongodb/common.sh
rhcarvalho: Not if it is part of the `if`...

(the cost of taming Bash...)

``` bash
set -o pipefail
set -o errexit

if cmd1 | cmd2 | cm3; then
   echo "no error in the pipe line"
else
   echo "ops, something went bad"
fi
```

Any error in the pipeline will just cause the `if` to branch differently, but it won't exit the current program.

2.4/root/usr/share/container-scripts/mongodb/common.sh
rhcarvalho: The error message goes to stdout or stderr?

php-coder: stderr

2.4/root/usr/share/container-scripts/mongodb/common.sh
rhcarvalho: it shouldn't run if `mongo_primary_member_addr` will have any error (the program should exit)

2.4/root/usr/share/container-scripts/mongodb/common.sh
php-coder: Your example is not the same as our code. Compare the following:

``` bash
if cmd1 | cmd2 | cm3; then
   echo "no error in the pipe line"
else
   echo "ops, something went bad"
fi
```

and

``` bash
out=$(cmd1 | cmd2)
if echo $out | cm3; then
   echo "no error in the pipe line"
else
   echo "ops, something went bad: $out"
fi
```

I believe that we can't inline the `cmd1 | cmd2` here.

2.4/root/usr/share/container-scripts/mongodb/common.sh
rhcarvalho: "Bash, every day a new lesson -- guaranteed" :sunrise: 

``` console
$ cat assignment.sh 
#!/bin/bash

foo="$(echo "stdout" && echo >&2 "stderr")"

echo "The value of foo is '${foo}'"
$ ./assignment.sh 
stderr
The value of foo is 'stdout'
```

@php-coder when you make an assignment with a subshell, only the `stdout` becomes the value of the variable.
So if the error from `mongo` goes to `stderr`, there is no point in grepping the value stored in `cmd_output`, just like in this particular line it seems that `cmd_output` will always be empty when you try to print it.

Or maybe I'm wrong?

I really miss being able to write unit tests for these functions.

2.4/root/usr/share/container-scripts/mongodb/common.sh
rhcarvalho: > I believe that we can't inline the cmd1 | cmd2 here.

Let's try:

``` bash

local addr
if addr="$(mongo admin -u admin -p "$MONGODB_ADMIN_PASSWORD" --host "$mongo_node:$CONTAINER_PORT" --eval 'print(rs.isMaster().primary)' --quiet | grep -x "$ip_and_port_regexp")"; then
  echo "$addr"
  exit 0
fi

# ...
```

And I really think we don't need the `grep` part.

WDYT? :)

2.4/root/usr/share/container-scripts/mongodb/common.sh
php-coder: I could show you how it looks like:

``` console
$ mongo admin -u admin -p $MONGODB_ADMIN_PASSWORD --host localhost:21071 --eval 'print(rs.isMaster().primary)' --quiet
Thu Mar  3 13:29:09.090 Error: couldn't connect to server localhost:21071 at src/mongo/shell/mongo.js:147
exception: connect failed
$ mongo admin -u admin -p $MONGODB_ADMIN_PASSWORD --host localhost:21071 --eval 'print(rs.isMaster().primary)' --quiet >/dev/null
exception: connect failed
$ mongo admin -u admin -p $MONGODB_ADMIN_PASSWORD --host localhost:21071 --eval 'print(rs.isMaster().primary)' --quiet 2>/dev/null
Thu Mar  3 13:29:15.890 Error: couldn't connect to server localhost:21071 at src/mongo/shell/mongo.js:147
$ mongo admin -u admin -p $MONGODB_ADMIN_PASSWORD --host rs0/localhost --eval 'print(rs.isMaster().primary)' --quiet
Thu Mar  3 13:34:46.884 starting new replica set monitor for replica set rs0 with seed of localhost:27017
Thu Mar  3 13:34:46.884 successfully connected to seed localhost:27017 for replica set rs0
Thu Mar  3 13:34:46.885 changing hosts to { 0: "172.17.0.6:27017", 1: "172.17.0.5:27017", 2: "172.17.0.4:27017" } from rs0/
Thu Mar  3 13:34:46.885 trying to add new host 172.17.0.4:27017 to replica set rs0
Thu Mar  3 13:34:46.886 successfully connected to new host 172.17.0.4:27017 in replica set rs0
Thu Mar  3 13:34:46.886 trying to add new host 172.17.0.5:27017 to replica set rs0
Thu Mar  3 13:34:46.886 successfully connected to new host 172.17.0.5:27017 in replica set rs0
Thu Mar  3 13:34:46.886 trying to add new host 172.17.0.6:27017 to replica set rs0
Thu Mar  3 13:34:46.887 successfully connected to new host 172.17.0.6:27017 in replica set rs0
Thu Mar  3 13:34:46.890 Primary for replica set rs0 changed to 172.17.0.4:27017
Thu Mar  3 13:34:46.899 replica set monitor for replica set rs0 started, address is rs0/172.17.0.4:27017,172.17.0.5:27017,172.17.0.6:27017
Thu Mar  3 13:34:46.899 [ReplicaSetMonitorWatcher] starting
172.17.0.4:27017
```

As you can see Mongo always prints logs to stdout as output :-( And `--quiet` doesn't hide it.

2.4/root/usr/share/container-scripts/mongodb/common.sh
php-coder: > Let's try:

If you're OK with swallowing the error message (`Thu Mar  3 13:29:09.090 Error: couldn't connect to server localhost:21071 at src/mongo/shell/mongo.js:147`) then we can use your approach.

> And I really think we don't need the grep part.

In this case you will connect to `Thu` host because today is Thursday. I was trying to fix this issue actually :-)

rhcarvalho: Heheh the shell wins :facepunch: 

I understood from https://github.com/openshift/mongodb/pull/133/files#r54871106 that the errors and other messages go to `stderr`.

Today is Thursday again! Let's not connect to it ;-)

2.4/root/usr/share/container-scripts/mongodb/common.sh
rhcarvalho: :-(

2.4/root/usr/share/container-scripts/mongodb/common.sh
php-coder: > I understood from https://github.com/openshift/mongodb/pull/133/files#r54871106 that the errors and other messages go to stderr.

In case of error error message printed to stderr _and_ additional log message printed to _stdout_:

``` console
$ mongo admin -u admin -p $MONGODB_ADMIN_PASSWORD --host localhost:21071 --eval 'print(rs.isMaster().primary)' --quiet 2>/dev/null
Thu Mar  3 13:29:15.890 Error: couldn't connect to server localhost:21071 at src/mongo/shell/mongo.js:147
```

rhcarvalho: Yeah I get it... it's a mess :-(
Your code is in the right direction :+1:

php-coder: @rhcarvalho Thank you for your effort anyway!

To fix it properly we should configure MongoDB to not print logs to stdout. And there is even option for that: [`logpath`](https://docs.mongodb.org/v2.4/reference/configuration-options/#logpath). I've tried to use `logpath = /dev/stderr` but it didn't help and logs still was printed to stdout.

Also there is a bug that sounds very similar to our problem: [--eval should be affected by the --quiet switch](https://jira.mongodb.org/browse/SERVER-4391) and it's not fixed.

php-coder:
Rebased and squashed.

@rhcarvalho do you have more comments on this? Or you don't mind of merging it?

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -119,25 +119,60 @@ function mongo_initiate() {
...
 
 # mongo_add advertise the current container to other mongo replicas
 function mongo_add() {
-  echo "=> Adding $(mongo_addr) to $(mongo_primary_member_addr) ..."
+  local primary_addr
+  primary_addr="$(mongo_primary_member_addr)"
+
+  if [ -z "$primary_addr" ]; then
+    echo >&2 "Cannot add node to the replica set: primary node was not found"
+    return
```
rhcarvalho: Shouldn't we `exit 1` here instead of returning?

When `mongo_primary_member_addr` returns nothing, that means `mongo_add` should fail and we should stop immediately.

Otherwise we'd had printed an error to stderr, but still proceed to execute whatever came after `mongo_add`.

The same for `mongo_remove`.

rhcarvalho: Hmm, actually, is it possible to get an empty string from `mongo_primary_member_addr`?
I thought that if the grep for the IP will fail, we will print and error message and `exit 1`:

https://github.com/openshift/mongodb/pull/133/files#diff-1763ec43870f7abf682b1057cd2daa84R133

php-coder: > Shouldn't we exit 1 here instead of returning?

For me it's no, it shouldn't. On this level we don't know should we exit or not, it's a responsibility of the caller code. I'm trying to make this functions universal and I treat them as small building blocks. What if after failing `mongo_add` if want to wait and try again N times?

php-coder: Yes, it's possible. If we have a 3 nodes, asking all of them and still don't have an IP-addr of the primary node, then we will get an empty string.

rhcarvalho: > On this level we don't know should we exit or not, it's a responsibility of the caller code.

How can the caller distinguish if `mongo_add` (and `mongo_remove`) succeeded or not?

This line will `return [0]`, meaning success; while the last line of the function will return the exit code of `mongo admin ...`, correct?

If not `exit 1`, then it should be `return 1`?

php-coder: > How can the caller distinguish if mongo_add (and mongo_remove) succeeded or not?

Currently only by checking is the result empty.

> If not exit 1, then it should be return 1?

Maybe, but this change involves changing code of all the callers.

php-coder:
@bparees I believe that I've fixed all the @rhcarvalho's comments and now it could be merged.

rhcarvalho:
lgtm

php-coder:
@bparees Could you merge it? :)

rhcarvalho:
take back my LGTM, there's a bug in the while loop @php-coder found

php-coder:
I've fixed this issue and I it could be merged.

bparees:
lgtm, will merge on test pass.

PR #112, opened 2015-11-19
Redesign of the scripts

omron93:
For 2.6 only. I will do the same for 2.4 after 2.6 will be ready for merge!

Additional refactoring of the scripts.
- run-mongod divided into several parts - pre-init (checking variables,...), running mongod on localhost, init (creating users,...), post-init (before execing mongod), exec mongod

Reworked replication too.
- no need to run one pod only for initiation and stopping it
- one pod run with command "run-mongod-replication initiation", other pods with command "run-mongod-replication" will connect it...

I've tested it only on localhost, so it probably won't work in this state - I am really not sure how the config .json should look like. @mfojtik @bparees Could someone please help me with this?

PR #131, opened 2016-02-25
Bash sanitization

rhcarvalho:
Tested deploy and scale up scenario described in #125.

This should give us clearer errors in the logs instead of failure to connect to "Thu" (because today is Thursday, right?) like @php-coder was seeing :scream_cat:

PR #135, opened 2016-03-07
Join wait_for_mongo_up and wait_for_mongo_down into one function.

omron93:
Join `wait_for_mongo_up` and `wait_for_mongo_down` into one function. This two functions should change only in one line (IF statement which checks return status of mongo shell). So this PR joins this two function into one `wait_mongo` which takes first argument - "UP" or "DOWN" to specify how to test return code of mongo shell.

This PR improves maintainability of this scripts - two original function should be same in master, but `wait_for_mongo_down` does not received updates from latest commits.

This PR is a part of #112

### DIFF
--- a/2.4/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.4/root/usr/share/container-scripts/mongodb/common.sh
@@ -40,41 +40,37 @@ function cache_container_addr() {
 
 # wait_for_mongo_up waits until the mongo server accepts incomming connections
 function wait_for_mongo_up() {
-  local mongo_host
-  mongo_host="${1-}"
-  local mongo_cmd
-  mongo_cmd="mongo admin "
-
-  if [ -n "${mongo_host}" ]; then
-    mongo_cmd+="--host ${mongo_host}:${CONTAINER_PORT} "
-  fi
-
-  local i
-  for i in $(seq $MAX_ATTEMPTS); do
-    echo "=> Waiting for MongoDB service startup ${mongo_host} ..."
-    if $mongo_cmd --eval 'help' &>/dev/null; then
-      echo "=> MongoDB service has started"
-      return 0
-    fi
-    sleep $SLEEP_TIME
-  done
-  echo "=> Giving up: Failed to start MongoDB service!"
-  exit 1
+  _wait_for_mongo 1
 }
 
 # wait_for_mongo_down waits until the mongo server is down
 function wait_for_mongo_down() {
+  _wait_for_mongo 0
+}
+
+# wait_for_mongo waits until the mongo server is up/down
+# $1 - 0 or 1 - to specify for what to wait (0 - down, 1 - up)
+# $2 - host where to connect (localhost by default)
+function _wait_for_mongo() {
+  local operation=${1:-1}
+  local message="up"
+  if [[ ${operation} -eq 0 ]]; then
+    message="down"
+  fi
+
+  local mongo_cmd="mongo admin --host ${2:-localhost} --port ${CONTAINER_PORT} "
+
   local i
   for i in $(seq $MAX_ATTEMPTS); do
-    echo "=> Waiting for MongoDB service shutdown ..."
-    if ! mongo admin --eval 'help' &>/dev/null; then
-      echo "=> MongoDB service has stopped"
+    echo "=> ${2:-} Waiting for MongoDB daemon ${message}"
+    if ([[ ${operation} -eq 1 ]] && ${mongo_cmd} --eval "quit()" &>/dev/null) || ([[ ${operation} -eq 0 ]] && ! ${mongo_cmd} --eval "quit()" &>/dev/null); then
+      echo "=> MongoDB daemon is ${message}"
       return 0
     fi
-    sleep $SLEEP_TIME
+    sleep ${SLEEP_TIME}
   done
-  echo "=> Giving up: Failed to stop MongoDB service"
-  exit 1
+  echo "=> Giving up: MongoDB daemon is not ${message}!"
+  return 1
 }
 
 # endpoints returns list of IP addresses with other instances of MongoDB
--- a/2.6/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/common.sh
@@ -40,41 +40,37 @@ function cache_container_addr() {
 
 # wait_for_mongo_up waits until the mongo server accepts incomming connections
 function wait_for_mongo_up() {
-  local mongo_host
-  mongo_host="${1-}"
-  local mongo_cmd
-  mongo_cmd="mongo admin "
-
-  if [ -n "${mongo_host}" ]; then
-    mongo_cmd+="--host ${mongo_host}:${CONTAINER_PORT} "
-  fi
-
-  local i
-  for i in $(seq $MAX_ATTEMPTS); do
-    echo "=> Waiting for MongoDB service startup ${mongo_host} ..."
-    if $mongo_cmd --eval 'help' &>/dev/null; then
-      echo "=> MongoDB service has started"
-      return 0
-    fi
-    sleep $SLEEP_TIME
-  done
-  echo "=> Giving up: Failed to start MongoDB service!"
-  exit 1
+  _wait_for_mongo 1
 }
 
 # wait_for_mongo_down waits until the mongo server is down
 function wait_for_mongo_down() {
+  _wait_for_mongo 0
+}
+
+# wait_for_mongo waits until the mongo server is up/down
+# $1 - 0 or 1 - to specify for what to wait (0 - down, 1 - up)
+# $2 - host where to connect (localhost by default)
+function _wait_for_mongo() {
+  local operation=${1:-1}
+  local message="up"
+  if [[ ${operation} -eq 0 ]]; then
+    message="down"
+  fi
+
+  local mongo_cmd="mongo admin --host ${2:-localhost} --port ${CONTAINER_PORT} "
+
   local i
   for i in $(seq $MAX_ATTEMPTS); do
-    echo "=> Waiting for MongoDB service shutdown ..."
-    if ! mongo admin --eval 'help' &>/dev/null; then
-      echo "=> MongoDB service has stopped"
+    echo "=> ${2:-} Waiting for MongoDB daemon ${message}"
+    if ([[ ${operation} -eq 1 ]] && ${mongo_cmd} --eval "quit()" &>/dev/null) || ([[ ${operation} -eq 0 ]] && ! ${mongo_cmd} --eval "quit()" &>/dev/null); then
+      echo "=> MongoDB daemon is ${message}"
       return 0
     fi
-    sleep $SLEEP_TIME
+    sleep ${SLEEP_TIME}
   done
-  echo "=> Giving up: Failed to stop MongoDB service"
-  exit 1
+  echo "=> Giving up: MongoDB daemon is not ${message}!"
+  return 1
 }
 
 # endpoints returns list of IP addresses with other instances of MongoDB


omron93:
Also this function uses `return` command instead of `exit` command.

In current MongoDB scripts `set -e` is used so after fail of `wait_mongo` shell exits. And this change allows to use this function without needs to exit after failure - return code can be checked.

omron93:
[test]

@mfojtik What to do to be able to trigger tests?

rhcarvalho:
[test] should work...

rhcarvalho:
@php-coder could you please try the extended tests here?

mnagy:
I don't like the "UP" or "DOWN" argument. Can't we keep the two functions with their name unchanged and just rewrite them to use one common function?

omron93:
So you suggest add two function "wrappers" `wait_for_mongo_up` and `wait_for_mongo_down` which will call function `wait_mongo` with arguments "UP" or "DOWN"? Or what arguments use in the common function? 0/1?

mnagy:
@omron93 exactly. And yes, I would use different arguments. 0 and 1 are fine.

omron93:
@mnagy You are right, it looks better. Changed.

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -40,47 +40,45 @@ function cache_container_addr() {
...
+}
+
+# wait_mongo waits until the mongo server is up/down
+# $1 - 0 or 1 - to specify for what to wait (0 - down, 1 - up)
+# $2 - host where to connect (localhost by default)
+function wait_mongo() {
+  local operation
+  operation=-eq
+  local message
+  message="up"
```
mnagy: initialize these right on the local line (`local message="up"`, etc.)

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -40,47 +40,45 @@ function cache_container_addr() {
...
+  if [[ ${1:-} -eq 0 ]]; then
+    operation=-ne
+    message="down"
+  fi
+
+  local mongo_cmd
+  mongo_cmd="mongo admin --host ${2:-localhost} --port ${CONTAINER_PORT} "
+
+  local i
```
mnagy, line 63: no need for this line

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -40,47 +40,45 @@ function cache_container_addr() {
...
+  operation=-eq
+  local message
+  message="up"
+  if [[ ${1:-} -eq 0 ]]; then
+    operation=-ne
+    message="down"
+  fi
+
+  local mongo_cmd
+  mongo_cmd="mongo admin --host ${2:-localhost} --port ${CONTAINER_PORT} "
```
mnagy: same as above

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: I am quite confused. So what is a bash code style used in OpenShift?

See - https://github.com/openshift/mongodb/pull/131#discussion_r54128139

@rhcarvalho ?

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: same as above :- )

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93, line 63: same as above

2.4/root/usr/share/container-scripts/mongodb/common.sh
mnagy: According to https://google.github.io/styleguide/shell.xml?showone=Use_Local_Variables#Use_Local_Variables the reason is that local discards the return value of a command executed during the expansion. This does not happen here, so it's OK to use just one line and in my opinion it will look a much better.

2.4/root/usr/share/container-scripts/mongodb/common.sh
mnagy: same as above :) I've verified that `local` won't hide unbound variable references when `-o nounset` is set.

2.4/root/usr/share/container-scripts/mongodb/common.sh
mnagy, line 63: ok, this line can stay :-)

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: I agree that is looks much better :-)

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -40,47 +40,42 @@ function cache_container_addr() {
...
+  wait_mongo 0
+}
+
+# wait_mongo waits until the mongo server is up/down
+# $1 - 0 or 1 - to specify for what to wait (0 - down, 1 - up)
+# $2 - host where to connect (localhost by default)
+function wait_mongo() {
+  local operation=-eq
+  local message="up"
+  if [[ ${1:-} -eq 0 ]]; then
```
mnagy: You either need to add double quotes around `${1:-}` or use 0 as a default

rhcarvalho: quotes please

omron93: `"${1:-}" -eq 0` does not work with `[[`. So added 1 (UP) as default.

2.4/root/usr/share/container-scripts/mongodb/common.sh
rhcarvalho: Hmm... my reasoning is that that fact about `local` is obscure and I assume most people don't keep that in mind when writing/editing code, myself included.

Therefore I prefer the "safe way". I agree with you guys that for some cases it's tempting to put all in a single line, indeed it looks better. But then we have some inconsistency -- in some places we have separate lines, in some others we have single lines.

Will we all be able to judge when to use one or two lines? I prefer to remove that mental burden by always writing in separate lines.

I was reading a book about code reviews today, and it had a story that would fit here perfectly: "one day somebody might see that in some places we have two lines, some places one, and then decides to simplify all places to use a single line" -- that would potentially introduce subtle bugs...

2.4/root/usr/share/container-scripts/mongodb/common.sh
rhcarvalho, line 63: The `i` variable should be declared `local` :)
Otherwise it pollutes the global namespace.

2.4/root/usr/share/container-scripts/mongodb/common.sh
rhcarvalho: @mnagy it is not about unset variables, but about the exit code of `mongo admin --host ...`. In fact, it's about the evaluation of the whole right hand side of the equal sign.

2.4/root/usr/share/container-scripts/mongodb/common.sh
rhcarvalho: Given a bit more thought about it, why should we have a default value?
I believe `wait_mongo` should have a _mandatory_ argument. If $1 is not set or set to some unrecognizable value, that would be an error, and not follow some default.

2.6/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -40,47 +40,42 @@ function cache_container_addr() {
...
 
 # wait_for_mongo_down waits until the mongo server is down
 function wait_for_mongo_down() {
+  wait_mongo 0
+}
+
+# wait_mongo waits until the mongo server is up/down
+# $1 - 0 or 1 - to specify for what to wait (0 - down, 1 - up)
+# $2 - host where to connect (localhost by default)
+function wait_mongo() {
```
rhcarvalho: How about we make it clear that this is an internal function not meant to be called directly?
We could do that by either comments and/or some naming scheme.
One should only be using one of the two wrappers.

php-coder: > We could do that by either comments and/or some naming scheme.

Maybe rename it to `_wait_mongo()`?

rhcarvalho:
@omron93 I'm still a bit divided as to this change being a good idea. I like to remove code duplication, but I have some concerns.

If we have a single function and two wrappers, I imagine debugging errors will become harder.

@php-coder you've been doing a lot of debugging in this scripts recently, what do you think?

php-coder:
> @php-coder you've been doing a lot of debugging in this scripts recently, what do you think?

I like the idea of joining functions together and reduce code duplication. Also I think that `wait_mongo 1` becomes more readable with a string as argument: `wait_mongo 'UP'`. Old and unused methods could be safely removed.

omron93:
@rhcarvalho Why? The function prints what it is doing. So it is the same as original functions...

2.6/root/usr/share/container-scripts/mongodb/common.sh
omron93: It is a good idea.
What about `_wait_mongo`?

rhcarvalho:
@omron93 I think that concern comes from some recent discussion related to when you see the line where the error happened, it's not so easy to know where you're coming from... we don't get a stack trace with Bash by default.

As I said, I feel divided about it.

2.6/root/usr/share/container-scripts/mongodb/common.sh
mfojtik: why was `_wait_for_mongo` bad? ;-)

mnagy:
How will the debugging be harder? Is the concern that you will not know whether the script got stuck trying to wait for mongo to start or stop? In that case, using -x with the script will fix it, you will know what was called before (which wrapper was called). You'll need that information anyway, I assume.

omron93:
@rhcarvalho If error occurs in first 7 lines of code, then OK you don't know if the function if waiting UP or DOWN. When it fails in the FOR loop you can see the echo output in log - this differ for up/down...

rhcarvalho:
@omron93 yeah, sorry for the FUD. Please disregard my concern.

2.6/root/usr/share/container-scripts/mongodb/common.sh
omron93: You think `wait_mongo` vs `wait_for_mongo` ?

mfojtik: @omron93 yes, "wait_for_mongo" better express what that function does, right?

omron93: Renamed.

rhcarvalho:
LGTM, @mfojtik please merge

php-coder:
LGTM too.

mfojtik:
@omron93 can you rebase please?

omron93:
@mfojtik Since last commit was reverted by @bparees this PR is up to date.

mfojtik:
[test]

omron93:
@mfojtik Rebased.

omron93:
@mfojtik @bparees Please merge.

bparees:
[test]

bparees:
@php-coder ptal

php-coder:
@bparees If it passes our tests, then let's merge it. I don't see a major differences with the code that I've already reviewed 27 days ago, so still "Looks OK to me".

bparees:
thanks @php-coder, @omron93

PR #136, opened 2016-03-07
Changes in user management functions.

omron93:
This PR splits `mongo_create_users` function into two functions: `mongo_create_admin` and `mongo_create_user`. This allow to create for example only admin user.

In this functions there is also ability to specify a host where to connect to create MongoDB users and also adds a checks for env. variable needed for creating user.

This PR is a part of #112

### DIFF
--- a/2.4/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.4/root/usr/share/container-scripts/mongodb/common.sh
@@ -177,21 +177,72 @@ function run_mongod_supervisor() {
   ${CONTAINER_SCRIPTS_PATH}/replica_supervisor.sh 2>&1 &
 }
 
-# mongo_create_users creates the MongoDB admin user and the database user
-# configured by MONGODB_USER
-function mongo_create_users() {
-  mongo ${MONGODB_DATABASE} --eval "db.addUser({user: '${MONGODB_USER}', pwd: '${MONGODB_PASSWORD}', roles: [ 'readWrite' ]})"
+# mongo_create_admin creates the MongoDB admin user with password: MONGODB_ADMIN_PASSWORD
+# $1 - login parameters for mongo (optional)
+# $2 - host where to connect (localhost by default)
+function mongo_create_admin() {
+  if [[ -z "${MONGODB_ADMIN_PASSWORD:-}" ]]; then
+    echo "=> MONGODB_ADMIN_PASSWORD is not set. Authentication can not be set up."
+    exit 1
+  fi
 
-  mongo admin --eval "db.addUser({user: 'admin', pwd: '${MONGODB_ADMIN_PASSWORD}', roles: ['dbAdminAnyDatabase', 'userAdminAnyDatabase' , 'readWriteAnyDatabase','clusterAdmin' ]})"
+  # Set admin password
+  local js_command="db.addUser({user: 'admin', pwd: '${MONGODB_ADMIN_PASSWORD}', roles: ['dbAdminAnyDatabase', 'userAdminAnyDatabase' , 'readWriteAnyDatabase','clusterAdmin' ]});"
+  if ! mongo admin ${1:-} --host ${2:-"localhost"} --eval "${js_command}"; then
+    echo "=> Failed to create MongoDB admin user."
+    exit 1
+  fi
+}
 
-  touch ${MONGODB_DATADIR}/.mongodb_datadir_initialized
+# mongo_create_user creates the MongoDB database user: MONGODB_USER,
+# with password: MONGDOB_PASSWORD, inside database: MONGODB_DATABASE
+# $1 - login parameters for mongo (optional)
+# $2 - host where to connect (localhost by default)
+function mongo_create_user() {
+  # Ensure input variables exists
+  if [[ -z "${MONGODB_USER:-}" ]]; then
+    echo "=> MONGODB_USER is not set. Failed to create MongoDB user: ${MONGODB_USER}"
+    exit 1
+  fi
+  if [[ -z "${MONGODB_PASSWORD:-}" ]]; then
+    echo "=> MONGODB_PASSWORD is not set. Failed to create MongoDB user: ${MONGODB_USER}"
+    exit 1
+  fi
+  if [[ -z "${MONGODB_DATABASE:-}" ]]; then
+    echo "=> MONGODB_DATABASE is not set. Failed to create MongoDB user: ${MONGODB_USER}"
+    exit 1
+  fi
+
+  # Create database user
+  local js_command="db.getSiblingDB('${MONGODB_DATABASE}').addUser({user: '${MONGODB_USER}', pwd: '${MONGODB_PASSWORD}', roles: [ 'readWrite' ]});"
+  if ! mongo admin ${1:-} --host ${2:-"localhost"} --eval "${js_command}"; then
+    echo "=> Failed to create MongoDB user: ${MONGODB_USER}"
+    exit 1
+  fi
 }
 
 # mongo_reset_passwords sets the MongoDB passwords to match MONGODB_PASSWORD
 # and MONGODB_ADMIN_PASSWORD
+# $1 - login parameters for mongo (optional)
+# $2 - host where to connect (localhost by default)
 function mongo_reset_passwords() {
-  mongo ${MONGODB_DATABASE} --eval "db.changeUserPassword('${MONGODB_USER}', '${MONGODB_PASSWORD}')"
-  mongo admin --eval "db.changeUserPassword('admin', '${MONGODB_ADMIN_PASSWORD}')"
+  # Reset password of MONGODB_USER
+  if [[ -n "${MONGODB_USER:-}" && -n "${MONGODB_PASSWORD:-}" && -n "${MONGODB_DATABASE:-}" ]]; then
+    local js_command="db.changeUserPassword('${MONGODB_USER}', '${MONGODB_PASSWORD}')"
+    if ! mongo ${MONGODB_DATABASE} ${1:-} --host ${2:-"localhost"} --eval "${js_command}"; then
+      echo "=> Failed to reset password of MongoDB user: ${MONGODB_USER}"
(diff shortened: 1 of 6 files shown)


2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -177,21 +177,92 @@ function run_mongod_supervisor() {
...
+# $2 - host where to connect (localhost by default)
+function mongo_create_admin() {
+  if [[ -z "${MONGODB_ADMIN_PASSWORD:-}" ]]; then
+    echo "=> MONGODB_ADMIN_PASSWORD is not set. Authentication can not be set up."
+    exit 1
+  fi
+
+  # Set admin password
+  set +e
+  mongo admin ${1:-} --host ${2:-"localhost"} --eval "db.addUser({user: 'admin', pwd: '${MONGODB_ADMIN_PASSWORD}', roles: ['dbAdminAnyDatabase', 'userAdminAnyDatabase' , 'readWriteAnyDatabase','clusterAdmin' ]});"
```
rhcarvalho: @omron93 if you move this line to the if condition [a few lines below](https://github.com/openshift/mongodb/pull/136/files#diff-1763ec43870f7abf682b1057cd2daa84R195), then you don't need to `set +e; set -e`, nor the `result` variable.

Like this:

``` bash
if ! mongo admin ...; then
  echo "=> Failed to create MongoDB admin user."
  exit 1
fi
```

Since the line is pretty long, you might want to store the string we pass to `--eval` as a variable, just like you did [here](https://github.com/openshift/mongodb/pull/135/files#diff-1763ec43870f7abf682b1057cd2daa84R62). Up to you :)

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -177,21 +177,92 @@ function run_mongod_supervisor() {
...
+  local result=$?
+  set -e
+
+  if [[ ${result} -ne 0 ]]; then
+    echo "=> Failed to create MongoDB admin user."
+    exit 1
+  fi
+
+  touch ${MONGODB_DATADIR}/.mongodb_datadir_initialized
```
rhcarvalho: :skull: This is problematic, but don't worry we have a solution :bulb: 

The existence of this file communicates that the database has been initialized. Now that both functions touch the same file, we're creating a situation where we declare that it is initialized before all users are created.

The file is checked here [run-mongod#L76](https://github.com/openshift/mongodb/blob/ed4299ed23620afe23fff09f9da1d373cfcedd4a/2.4/root/usr/bin/run-mongod#L76). With this change, if only one user is created, and then there's a crash, upon restart we'll assume that the database has been initialized when it actually haven't completed fully.

To fix the problem, we should move this line outside of the functions and have it only once [after the users are created](https://github.com/openshift/mongodb/blob/ed4299ed23620afe23fff09f9da1d373cfcedd4a/2.4/root/usr/bin/run-mongod#L81).

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -177,21 +177,92 @@ function run_mongod_supervisor() {
...
+  if [[ -z "${MONGODB_PASSWORD:-}" ]]; then
+    echo "=> MONGODB_PASSWORD is not set. Failed to create MongoDB user: ${MONGODB_USER}"
+    exit 1
+  fi
+  if [[ -z "${MONGODB_DATABASE:-}" ]]; then
+    echo "=> MONGODB_DATABASE is not set. Failed to create MongoDB user: ${MONGODB_USER}"
+    exit 1
+  fi
+
+  # Crate database user
```
rhcarvalho: :star: Little typo: s/Crate/Create

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -177,21 +177,92 @@ function run_mongod_supervisor() {
...
+  set +e
+  mongo admin ${1:-} --host ${2:-"localhost"} --eval "db.getSiblingDB('${MONGODB_DATABASE}').addUser({user: '${MONGODB_USER}', pwd: '${MONGODB_PASSWORD}', roles: [ 'readWrite' ]});"
```
rhcarvalho: Same idea here about using the `if` and avoiding `set +e; -e; result` :arrow_right: https://github.com/openshift/mongodb/pull/136/files#r56188614

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -177,21 +177,92 @@ function run_mongod_supervisor() {
...
 # and MONGODB_ADMIN_PASSWORD
+# $1 - login parameters for mongo (optional)
+# $2 - host where to connect (localhost by default)
 function mongo_reset_passwords() {
-  mongo ${MONGODB_DATABASE} --eval "db.changeUserPassword('${MONGODB_USER}', '${MONGODB_PASSWORD}')"
-  mongo admin --eval "db.changeUserPassword('admin', '${MONGODB_ADMIN_PASSWORD}')"
+  # Reset password of MONGODB_USER
+  if [[ -n "${MONGODB_USER:-}" && -n "${MONGODB_PASSWORD:-}" && -n "${MONGODB_DATABASE:-}" ]]; then
+    set +e
+    mongo ${MONGODB_DATABASE} ${1:-} --host ${2:-"localhost"} --eval "db.changeUserPassword('${MONGODB_USER}', '${MONGODB_PASSWORD}')"
```
rhcarvalho: Same again about using the `if` again: https://github.com/openshift/mongodb/pull/136/files#r56188614

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -177,21 +177,92 @@ function run_mongod_supervisor() {
...
+    if [[ ${result} -ne 0 ]]; then
+      echo "=> Failed to reset password of MongoDB user: ${MONGODB_USER}"
+      exit 1
+    fi
+  fi
+
+  # Reset password of admin
+  if [[ -n "${MONGODB_ADMIN_PASSWORD:-}" ]]; then
+    set +e
+    mongo admin --eval "db.changeUserPassword('admin', '${MONGODB_ADMIN_PASSWORD}')"
```
rhcarvalho: Here too: https://github.com/openshift/mongodb/pull/136/files#r56188614

rhcarvalho:
@omron93 thanks for splitting the big rewrite into smaller parts, this is definitely easier to review :+1: 

Overall looks good, just have a suggestion about using ifs and return code of a command, a typo and a problem with marking the database as initialized. Thanks again!

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: `set` removed.

Even separate JS code is too long. However separated - it is more clear.

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: Used your workaround.

However I would like to get rid of this file in the future - mongo should be able to detect already created users.

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: Fixed.

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: Fixed.

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: Fixed.

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: Fixed.

2.4/root/usr/share/container-scripts/mongodb/common.sh
php-coder: BTW, another tiny benefit of creating `.mongodb_datadir_initialized` file outside of this function: we don't use this file in replica set configuration. But with old approach we've always created it.

omron93: Good point :+1:

rhcarvalho:
@omron93 LGTM.

@bparees @mfojtik please have a final look / merge.

mfojtik:
LGTM

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -177,21 +177,72 @@ function run_mongod_supervisor() {
...
+  if ! mongo admin ${1:-} --host ${2:-"localhost"} --eval "${js_command}"; then
+    echo "=> Failed to create MongoDB user: ${MONGODB_USER}"
+    exit 1
+  fi
 }
 
 # mongo_reset_passwords sets the MongoDB passwords to match MONGODB_PASSWORD
 # and MONGODB_ADMIN_PASSWORD
+# $1 - login parameters for mongo (optional)
+# $2 - host where to connect (localhost by default)
```
rhcarvalho, line 227: I was reviewing the code in master realized this two optional arguments currently only apply to the part that resets the MONGODB_PASSWORD password, but not MONGODB_ADMIN_PASSWORD.

@omron93 do we ever need to call `mongo_reset_passwords` with arguments?
Could you please make this consistent? Either remove the args or make them work for both calls to `mongo` below. Thanks!

omron93, line 227: @rhcarvalho Thanks for finding this. Probably these arguments won't be used. It might be used in replication to fix users with already initialized database. However it is not used now and probably wont be used in the future...

I will create a PR to remove it.

PR #137, opened 2016-03-07
Change workflow with environmental variables and config file.

omron93:
This PR stores all mongod default values in config file `/etc/mongod.conf` - to be able to read this file after `docker run` or `docker exec` and read there default or actual values. New values from OpenShift configuration options are updated in config file at the begging of run-mongod.

Enviromental variable are now declared in run-mongod to be declared in one place to improve maintainability (for example MONGODB_KEYFILE_PATH was declared twice). The values of these variables are obtained from /etc/mongod.conf.


omron93:
This PR is part of #112

rhcarvalho:
@omron93 to be honest I didn't understand very well the description :confused: . Could you please explain what is the purpose of this change?

The manipulation of the configuration file using `grep` and `sed` is much more fragile compared to the existing `envsubst` being replaced here.

omron93:
If you run this image without default command (`docker run -ti IMAGE bash`), you can't get default config options. There is no MongoDB config file.

The purpose of the redesign of the scripts in #100 is to do entrypoint very simple and move a main script in CMD - normaly executable script which can be easily run inside container. To be able to make some changes after `docker run -ti IMAGE bash` and after that run the same script as during normal start.

With this change it should be possible to change some values in MongoDB config file and and the run the same script (`run-mongod`) and have mongod working.

Without this change I don't see any benefits of having config file (since it is not available before running run-mongod, and also some mongod options are passed in the command line - so there is not everything in config file and config file do this more complicated). So it would be better to pass all mongod options in command line.

mfojtik:
[test]

mfojtik:
@omron93 i would prefer `envsubst` here as well... sed/grep are not the greatest tools for doing templating of config files.

omron93:
> Without this change I don't see any benefits of having config file (since it is not available before running run-mongod, and also some mongod options are passed in the command line - so there is not everything in config file and config file do this more complicated). So it would be better to pass all mongod options in command line.

@mfojtik @rhcarvalho So what about removing config file completely?

rhcarvalho:
@omron93 your idea of moving everything to the config file is good. It's much easier to look at the config file than having too many command line flags. I'd also suspect that not everything you can do in the config file can be done with flags (?).

I don't see a use case like you described, that a user runs the image with some alternative entrypoint/run bypassing the `envsubst`, and later wants to start mongo using the config file.
I think this alternate startup is unlikely, and even if a user does that, she can later use the provided image scripts to start mongo. Why would somebody want to do this? Perhaps there's some other improvement around that.

My proposal would be trying to get "best of both worlds", to use the config (template) for all we can (moving certain flags we pass today into the config template), and use `envsubst` to replace placeholders using environment variables. @mfojtik @omron93 agree with that?

omron93:
> I'd also suspect that not everything you can do in the config file can be done with flags (?).

Both methods are equivalent.

> your idea of moving everything to the config file is good. It's much easier to look at the config file than having too many command line flags.

Yes, that's right. However in current situation from user POV configuration is not placed in standard location and also it contains content that have to precessed by `envsubst` so does not contain values and IMPOV it has no value for user.
From devel POV I think it is similar to situation where all config options for mongod is set on one place in the script. Also not all options are easily movable to config file - for example configs for replication (in that case, there should be two config files).

> Why would somebody want to do this? Perhaps there's some other improvement around that.

For example to change some mongod configuration is same way with the same file as for other mongodb installations (from rpm,...). And then run image scripts.

rhcarvalho:
> However in current situation from user POV configuration is not placed in standard location and also it contains content that have to precessed by envsubst so does not contain values and IMPOV it has no value for user.

AFAICT, the current state reflects what worked for v2, and in that case the normal means of customizing configuration was via environment variables. I don't know if there was any way to provide your own configuration file. Please correct me if I'm wrong.

@omron93 as you pointed out, and I agree, there might be cases where a user wants to reuse an existing configuration file from other installations.

Would it be reasonable to support both scenarios?
1. If a configuration file is provided, use it as-is, try not to override anything with command line flags;
2. If no configuration file is provide, the current case today, allow customization via environment variables and use defaults where applicable.

Could that be done for both the standalone and replication cases?

How to provide a configuration file through OpenShift and plain Docker?

@omron93 @mfojtik @bparees what do you think about that?

omron93:
> Would it be reasonable to support both scenarios?

@rhcarvalho I like this idea :+1: 

> Could that be done for both the standalone and replication cases?

Replication adds two mongod options which affects standalone mongod server. So they can't be in standard config file. So it is possible to have two config file templates (one for standalone, second for replication) or add these two options in command line.

> How to provide a configuration file through OpenShift and plain Docker?

In plain Docker it could be possible to mount config file (and if we support both cases, it should be normally detected). In OpenShift I don't know. However current method will still work.

omron93:
@rhcarvalho So can I try to implement it?

rhcarvalho:
@omron93 yes, sure and thank you! A combination of config templates for the common cases and ability to drop-in a custom config file may work fine for both OpenShift and pure Docker use cases.

@grdryn would it be useful for RHMAP to be able to provide a custom config file for mongo?

omron93:
Closing in favor of #147

PR #138, opened 2016-03-09
Add ready/lifeness probes for replica

mfojtik:
@php-coder can you try this one with your extended test? I wonder if we not get into a chicken-egg problems (the service will not be "ready", meaning no endpoints will be returned in dig) till the mongostat succeed.

### DIFF
--- a/2.4/examples/replica/mongodb-clustered.json
+++ b/2.4/examples/replica/mongodb-clustered.json
@@ -119,6 +119,20 @@
               {
                 "name":  "mongodb",
                 "image": "openshift/mongodb-24-centos7",
+                "readinessProbe": {
+                  "timeoutSeconds": 1,
+                  "initialDelaySeconds": 3,
+                  "exec": {
+                    "command": [ "/bin/sh", "-i", "-c", "mongostat --host 127.0.0.1 -u admin -p $MONGODB_ADMIN_PASSWORD -n 1 --noheaders"]
+                  }
+                },
+                "livenessProbe": {
+                  "timeoutSeconds": 1,
+                  "initialDelaySeconds": 30,
+                  "tcpSocket": {
+                    "port": 27017
+                  }
+                },
                 "env": [
                   {
                     "name": "MONGODB_USER",


2.4/examples/replica/mongodb-clustered.json
```diff
@@ -119,6 +119,20 @@
               {
                 "name":  "mongodb",
                 "image": "openshift/mongodb-24-centos7",
+                "readinessProbe": {
+                  "timeoutSeconds": 1,
+                  "initialDelaySeconds": 3,
+                  "exec": {
+                    "command": [ "/bin/sh", "-i", "-c", "mongostat --host 127.0.0.1 -u admin -p $MONGODB_ADMIN_PASSWORD -n 1 --noheaders"]
```
omron93, line 126: Why did you choose `mongostat`?

php-coder:
> @php-coder can you try this one with your extended test?

@mfojtik After merging my tests you can easily test your changes just by running test on you laptop/Jenkins.

2.4/examples/replica/mongodb-clustered.json
mfojtik, line 126: @omron93 seen used that in mongodb documentation for checking the state of mongo... i think any other `mongo` command will work fine.

omron93, line 126: Since mongodb 2.6 mongo tools are developed in separate project. It is written in golang and it is quite large. mongo-tools package is not shipped in current images and with it image could be 80MB larger.

So I think it is better to use commands already installed in image. `mongo` shell,  `mongosniff` or `mongoperf`.

mfojtik, line 126: @omron93 good to know, I will change that to simple mongo call... Can you think of any simple mongo command to check that the DB/mongo is ready for accepting connections/storing data?

omron93:
@mfojtik What is wrong with current readinessProbe in 2.4 branch?

From kubernetes doc - http://kubernetes.io/docs/user-guide/pod-states/#container-probes
ReadinessProbe is a function to ensure that pod should be listed in service endpoits, am I right? In MongoDB case in current implementation should be included whet it accepts connections (port 27017) and after that it is added into replicaset - since this operation is not finished MongoDB is not "ready" (are not part of replset and does not store data), however it have to be listed in service...

2.4/examples/replica/mongodb-clustered.json
omron93, line 126: @mfojtik It depends on the purpose of this call. To simply detect that MongoDB accepts connection - `mongo --eval "quit()"`, in case of replication it is complicated.

bparees:
@rhcarvalho @php-coder i'm closing this since the old replica example is being deprecated, but please take a look and see if we should be pulling any of this logic into the petset replica example.

2.4/examples/replica/mongodb-clustered.json
```diff
@@ -119,6 +119,20 @@
               {
                 "name":  "mongodb",
                 "image": "openshift/mongodb-24-centos7",
+                "readinessProbe": {
+                  "timeoutSeconds": 1,
```
rhcarvalho, line 123: FTR, timeouts are ignored in Exec probes: https://github.com/kubernetes/kubernetes/issues/26895

PR #139, opened 2016-03-14
Don't stop the script when unable to determine ip of the primary node

php-coder:
When we are trying to add or remove MongoDB to/from replica set and we didn't determine IP-address of the primary node we shouldn't exit immediately from the whole script because there is also the code that should be executed always (for example, we are invoking `mongo_remove` from the cleanup method before killing the mongod process).

Other bash improvements/fixes:
- fixed a bug when we're comparing with a static string (`[ "$node" == container_addr ]`)
- use local keyword in more places
- remove `set +e/`set -e` hack
- prefer `-n` over `! -z`

PTAL @bparees @rhcarvalho

### DIFF
--- a/2.4/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.4/root/usr/share/container-scripts/mongodb/common.sh
@@ -45,17 +45,14 @@ function wait_for_mongo_up() {
   local mongo_cmd
   mongo_cmd="mongo admin "
 
-  if [ ! -z "${mongo_host}" ]; then
+  if [ -n "${mongo_host}" ]; then
     mongo_cmd+="--host ${mongo_host}:${CONTAINER_PORT} "
   fi
 
+  local i
   for i in $(seq $MAX_ATTEMPTS); do
     echo "=> Waiting for MongoDB service startup ${mongo_host} ..."
-    set +e
-    $mongo_cmd --eval "help" &>/dev/null
-    status=$?
-    set -e
-    if [ $status -eq 0 ]; then
+    if ! $mongo_cmd --eval 'help' &>/dev/null; then
       echo "=> MongoDB service has started"
       return 0
     fi
@@ -67,13 +64,10 @@ function wait_for_mongo_up() {
 
 # wait_for_mongo_down waits until the mongo server is down
 function wait_for_mongo_down() {
+  local i
   for i in $(seq $MAX_ATTEMPTS); do
     echo "=> Waiting for MongoDB service shutdown ..."
-    set +e
-    mongo admin --eval "help" &>/dev/null
-    status=$?
-    set -e
-    if [ $status -ne 0 ]; then
+    if ! mongo admin --eval 'help' &>/dev/null; then
       echo "=> MongoDB service has stopped"
       return 0
     fi
@@ -103,10 +97,12 @@ function build_mongo_config() {
   member_id=1
   local container_addr
   container_addr="$(container_addr)"
+  local node
   for node in ${current_endpoints}; do
-    [ "$node" == container_addr ] && continue
-    members+="{ _id: ${member_id}, host: \"${node}:${CONTAINER_PORT}\"},"
-    let member_id++
+    if [ "$node" != "$container_addr" ]; then
+      members+="{ _id: ${member_id}, host: \"${node}:${CONTAINER_PORT}\"},"
+      let member_id++
+    fi
   done
   echo -n "var config={ _id: \"${MONGODB_REPLICA_NAME}\", members: [ ${members%,} ] }"
 }
@@ -152,7 +148,12 @@ function mongo_primary_member_addr() {
 # mongo_remove removes the current MongoDB from the cluster
 function mongo_remove() {
   local primary_addr
-  primary_addr="$(mongo_primary_member_addr)"
+  # if we cannot determine the IP address of the primary, exit without an error
+  # to allow callers to proceed with their logic
+  primary_addr="$(mongo_primary_member_addr || true)"
+  if [ -z "$primary_addr" ]; then
+    return
+  fi
 
   local mongo_addr
   mongo_addr="$(mongo_addr)"
@@ -165,7 +166,12 @@ function mongo_remove() {
 # mongo_add advertise the current container to other mongo replicas
 function mongo_add() {
   local primary_addr
-  primary_addr="$(mongo_primary_member_addr)"
+  # if we cannot determine the IP address of the primary, exit without an error
+  # to allow callers to proceed with their logic
+  primary_addr="$(mongo_primary_member_addr || true)"
+  if [ -z "$primary_addr" ]; then
+    return
+  fi
 
   local mongo_addr
   mongo_addr="$(mongo_addr)"
--- a/2.6/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/common.sh
@@ -45,17 +45,14 @@ function wait_for_mongo_up() {
   local mongo_cmd
   mongo_cmd="mongo admin "
 
-  if [ ! -z "${mongo_host}" ]; then
+  if [ -n "${mongo_host}" ]; then
     mongo_cmd+="--host ${mongo_host}:${CONTAINER_PORT} "
   fi
 
+  local i
   for i in $(seq $MAX_ATTEMPTS); do
     echo "=> Waiting for MongoDB service startup ${mongo_host} ..."
-    set +e
-    $mongo_cmd --eval "help" &>/dev/null
-    status=$?
-    set -e
-    if [ $status -eq 0 ]; then
+    if ! $mongo_cmd --eval 'help' &>/dev/null; then
       echo "=> MongoDB service has started"
       return 0
     fi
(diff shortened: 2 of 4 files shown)


rhcarvalho:
@php-coder part of these changes conflict with https://github.com/openshift/mongodb/pull/135 (`wait_for_mongo_up` & `wait_for_mongo_down` functions).

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -151,8 +150,12 @@ function mongo_primary_member_addr() {
 
 # mongo_remove removes the current MongoDB from the cluster
 function mongo_remove() {
-  local primary_addr
-  primary_addr="$(mongo_primary_member_addr)"
+  # Declare and assign variable in the single line to ignore possible non-zero
+  # return code from mongo_primary_member_addr()
+  local primary_addr="$(mongo_primary_member_addr)"
```
rhcarvalho: This is a subtle hack...
If we are to ignore an error, please let's at least ignore it explicitly.

2.6/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -67,12 +66,11 @@ function wait_for_mongo_up() {
...
 function wait_for_mongo_down() {
+  local i
   for i in $(seq $MAX_ATTEMPTS); do
     echo "=> Waiting for MongoDB service shutdown ..."
-    set +e
-    mongo admin --eval "help" &>/dev/null
-    status=$?
-    set -e
+    local status=0
+    mongo admin --eval 'help' &>/dev/null || status=$?
```
rhcarvalho: This works, but seems unnecessarily complex to me.
If the `status=$?` part is ever executed, we know that the if condition in the next line will be always true.

We don't need status if we execute `mongo` as the if condition. Suggestion:

``` bash
# wait_for_mongo_down waits until the mongo server is down
function wait_for_mongo_down() {
  local i
  for i in $(seq $MAX_ATTEMPTS); do
    echo "=> Waiting for MongoDB service shutdown ..."
    if ! mongo admin --eval "help" &>/dev/null; then
      echo "=> MongoDB service has stopped"
      return 0
    fi
    sleep $SLEEP_TIME
  done
  echo "=> Giving up: Failed to stop MongoDB service"
  exit 1
}
```

I don't know what's the canonical way to format the code when using "!" (not) in front of a command...

In pseudocode, maybe I'd even find it easier to reason about something flatter like:

```
print "Waiting ..."
tries = 5
while mongo_is_up() && tries > 0:
   sleep 1
   tries--
if mongo_is_up():
   print "Giving up..."
   return 1
else
   print "Mongo has stopped"
   return 0
```

But again this code is also subject of another PR...

omron93: +1 to @rhcarvalho  that handling this in IF condition is the best solution.

2.6/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -151,8 +150,12 @@ function mongo_primary_member_addr() {
 
 # mongo_remove removes the current MongoDB from the cluster
 function mongo_remove() {
-  local primary_addr
-  primary_addr="$(mongo_primary_member_addr)"
+  # Declare and assign variable in the single line to ignore possible non-zero
```
omron93: @php-coder How this "hack" with `local` is better that "`set +e/set -e` hack"?

Using `set` is more expressive IMPOV.

2.6/root/usr/share/container-scripts/mongodb/common.sh
php-coder: Thanks, I've added `if`.

2.4/root/usr/share/container-scripts/mongodb/common.sh
php-coder: Thanks! Added explicit ignoring.

php-coder:
> part of these changes conflict with #135 (wait_for_mongo_up & wait_for_mongo_down functions).

Yes, I know. I will update my PR if #135 will be merged first. Otherwise @omron93 updates his PR.

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -103,10 +97,12 @@ function build_mongo_config() {
   member_id=1
   local container_addr
   container_addr="$(container_addr)"
+  local node
   for node in ${current_endpoints}; do
-    [ "$node" == container_addr ] && continue
-    members+="{ _id: ${member_id}, host: \"${node}:${CONTAINER_PORT}\"},"
-    let member_id++
+    if [ "$node" != "$container_addr" ]; then
```
rhcarvalho, line 102: Nice! :balloon: :tada:

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -45,17 +45,14 @@ function wait_for_mongo_up() {
...
 
+  local i
   for i in $(seq $MAX_ATTEMPTS); do
     echo "=> Waiting for MongoDB service startup ${mongo_host} ..."
-    set +e
-    $mongo_cmd --eval "help" &>/dev/null
-    status=$?
-    set -e
-    if [ $status -eq 0 ]; then
+    if ! $mongo_cmd --eval 'help' &>/dev/null; then
```
rhcarvalho, line 55: :+1:

2.6/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -151,8 +147,10 @@ function mongo_primary_member_addr() {
 
 # mongo_remove removes the current MongoDB from the cluster
 function mongo_remove() {
-  local primary_addr
-  primary_addr="$(mongo_primary_member_addr)"
+  local primary_addr="$(mongo_primary_member_addr ||  true)"
```
rhcarvalho: @php-coder how about we get this back to 2 lines with the explicit `|| true`?
It would be nice to have a comment/reminder as to why we are ignoring the exit code.

In this particular line here you have an extra space between the `||` and `true`.

php-coder: Thanks, updated.

I can add any comment that you will like, because I don't know what to write here. After adding `|| true` code became self-readable for me and I wouldn't add comment here.

rhcarvalho: The comment is not to say "ignore result of the command", but why are we doing it.

Why are we doing it? Why is `|| true` necessary?

2.4/root/usr/share/container-scripts/mongodb/common.sh
rhcarvalho, line 102: I just looked at this again and wow... we were always comparing "$node" to "container_addr" (the string), crazy... that would never match...

What's the consequence of that?

2.6/root/usr/share/container-scripts/mongodb/common.sh
php-coder: > How this "hack" with local is better that "set +e/set -e hack"?

`set -e` **always** enables strict mode, even if it wasn't activated before. Also using `||` looks more idiomatic (at least for me).

2.4/root/usr/share/container-scripts/mongodb/common.sh
mfojtik, line 102: @rhcarvalho no consequence... you basically add the container itself to the config, mongo will either refuse it or just drop.

php-coder, line 102: > What's the consequence of that?

We was always add our host twice to the list of members of replica set. I'm not sure but probably it didn't cause any problems, otherwise we would found it earlier.

omron93, line 102: @rhcarvalho is right. The IP of post-deploy-hook is not returned by `endpoints`, is it?

rhcarvalho, line 102: Indeed looks like in practice this was not a big deal.

> The IP of post-deploy-hook is not returned by endpoints, is it?

I believe it is not. And I believe it doesn't matter here. When we initialize the replica set the pod running the hook is the PRIMARY, and later it steps down.

MongoDB has been tolerant to us :smile:

omron93, line 102: > I believe it is not. And I believe it doesn't matter here. When we initialize the replica set the pod running the hook is the PRIMARY, and later it steps down.

And this pod runs the initialization. So the compating with container_addr is redundant - post-hook-pod IP (=container_addr) won't be in current_endpoints...

mfojtik, line 102: hook pods are not labeled to match the service, so they are not added as endpoints

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -152,7 +148,11 @@ function mongo_primary_member_addr() {
 # mongo_remove removes the current MongoDB from the cluster
 function mongo_remove() {
   local primary_addr
-  primary_addr="$(mongo_primary_member_addr)"
+  # we're ignoring exit code because without it, next if never gets executed
```
rhcarvalho: As we talked on IRC, this comment is not very helpful because the next if is being introduced in this same patch.
AFAIU the reason we're doing this is to circumvent `set -o errexit`, ignore the error and allow the caller to proceed as if nothing has happened.

A concern I have is that whatever code calls `mongo_remove` or `mongo_add` cannot tell if it effectively removed or added a node from/to the replica set, or if the function terminated earlier.

It's more like they should be called `maybe_remove_from_replica_set` and `maybe_add_to_replica_set`.

php-coder: Done.

rhcarvalho:
@php-coder excellent! Very good catches, thanks for improving the scripts. LGTM.

@mfojtik could you please merge this?

PR #141, opened 2016-03-16
Revert "Don't stop the script when unable to determine ip of the primary node"

bparees:
Reverts openshift/mongodb#139

### DIFF
--- a/2.4/root/usr/bin/run-mongod
+++ b/2.4/root/usr/bin/run-mongod
@@ -41,7 +41,7 @@ function unset_env_vars() {
 }
 
 function cleanup() {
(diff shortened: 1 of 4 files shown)


bparees:
near as i can tell the original PR never passed tests before being merged, and the push job is now failing.  reverting.

@mfojtik @rhcarvalho @php-coder

php-coder:
@bparees What I should do to merge it back? What and where I have to fix?

bparees:
@php-coder the mongo 2.4 tests were failing, so you'll need to unrevert your change, test it locally (or via a PR), figure out why it's failing the tests, fix it, and then we can merge it again.

PR #142, opened 2016-03-16
Don't stop the script when unable to determine ip of the primary

php-coder:
(Copy&pasted from #139 because it's the same PR in essence.)

When we are trying to add or remove MongoDB to/from replica set and we didn't determine IP-address of the primary node we shouldn't exit immediately from the whole script because there is also the code that should be executed always (for example, we are invoking mongo_remove from the cleanup method before killing the mongod process).

Other bash improvements/fixes:
- fixed a bug when we're comparing with a static string (`[ "$node" == container_addr ]`)
- use local keyword in more places
- remove `set +e`/`set -e` hack
- prefer `-n` over `! -z`

PTAL @bparees @rhcarvalho

### DIFF
--- a/2.4/root/usr/bin/run-mongod
+++ b/2.4/root/usr/bin/run-mongod
@@ -41,7 +41,7 @@ function unset_env_vars() {
 }
 
 function cleanup() {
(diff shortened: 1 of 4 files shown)


bparees:
lgtm.

PR #143, opened 2016-03-18
Fix replica set example by specifying for how many members we need to wait

php-coder:
Add capability to specify to post-deployment hook for how many members we need to wait before beginning to create replica set. This shouldn't break anything because I left old code and added warning to it. In other hand it fixes replica set example on my laptop.

One downside: we should specify count of replicas twice.

PTAL @bparees @rhcarvalho @mfojtik

### DIFF
--- a/2.4/examples/replica/mongodb-clustered.json
+++ b/2.4/examples/replica/mongodb-clustered.json
@@ -94,7 +94,13 @@
                 "command": [
                   "run-mongod", "initiate"
                 ],
-                "containerName": "mongodb"
+                "containerName": "mongodb",
+                "env": [
+                  {
+                    "name": "MONGODB_INITIAL_REPLICA_COUNT",
+                    "value": "3"
+                  }
+                ]
               }
             }
           }
--- a/2.6/root/usr/share/container-scripts/mongodb/initiate_replica.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/initiate_replica.sh
@@ -6,13 +6,24 @@ set -o pipefail
 
 source  ${CONTAINER_SCRIPTS_PATH}/common.sh
 
-echo -n "=> Waiting for MongoDB endpoints ..."
-
 current_endpoints=$(endpoints)
-while [ -z "${current_endpoints}" ]; do
-  sleep 2
-  current_endpoints=$(endpoints)
-done
+if [ -n "${MONGODB_INITIAL_REPLICA_COUNT:-}" ]; then
+  echo -n "=> Waiting for $MONGODB_INITIAL_REPLICA_COUNT MongoDB endpoints ..."
+  while [[ "$(echo "${current_endpoints}" | wc -l)" -lt ${MONGODB_INITIAL_REPLICA_COUNT} ]]; do
+    sleep 2
+    current_endpoints=$(endpoints)
+  done
+else
+  echo "Attention: MONGODB_INITIAL_REPLICA_COUNT is not set and it could lead to a improperly configured replica set."
+  echo "To fix this, set MONGODB_INITIAL_REPLICA_COUNT variable to the number of members in the replica set in"
+  echo "the configuration of post deployment hook."
+
+  echo -n "=> Waiting for MongoDB endpoints ..."
+  while [ -z "${current_endpoints}" ]; do
+    sleep 2
+    current_endpoints=$(endpoints)
+  done
+fi
 echo "${current_endpoints}"
 
 # Let initialize the first member of the cluster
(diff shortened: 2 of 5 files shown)


2.4/examples/replica/mongodb-clustered.json
```diff
@@ -94,7 +94,13 @@
                 "command": [
                   "run-mongod", "initiate"
                 ],
-                "containerName": "mongodb"
+                "containerName": "mongodb",
+                "env": [
+                  {
+                    "name": "MONGODB_REPLICA_NODES",
+                    "value": "3"
```
bparees, line 101: what actually ensures that we get this number of replica endpoints?

php-coder, line 101: @bparees [Here](https://github.com/openshift/mongodb/pull/143/files/b43f319ed916a1c9c108b349971b90b52dc87c23#diff-25d0fbd40ccee5d2d449d8824f9038c9L107) we're specifying desired number of replicas.

bparees, line 101: I think you need to rename this env var.  Right now it sounds like by setting it, I am controlling the number of replicas I will get, but you're not, you're setting the number of replicas to expect/wait for.

php-coder, line 101: @bparees I've renamed it to `MONGODB_NODES_TO_WAIT`. If you can suggest better name I would be glad.

rhcarvalho:
If we really go this way we need at least documentation explaining what that env var is for, how to use it, when to change it, what happens when you scale up/down, etc.

I'd rather see us improving the example by cleaning up the technical debt / reviewing the architecture rather than introducing what looks like a hack.

rhcarvalho:
@php-coder there are docs for the replication example in at least two places:
- https://github.com/openshift/mongodb/blob/master/2.4/examples/replica/README.md
- https://docs.openshift.org/latest/using_images/db_images/mongodb.html#using-mongodb-replication

But before you invest time documenting it, I'd like to understand better why we're doing this...

2.4/examples/replica/mongodb-clustered.json
```diff
@@ -94,7 +94,13 @@
                 "command": [
                   "run-mongod", "initiate"
                 ],
-                "containerName": "mongodb"
+                "containerName": "mongodb",
+                "env": [
+                  {
+                    "name": "MONGODB_NODES_TO_WAIT",
```
mfojtik: is this only for the initial template instantiation? or user will have to modify this every time the "replicas" value is changed?

mfojtik: if former, I would name this `INITIAL_REPLICA_COUNT` or something similar.

php-coder: @mfojtik yes, it's only for the replica set initialization.

> if former, I would name this INITIAL_REPLICA_COUNT or something similar.

Just `INITIAL_REPLICA_COUNT` or `MONGODB_INITIAL_REPLICA_COUNT`?

mfojtik: I prefer shorter, but I will leave decision on you ;-) MONGODB_ is good for namespacing ;-)

php-coder:
@rhcarvalho I've updated `README.md` file. Because [MongoDB doc is outdated](https://github.com/openshift/openshift-docs/issues/1775), it's hard to update it quickly because it even doesn't describe usage of post hook. I suggest to document this parameter as part of documenting post-deployment hook.

php-coder:
@bparees What else I need to fix/improve to get it merged?

2.6/root/usr/share/container-scripts/mongodb/initiate_replica.sh
```diff
@@ -6,13 +6,24 @@ set -o pipefail
...
-  current_endpoints=$(endpoints)
-done
+if [ -n "${MONGODB_INITIAL_REPLICA_COUNT:-}" ]; then
+  echo -n "=> Waiting for $MONGODB_INITIAL_REPLICA_COUNT MongoDB endpoints ..."
+  while [[ "$(echo "${current_endpoints}" | wc -l)" -lt ${MONGODB_INITIAL_REPLICA_COUNT} ]]; do
+    sleep 2
+    current_endpoints=$(endpoints)
+  done
+else
+  echo "Attention: MONGODB_INITIAL_REPLICA_COUNT is not set and it could lead to non-properly configured replica set."
```
bparees: s/to non-properly/to a improperly/

bparees:
@php-coder one nit, and also i thought you can @rhcarvalho were discussing a different approach that wouldn't require this, did that not work out?

2.6/root/usr/share/container-scripts/mongodb/initiate_replica.sh
php-coder: Thanks, I've fixed it.

php-coder:
> i thought you can @rhcarvalho were discussing a different approach that wouldn't require this, did that not work out?

@bparees On Scrum meeting @mfojtik said that this approach requires extra permissions that user should grant with a special command. As far I understood, we don't want to add this complexity to the current work flow and it means that this approach won't be accepted later. Correct me if I'm wrong.

bparees:
@php-coder ah, i missed that the discussion we had about retrieving container lists was related to what you wanted to do here.  agreed that we don't want to do that.

PR #146, opened 2016-04-18
Allow to set MONGO_TEXT_SEARCH_ENABLED for mongo 2.4

mfojtik:


mfojtik:
[test]

2.4/README.md
```diff
@@ -25,6 +25,9 @@ The following environment variables influence the MongoDB configuration file. Th
 |  `MONGODB_NOPREALLOC` | Disable data file preallocation.                                          |  true
 |  `MONGODB_SMALLFILES` | Set MongoDB to use a smaller default data file size.                      |  true
 |  `MONGODB_QUIET`      | Runs MongoDB in a quiet mode that attempts to limit the amount of output. |  true
+|  `MONGODB_TEXT_SEARCH_ENABLED` | Enables the [text search](https://docs.mongodb.org/v2.4/core/index-text/#text-search-text-command) feature.  |  false
+
+**Notice: The `MONGODB_TEXT_SEARCH_ENABLED` option is deprecated in MongoDB 2.6.**
```
mfojtik: @bparees first off, this option is enabled by default in mongo 2.6 which will change our default to "true" there and cause inconsistency between image functionality....  after talking to @omron93 I would rather opt-out this option for 2.6 and wait if somebody requests disabling.

mfojtik:
[test]

mfojtik:
@bparees test passed (I also added special test for this option :-)

2.4/README.md
bparees: it's not deprecated in 2.6, it doesn't exist (the env var is not respected).

omron93: @bparees It is deprecated in MongoDB https://docs.mongodb.org/v2.6/reference/parameters/#param.textSearchEnabled and enabled by default.

It makes sense to enable textSearch. Does someone requested to be able to disable it?
MongoDB 2.4 image may be EOLed (https://github.com/openshift/mongodb/pull/127#issuecomment-185800731) and newer version have it enable by default.

IMPOV having this option only for consistency with 3 years old version which we don't know how long it will be supported by OpenShift does not worth adding more and more options to images and increasing complexity of images.

bparees: @omron93 my point was that our 2.6 image doesn't respect the MONGODB_TEXT_SEARCH_ENABLED env var, so the readme should not imply that it does  (deprecated implies supported/functional)

so what it seems like we should/will have:
2.4 - default(from mongo itself) is off, env variable to turn it on
2.6 - default(from mongo itself) is on, no option to turn it off currently.

2.4/test/run
```diff
@@ -145,11 +145,34 @@ function test_config_option() {
     docker stop $(get_cid ${name})
 }
 
+function test_text_search() {
```
rhcarvalho: This test is weird to me, it will pass even if we remove the code that uses MONGODB_TEXT_SEARCH_ENABLED. Shouldn't we test for something specific to having that value set to true?

2.4/README.md
rhcarvalho: Maybe include a note here that this is not recommended to use in production? (As the warning in https://docs.mongodb.org/v2.4/reference/parameters/#param.textSearchEnabled)

2.4/test/run
omron93:
``` bash
# need to set these because `mongo_cmd` relies on global variables
USER=admin
PASS=adminPassword
DB=admin

test_connection ${name}
mongo_local_cmd ${name} 'admin --quiet --eval "JSON.stringify(db.runCommand({getParameter: 1, "textSearchEnabled":1}))"' | grep '"textSearchEnabled":true'
```

bparees:
[test]

rhcarvalho:
I'm missing a PR description to understand why we're adding this, and some resolution to comments that were not addressed.

2.4/README.md
mfojtik: @bparees OK, I removed the notice saying this option is deprecated in 2.6, it will simply not exist there. 

I added the warning for not using this in production systems.

2.4/test/run
mfojtik: I like @omron93 suggestion... I don't think we need to build complex test, the point of my test was just to verify nothing breaks when you set that variable up and mongo will start successfully.

mfojtik:
[test]

mfojtik:
@rhcarvalho https://trello.com/c/XGCb85EF/908-1-allow-enabling-text-search-with-mongodb-image

bparees:
[test]

bparees:
@mfojtik the new test appears to be failing.

omron93:
@mfojtik `mongo_cmd_local` also needs `-u admin -p $MONGODB_ADMIN_PASSWORD`.

Also  #162 is fixing related problem.

bparees:
[test]

bparees:
@mfojtik still no good

bparees:
@mfojtik buuuump

mfojtik:
@bparees sooory for the delay... rebase is killing me :(( looking at this right now

bparees:
@mfojtik sorry.  it's really not that critical, just an item on my todo list that i'm trying to get rid of :)

mfojtik:
@bparees believe me, I'm trying to get rid of this as well :))

mfojtik:
OK @bparees , I'm giving up on the test... tried 7 different ways to make it pass, but I can't get thought.
I tested this locally and it works, I can follow up with test at some point ;-) Or will ask @mnagy or @rhcarvalho ;)

bparees:
can you at least add a test that confirms the container starts successfully when the value is set to true?  i don't care if you confirm the value is set, just set the env variable and make sure it comes up.

mfojtik:
@bparees ok, I hope this will do it :-)

bparees:
woo.

PR #17, opened 2015-04-09
Add MongoDB replica set example

mfojtik:
TODO:
- [x] Add authentication back and use keyFile for inter-mongo communication
- [ ] Re-create the replicaSet if all Pods died, but there is a persistent storage
- [x] Add more documentation about how this works

PR #147, opened 2016-04-18
Added ability to use user config file.

omron93:
As discussed in #137 

Added ability to use user config file.
- user is able to create or mount custom config file  in `/etc/mongod.conf` (for example from non-docker MongoDB deployments)
- this config file does not affect MONGODB_DATADIR and MONGODB_REPLICA_NAME

Cleaned up MONGODB env variables and mongod parameters.
- all env var declations were moved into one place
- removed PIDFILE option (mongod stores PID into mongod.lock file in dbpath)
- fixed double declaration of MONGODB_KEYFILE_PATH

### DIFF
--- a/2.4/README.md
+++ b/2.4/README.md
@@ -58,6 +58,14 @@ will be started. If you are re-attaching the volume to another container, the
 creation of the database user and admin user will be skipped and only the
 MongoDB daemon will be started.
 
+Custom configuration file
+---------------------------------
+
+It is allowed to use custom configuration file for mongod server. Providing a custom configuration file supercedes the individual configuration environment variable values.
+
+To use custom configuration file in container it has to be mounted into `/etc/mongod.conf`. For example to use configuration file stored in `/home/user` directory use this option for `docker run` command: `-v /home/user/mongod.conf:/etc/mongod.conf:Z`.
+
+**Notice: Custom config file does not affect name of replica set. It has to be set in `MONGODB_REPLICA_NAME` environment variable.**
 
 MongoDB admin user
 ---------------------------------
--- a/2.4/root/usr/bin/run-mongod
+++ b/2.4/root/usr/bin/run-mongod
@@ -6,20 +6,6 @@ set -o pipefail
 
 source ${CONTAINER_SCRIPTS_PATH}/common.sh
 
-# Data directory where MongoDB database files live. The data subdirectory is here
-# because mongodb.conf lives in /var/lib/mongodb/ and we don't want a volume to
-# override it.
-export MONGODB_DATADIR=/var/lib/mongodb/data
-
-# Configuration settings.
-export MONGODB_NOPREALLOC=${MONGODB_NOPREALLOC:-true}
-export MONGODB_SMALLFILES=${MONGODB_SMALLFILES:-true}
-export MONGODB_QUIET=${MONGODB_QUIET:-true}
-export MONGODB_TEXT_SEARCH_ENABLED=${MONGODB_TEXT_SEARCH_ENABLED:-false}
-
-export MONGODB_KEYFILE_SOURCE_PATH="/var/run/secrets/mongo/keyfile"
-export MONGODB_KEYFILE_PATH="${HOME}/keyfile"
-
 function usage() {
   echo "You must specify the following environment variables:"
   echo "  MONGODB_USER"
@@ -61,12 +47,15 @@ if [ $# -gt 0 ] && [ "$1" == "initiate" ]; then
   exec ${CONTAINER_SCRIPTS_PATH}/initiate_replica.sh
 fi
 
-# Generate config file for MongoDB
-envsubst < ${CONTAINER_SCRIPTS_PATH}/mongodb.conf.template > $MONGODB_CONFIG_PATH
+# If user provides own config file use it and do not generate new one
+if [ ! -s $MONGODB_CONFIG_PATH ]; then
+  # Generate config file for MongoDB
+  envsubst < ${CONTAINER_SCRIPTS_PATH}/mongodb.conf.template > $MONGODB_CONFIG_PATH
+fi
 
 # Need to cache the container address for the cleanup
 cache_container_addr
-mongo_common_args="-f $MONGODB_CONFIG_PATH --oplogSize 64"
+mongo_common_args="-f $MONGODB_CONFIG_PATH"
 if [ -z "${MONGODB_REPLICA_NAME-}" ]; then
   if ! [[ -v MONGODB_USER && -v MONGODB_PASSWORD && -v MONGODB_DATABASE && -v MONGODB_ADMIN_PASSWORD ]]; then
     usage
@@ -110,8 +99,7 @@ else
   # `cleanup` references it.
   (
     unset_env_vars
-    mongod $mongo_common_args --replSet "${MONGODB_REPLICA_NAME}" \
-      --keyFile "${MONGODB_KEYFILE_PATH}"
+    mongod $mongo_common_args --replSet "${MONGODB_REPLICA_NAME}"
   ) &
   wait
 fi
--- a/2.4/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.4/root/usr/share/container-scripts/mongodb/common.sh
@@ -4,14 +4,23 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-# Used for wait_for_mongo_* functions
-MAX_ATTEMPTS=60
-SLEEP_TIME=1
-
-export MONGODB_CONFIG_PATH=/etc/mongod.conf
-export MONGODB_PID_FILE=/var/lib/mongodb/mongodb.pid
-export MONGODB_KEYFILE_PATH=/var/lib/mongodb/keyfile
+# Data directory where MongoDB database files live. The data subdirectory is here
+# because mongodb.conf lives in /var/lib/mongodb/ and we don't want a volume to
+# override it.
+export MONGODB_DATADIR=/var/lib/mongodb/data
 export CONTAINER_PORT=27017
+# Configuration settings.
+export MONGODB_NOPREALLOC=${MONGODB_NOPREALLOC:-true}
+export MONGODB_SMALLFILES=${MONGODB_SMALLFILES:-true}
+export MONGODB_QUIET=${MONGODB_QUIET:-true}
+export MONGODB_TEXT_SEARCH_ENABLED=${MONGODB_TEXT_SEARCH_ENABLED:-false}
+
+MONGODB_CONFIG_PATH=/etc/mongod.conf
+MONGODB_KEYFILE_PATH="${HOME}/keyfile"
+
+# Constants used for waiting
+readonly MAX_ATTEMPTS=60
+readonly SLEEP_TIME=1
 
 # container_addr returns the current container external IP address
 function container_addr() {
@@ -251,11 +260,17 @@ function mongo_reset_admin() {
 
 # setup_keyfile fixes the bug in mounting the Kubernetes 'Secret' volume that
 # mounts the secret files with 'too open' permissions.
+# add --keyFile argument to mongo_common_args
 function setup_keyfile() {
+  # If user specify keyFile in config file do not use generated keyFile
+  if grep -q "^\s*keyFile" ${MONGODB_CONFIG_PATH}; then
+    exit 0
+  fi
   if [ -z "${MONGODB_KEYFILE_VALUE-}" ]; then
     echo "ERROR: You have to provide the 'keyfile' value in MONGODB_KEYFILE_VALUE"
     exit 1
   fi
   echo ${MONGODB_KEYFILE_VALUE} > ${MONGODB_KEYFILE_PATH}
   chmod 0600 ${MONGODB_KEYFILE_PATH}
+  mongo_common_args+=" --keyFile ${MONGODB_KEYFILE_PATH}"
 }
--- a/2.4/root/usr/share/container-scripts/mongodb/mongodb.conf.template
+++ b/2.4/root/usr/share/container-scripts/mongodb/mongodb.conf.template
@@ -1,7 +1,6 @@
 # mongodb.conf
 
-port = 27017
-pidfilepath = ${MONGODB_PID_FILE}
+port = ${CONTAINER_PORT}
 
 # Set this value to designate a directory for the mongod instance to store its data.
 # Default: /var/lib/mongodb/data
@@ -19,6 +18,9 @@ quiet = ${MONGODB_QUIET}
 
 # Disable the HTTP interface (Defaults to localhost:28017).
 nohttpinterface = true
+
+# Size to use (in MB) for replication op log (default 5% of disk space)
+oplogSize = 64
 
 # If the flag is not enabled, you cannot create new text indexes, and you cannot perform text searches. 
 # Since it is not recommended to have this on in production, the default is "false".
--- a/2.4/test/run
+++ b/2.4/test/run
@@ -295,9 +295,57 @@ function run_change_password_test() {
     echo "  Success!"
 }
 
+function run_mount_config_test() {
+    local name="mount_config"
+    echo "  Testing config file mount"
+    local database='db'
+    local user='user'
+    local password='password'
+    local admin_password='adminPassword'
+
+    local volume_dir
+    volume_dir=`mktemp -d --tmpdir mongodb-testdata.XXXXX`
+    chmod a+rwx ${volume_dir}
+    config_file=$volume_dir/mongod.conf
+    echo "dbpath=/var/lib/mongodb/dbpath
+unixSocketPrefix = /var/lib/mongodb
+smallfiles = true
+noprealloc = true" > $config_file
+    chmod a+r ${config_file}
+
+    DOCKER_ARGS="
+-e MONGODB_DATABASE=${database}
+-e MONGODB_USER=${user}
+-e MONGODB_PASSWORD=${password}
+-e MONGODB_ADMIN_PASSWORD=${admin_password}
+-v ${config_file}:/etc/mongod.conf:Z
+-v ${volume_dir}:/var/lib/mongodb/dbpath:Z
+"
+    create_container ${name} ${DOCKER_ARGS}
+
+    # need to set these because `mongo_cmd` relies on global variables
+    USER=${user}
+    PASS=${password}
+    DB=${database}
+
+    # need this to wait for the container to start up
+    CONTAINER_IP=$(get_container_ip ${name})
+    echo "  Testing mongod is running"
+    test_connection ${name}
+    echo "  Testing config file works"
+    docker exec $(get_cid ${name}) bash -c "test -S /var/lib/mongodb/mongodb-27017.sock"
+
+    # need to remove volume_dir with sudo because of permissions of files written
+    # by the Docker container
+    sudo rm -rf ${volume_dir}
+
+    echo "  Success!"
+}
+
 # Tests.
 run_configuration_tests
 USER="user1" PASS="pass1" DB="test_db" ADMIN_PASS="r00t" run_tests admin
 # Test with random uid in container
 CONTAINER_ARGS="-u 12345" USER="user1" PASS="pass1" DB="test_db" ADMIN_PASS="r00t" run_tests admin_altuid
 run_change_password_test
+run_mount_config_test
--- a/2.6/test/run
+++ b/2.6/test/run
@@ -280,9 +280,57 @@ function run_change_password_test() {
     echo "  Success!"
 }
 
+function run_mount_config_test() {
+    local name="mount_config"
+    echo "  Testing config file mount"
+    local database='db'
+    local user='user'
+    local password='password'
+    local admin_password='adminPassword'
+
+    local volume_dir
+    volume_dir=`mktemp -d --tmpdir mongodb-testdata.XXXXX`
+    chmod a+rwx ${volume_dir}
+    config_file=$volume_dir/mongod.conf
+    echo "dbpath=/var/lib/mongodb/dbpath
+unixSocketPrefix = /var/lib/mongodb
+smallfiles = true
+noprealloc = true" > $config_file
+    chmod a+r ${config_file}
+
+    DOCKER_ARGS="
+-e MONGODB_DATABASE=${database}
+-e MONGODB_USER=${user}
+-e MONGODB_PASSWORD=${password}
+-e MONGODB_ADMIN_PASSWORD=${admin_password}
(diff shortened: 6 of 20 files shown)


2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -253,11 +261,17 @@ function mongo_reset_passwords() {
 
 # setup_keyfile fixes the bug in mounting the Kubernetes 'Secret' volume that
 # mounts the secret files with 'too open' permissions.
+# add --keyFile argument to mongo_common_args
 function setup_keyfile() {
+  # If user specify keyFile in cofig file do not use generated keyFile
+  if [ grep -q "keyFile" ${MONGODB_CONFIG_PATH} ]; then
```
php-coder: BTW, you could remove `[` here: `if grep -q "keyFile" ${MONGODB_CONFIG_PATH} ; then` should work the same.

2.6/test/run
```diff
@@ -280,9 +280,53 @@ function run_change_password_test() {
...
+    local password='password'
+    local admin_password='adminPassword'
+
+    local volume_dir
+    volume_dir=`mktemp -d --tmpdir mongodb-testdata.XXXXX`
+    config_file=$volume_dir/mongod.conf
+    echo """dbpath=/var/lib/mongodb/dbpath
+    unixSocketPrefix = /var/lib/mongodb
+    smallfiles = true
+    nopreallocj = true""" > $config_file
```
php-coder: `nopreallocj` <-- is it a typo? (`j` at the end)

omron93: Yes. Fixed.

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: Thanks.

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -4,15 +4,23 @@ set -o errexit
...
 export CONTAINER_PORT=27017
 
+# Configuration settings.
+export MONGODB_NOPREALLOC=${MONGODB_NOPREALLOC:-true}
+export MONGODB_SMALLFILES=${MONGODB_SMALLFILES:-true}
+export MONGODB_QUIET=${MONGODB_QUIET:-true}
+
+# Constants used for waiting
+export MAX_ATTEMPTS=60
+export SLEEP_TIME=1
```
rhcarvalho: Why export these variables?

omron93: You are right. Exports are needed only for envsub, right?

So removed other unnecessary exports.

rhcarvalho: Yeah, MAX_ATTEMPTS and SLEEP_TIME are globals only used in this file.

omron93: Also MONGODB_CONFIG_PATH and MONGODB_KEYFILE_PATH, or not?

/usr/share/container-scripts/mongodb/common.sh is sourced in every run or execed file and config file does not contain them... so it is necessary to export them.

rhcarvalho: > Also MONGODB_CONFIG_PATH and MONGODB_KEYFILE_PATH, or not?

Here's where Bash haunts our souls again... :cry: 
I'm not sure if it is necessary to export in that case, I guess not necessary, need to try it out.

Here are the occurrences of MONGODB_CONFIG_PATH:

``` console
$ git grep -n MONGODB_CONFIG_PATH upstream/master -- 2.4
upstream/master:2.4/root/usr/bin/run-mongod:67:envsubst < ${CONTAINER_SCRIPTS_PATH}/mongodb.conf.template > $MONGODB_CONFIG_PATH
upstream/master:2.4/root/usr/bin/run-mongod:71:mongo_common_args="-f $MONGODB_CONFIG_PATH --oplogSize 64"
upstream/master:2.4/root/usr/share/container-scripts/mongodb/common.sh:11:export MONGODB_CONFIG_PATH=/etc/mongod.conf
upstream/master:2.4/root/usr/share/container-scripts/mongodb/initiate_replica.sh:72:mongod -f ${MONGODB_CONFIG_PATH} --shutdown &>/dev/null
```

Trying it out:

``` console
$ cat foo
#!/bin/bash

set -o errexit
set -o nounset
set -o pipefail

source bar

echo $V_EXPORTED
echo $V_NOT_EXPORTED
$ cat bar
#!/bin/bash

export V_EXPORTED="exported var"
V_NOT_EXPORTED="not exported var"
$ ./foo 
exported var
not exported var
```

Conclusion: no need to export MONGODB_CONFIG_PATH and MONGODB_KEYFILE_PATH, @omron93 agree?

BTW, MONGODB_KEYFILE_PATH is defined/exported both in `2.4/root/usr/bin/run-mongod:21` and `2.4/root/usr/share/container-scripts/mongodb/common.sh:13`, we should likely get rid of one of those lines. Notice that $HOME == /var/lib/mongodb.

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -8,8 +8,8 @@ set -o pipefail
 # because mongodb.conf lives in /var/lib/mongodb/ and we don't want a volume to
 # override it.
 export MONGODB_DATADIR=/var/lib/mongodb/data
-export MONGODB_CONFIG_PATH=/etc/mongod.conf
-export MONGODB_KEYFILE_PATH="${HOME}/keyfile"
+MONGODB_CONFIG_PATH=/etc/mongod.conf
+MONGODB_KEYFILE_PATH="${HOME}/keyfile"
```
rhcarvalho: How about organizing it such that exported and non-exported variables are separated?
I think it would be less puzzling for readers.

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -18,8 +18,8 @@ export MONGODB_SMALLFILES=${MONGODB_SMALLFILES:-true}
 export MONGODB_QUIET=${MONGODB_QUIET:-true}
 
 # Constants used for waiting
-export MAX_ATTEMPTS=60
-export SLEEP_TIME=1
+MAX_ATTEMPTS=60
+SLEEP_TIME=1
```
rhcarvalho: We may want to make these `readonly`, if they're really constants not intended to be tweaked.
Or if we want to make it possible to customize them, maybe something like `MAX_ATTEMPTS={MAX_ATTEMPTS:-60}`, WDYT?

Since we never tweaked these values (not possible with current implementation), it'd vote for `readonly`.

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: > Conclusion: no need to export MONGODB_CONFIG_PATH and MONGODB_KEYFILE_PATH, 

Agree.

> BTW, MONGODB_KEYFILE_PATH is defined/exported both in 2.4/root/usr/bin/run-mongod:21 and 2.4/root/usr/share/container-scripts/mongodb/common.sh:13, we should likely get rid of one of those lines. Notice that $HOME == /var/lib/mongodb.

This is not true in this PR.

2.4/root/usr/bin/run-mongod
```diff
@@ -6,20 +6,6 @@ set -o pipefail
...
-# override it.
-export MONGODB_DATADIR=/var/lib/mongodb/data
-
-# Configuration settings.
-export MONGODB_NOPREALLOC=${MONGODB_NOPREALLOC:-true}
-export MONGODB_SMALLFILES=${MONGODB_SMALLFILES:-true}
-export MONGODB_QUIET=${MONGODB_QUIET:-true}
-
-
-export MONGODB_KEYFILE_SOURCE_PATH="/var/run/secrets/mongo/keyfile"
```
rhcarvalho, line 20: I can't find where else this is defined?

2.4/root/usr/bin/run-mongod
```diff
@@ -6,20 +6,6 @@ set -o pipefail
...
-export MONGODB_KEYFILE_PATH="${HOME}/keyfile"
```
rhcarvalho, line 21: I see this was defined both in `common.sh` and `run-mongod`, removal looks good.

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: This sounds good.

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: Agree. Added readonly.

2.4/root/usr/bin/run-mongod
omron93, line 20: Nowhere. I can't fine where it is used?

So I removed it.

rhcarvalho, line 20: Seems right to remove it. It was originally added together with the replication example in https://github.com/openshift/mongodb/pull/17, but later its usage was removed in https://github.com/openshift/mongodb/pull/45 and the export was left over. Thanks for spotting it!

omron93:

2.4/root/usr/bin/run-mongod
```diff
@@ -46,8 +32,8 @@ function cleanup() {
     mongo_remove
   fi
   echo "=> Shutting down MongoDB server ..."
-  if [ -f "${MONGODB_PID_FILE}" ]; then
-    kill -2 $(cat ${MONGODB_PID_FILE})
+  if [[ -s ${MONGODB_DATADIR}/mongod.lock ]]; then
+    kill -2 $(cat ${MONGODB_DATADIR}/mongod.lock)
```
rhcarvalho: How is this change related to user ability to provide a config file?

2.4/root/usr/bin/run-mongod
```diff
@@ -63,12 +49,15 @@ if [ $# -gt 0 ] && [ "$1" == "initiate" ]; then
...
+# If user provides own config file use it and do not generate new one
+if [ ! -s $MONGODB_CONFIG_PATH ]; then
+  # Generate config file for MongoDB
+  envsubst < ${CONTAINER_SCRIPTS_PATH}/mongodb.conf.template > $MONGODB_CONFIG_PATH
+fi
 
 # Need to cache the container address for the cleanup
 cache_container_addr
-mongo_common_args="-f $MONGODB_CONFIG_PATH --oplogSize 64"
+mongo_common_args="-f $MONGODB_CONFIG_PATH --dbpath $MONGODB_DATADIR"
```
rhcarvalho: Why now we need to pass a custom dbpath?

2.4/test/run
```diff
@@ -280,9 +280,53 @@ function run_change_password_test() {
...
+    echo "  Testing config file mount"
+    local database='db'
+    local user='user'
+    local password='password'
+    local admin_password='adminPassword'
+
+    local volume_dir
+    volume_dir=`mktemp -d --tmpdir mongodb-testdata.XXXXX`
+    config_file=$volume_dir/mongod.conf
+    echo """dbpath=/var/lib/mongodb/dbpath
```
rhcarvalho: Why the triple double quotes here?!

Looks like Python, but this is not Python :P

2.4/root/usr/bin/run-mongod
```diff
@@ -114,8 +103,7 @@ else
   # `cleanup` references it.
   (
     unset_env_vars
-    mongod $mongo_common_args --replSet ${MONGODB_REPLICA_NAME} \
-      --keyFile ${MONGODB_KEYFILE_PATH} ${auth_args}
+    mongod $mongo_common_args --replSet ${MONGODB_REPLICA_NAME} ${auth_args}
```
rhcarvalho: I see this was already lacking proper quoting... but maybe it's a good time to add it?

`MONGODB_REPLICA_NAME` should surely be double-quoted, while `mongo_common_args` and `auth_args` we may want them to expand to multiple things... seems weird to have 2 vars that expand to arbitrary things.
Even then, maybe we can quote and treat them as arrays with the magical @ syntax...

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -253,11 +261,17 @@ function mongo_reset_passwords() {
...
 # mounts the secret files with 'too open' permissions.
+# add --keyFile argument to mongo_common_args
 function setup_keyfile() {
+  # If user specify keyFile in cofig file do not use generated keyFile
+  if grep -q "keyFile" ${MONGODB_CONFIG_PATH}; then
+    exit 0
+  fi
   if [ -z "${MONGODB_KEYFILE_VALUE-}" ]; then
-    echo "ERROR: You have to provide the 'keyfile' value in $MONGODB_KEYFILE_VALUE"
+    echo "ERROR: You have to provide the 'keyfile' value in MONGODB_KEYFILE_VALUE"
```
rhcarvalho, line 270: Don't know why this is still showing up here, but it is already in master... could you please rebase your PR (and probably squash the commits)?

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -253,11 +261,17 @@ function mongo_reset_passwords() {
 
 # setup_keyfile fixes the bug in mounting the Kubernetes 'Secret' volume that
 # mounts the secret files with 'too open' permissions.
+# add --keyFile argument to mongo_common_args
 function setup_keyfile() {
+  # If user specify keyFile in cofig file do not use generated keyFile
```
rhcarvalho: s/cofig/config

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -253,11 +261,17 @@ function mongo_reset_passwords() {
...
+  if grep -q "keyFile" ${MONGODB_CONFIG_PATH}; then
```
rhcarvalho: this is a very broad grep... I'd prefer something more specific to avoid matching the string "keyFile" in some arbitrary place of the config, like a value or comment, not a key.

2.4/root/usr/share/container-scripts/mongodb/mongodb.conf.template
```diff
@@ -1,7 +1,6 @@
 # mongodb.conf
 
-port = 27017
-pidfilepath = ${MONGODB_PID_FILE}
```
rhcarvalho, line 4: Why are we removing MONGODB_PID_FILE? If we are to remove it, we should update the README / docs too (there might be OpenShift docs referring to it).

2.4/test/run
omron93: You are right. Fixed.

2.4/root/usr/bin/run-mongod
omron93: If the image provides ability to mount config file scripts should avoid getting values from it. (I thought that OpenShift suggested not to do this to avoid bugs for some user weird config files)

MONGODB_PID_FILE is one example of it. The same value is stored in file `mongod.lock` in dbpath. So no reason to get value of MONGODB_PID_FILE from config file.

2.4/root/usr/bin/run-mongod
omron93: To avoid getting values from config file - https://github.com/openshift/mongodb/pull/147#discussion_r61854214

`dbpath` is one example and it is really necessary for MongoDB scripts. Also for better performance database should be stored in the docker volume. And user mounting config files from "classical" deployments don't have to be aware of it.

So explicit `--dbpath` overrides value from config file.

2.4/root/usr/share/container-scripts/mongodb/mongodb.conf.template
omron93, line 4: Same as in https://github.com/openshift/mongodb/pull/147#discussion_r61854214 MongoDB scripts won't use this variable. 

@rhcarvalho Who can use this? Is it somehow possible use this file with PID in other dependent applications in different containers?

What docs please? I can't find any...

2.4/root/usr/bin/run-mongod
omron93: I agree with quoting  `MONGODB_REPLICA_NAME`. I don't know why `auth_args` are separated (Also this is not necessary for MongoDB 2.6+ because `--keyFile` implies auth). So `--auth` could be simply added to `mongo_common_args` for MongoDB 2.4.

@rhcarvalho Are you fine with doing it in this PR?

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: Now matching only lines starting with `keyFile`. Should be valid for both MongoDB config file formats.

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: Fixed.

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93, line 270: Rebased.

2.4/root/usr/bin/run-mongod
rhcarvalho: Hmm, wasn't it the whole point to allow users to pass in whatever config they want? Weird to me to overwrite the dbpath.

Our default config should point to the correct dir, and perhaps we document this concern that you mention for users providing their own custom config file?

2.4/root/usr/bin/run-mongod
rhcarvalho: Maybe there should be some thought about backwards-compat before moving on with this change?

Is MONGODB_PID_FILE being read from the config file?!

omron93: What "backward-compat" do you mean?

`MONGODB_PID_FILE` before this PR is specified in `/usr/share/container-scripts/mongodb/common.sh` and it is used to stop container. In this PR `MONGODB_PID_FILE` is not used in scripts, so no need have it present in scripts.
However if you know some usecase when MONGODB_PID_FILE should be created with default config file. It could be easily added back to config.tepmlate.

2.4/root/usr/bin/run-mongod
omron93: Yes, it was. And I agree with you.

I would be happier if our scripts would not rely on this. I will try to remove this dependency.

2.4/root/usr/bin/run-mongod
omron93: @rhcarvalho Do you know why `cleanup()` uses `kill` or `pkill` instead of `mongod --shutdown`?

2.4/root/usr/share/container-scripts/mongodb/mongodb.conf.template
rhcarvalho, line 4: If it's just used internally then I see no problems with removing it.

2.4/root/usr/bin/run-mongod
rhcarvalho: > Do you know why cleanup() uses kill or pkill instead of mongod --shutdown?

I don't know, probably using `--shutdown` would be preferred. But again, it's easier to review PRs that do one thing.
That change would be much easier to review in a PR of itself IMHO.

omron93: > it's easier to review PRs that do one thing.
> That change would be much easier to review in a PR of itself IMHO.

Yes. But it depends on graininess. If "every line change" is submited in separate PR, it really makes improvements harder and makes thinks slower.

However, there is a new PR #155

2.4/root/usr/bin/run-mongod
omron93: New PR for this #156

2.4/root/usr/bin/run-mongod
omron93: Necessity to override user dbpath was removed in #157. So removed from this PR.

omron93:
Rebased and also added 3.2 version.

So now:

>    Added ability to use user config file.
> - user is able to create or mount custom config file in /etc/mongod.conf (for example from non-docker MongoDB deployments)
> - this config file does not affect MONGODB_REPLICA_NAME (possibility to use user config file is not aimed at replicated environment)

@hhorak @bparees @rhcarvalho If there are no other suggestions please merge.

bparees:
lgtm, but please update the READMEs to describe the new supported behavior for supplying a config file:
https://github.com/sclorg/mongodb-container/blob/master/2.4/README.md
https://github.com/sclorg/mongodb-container/blob/master/2.6/README.md
https://github.com/sclorg/mongodb-container/blob/master/3.2/README.md

bparees:
[test]

omron93:
@bparees README updated.

Test failed, but it seems that it is not a problem in this PR.

bparees:
i'm not sure the failure isn't the fault of this PR:

```
+ docker run --rm openshift/mongodb-24-rhel7-candidate mongo db --host 172.17.0.2 -u user -p password --eval 'db.getSiblingDB('\''test_database'\'');'
exception: connect failed
+ status=1
+ set -e
+ '[' 1 -eq 0 ']'
+ sleep 2
+ echo '  Giving up: Failed to connect. Logs:'
```

seems like it's not able to connect.  are you able to run `make test` on this PR locally successfully?

omron93:
It is fix in this PR, I am sorry.

Should be fixed. Please retest.

bparees:
[test]

2.4/README.md
```diff
@@ -58,6 +58,14 @@ will be started. If you are re-attaching the volume to another container, the
 creation of the database user and admin user will be skipped and only the
 MongoDB daemon will be started.
 
+Custom configuration file
+---------------------------------
+
+It is allowed to use custom configuration file for mongod server.
+
+To use custom configuration file in container it has to be mounted into `/etc/mongod.conf`. For example to use configuration file stored in `/home/user` directory use this option for `docker run` command: `-v /home/user/mongod.conf:/etc/mongod.conf:Z`.
```
bparees, line 66: "Providing a custom configuration file supercedes the individual configuration environment variable values."

bparees:
one readme nit and lgtm, @hhorak ?

2.4/README.md
omron93: Look good, thanks. Added.

hhorak:
@omron93, after 30-upg was merged, could you do the same changes for that version as well, pls?
Otherwise looks good to me, so LGTM.

omron93:
Rebased and added 30-upg.

@bparees @hhorak Please merge.

PR #151, opened 2016-04-25
Fix missing argument to _wait_for_mongo

rhcarvalho:
Host was always 'localhost', and the documented argument '$2' was never set.

Use '$@' instead of '$1' because 'wait_for_mongo_up' and 'wait_for_mongo_down' can be (and are) called without arguments. Using '${1:-}' would be more brittle.

Introduced in https://github.com/openshift/mongodb/pull/135.


rhcarvalho:
@bparees PTAL
@omron93 @php-coder FYI

bparees:
looks like a repo failure.  [test]

rhcarvalho:
[test] again

bparees:
lgtm

rhcarvalho:
And of course I missed the 2.6 again... :(
Will follow up.

PR #152, opened 2016-04-28
Fix missing argument to _wait_for_mongo (2.6)

rhcarvalho:
Host was always 'localhost', and the documented argument '$2' was never
set.

Use '$@' instead of '$1' because 'wait_for_mongo_up' and
'wait_for_mongo_down' can be (and are) called without arguments.
Using '${1:-}' would be more brittle.

A previous commit added the same change set to 2.4.

Follow up on #151.


rhcarvalho:
@bparees PTAL

bparees:
lgtm, will wait on test success.

PR #153, opened 2016-04-29
Fix inconsitencies between 2.4 and 2.6

rhcarvalho:
Some commits added/removed things from 2.4 and changes were not applied
to 2.6.


rhcarvalho:
@bparees PTAL

bparees:
lgtm, thanks @rhcarvalho

bparees:
(looks safe enough but i'll wait for tests)

PR #154, opened 2016-05-05
2.4: fix user creation in replica set configuration

php-coder:
Extended tests for MongoDB replication are failing (https://github.com/openshift/origin/issues/8753) because of changes made in #136 (we updated code for 2.6 but forgot to update it for 2.4).

PTAL @bparees @mfojtik 
CC @omron93


mfojtik:
LGTM

PR #155, opened 2016-05-06
Use mongod to stop server instead of kill or pkill.

omron93:
This method of stopping `mongod` is equivalent to using `kill -2` (https://docs.mongodb.com/manual/tutorial/manage-mongodb-processes/#use-shutdown).

This change is a part of removing dependencies of scripts on mongodb options - it necessary to allow using custom config file in #147

### DIFF
--- a/2.4/root/usr/bin/run-mongod
+++ b/2.4/root/usr/bin/run-mongod
@@ -46,9 +46,7 @@ function cleanup() {
     mongo_remove
   fi
   echo "=> Shutting down MongoDB server ..."
-  if [ -f "${MONGODB_PID_FILE}" ]; then
-    kill -2 $(cat ${MONGODB_PID_FILE})
-  else
+  if pgrep mongod; then
     pkill -2 mongod
   fi
   wait_for_mongo_down
--- a/2.6/root/usr/bin/run-mongod
+++ b/2.6/root/usr/bin/run-mongod
@@ -46,9 +46,7 @@ function cleanup() {
     mongo_remove
   fi
   echo "=> Shutting down MongoDB server ..."
-  if [ -f "${MONGODB_PID_FILE}" ]; then
-    kill -2 $(cat ${MONGODB_PID_FILE})
-  else
+  if pgrep mongod; then
     pkill -2 mongod
   fi
   wait_for_mongo_down


rhcarvalho:
LGTM

@bparees @mfojtik could any of you merge please?

mfojtik:
[test]

2.4/root/usr/bin/run-mongod
```diff
@@ -47,9 +47,7 @@ function cleanup() {
   fi
   echo "=> Shutting down MongoDB server ..."
   if [ -f "${MONGODB_PID_FILE}" ]; then
-    kill -2 $(cat ${MONGODB_PID_FILE})
-  else
-    pkill -2 mongod
+    mongod --shutdown $mongo_common_args
```
bparees: what are we doing if there's no mongo pid file now?

omron93: `mongod` handles this itself. It has path to the database where the PID is also stored in `mongod.lock`

bparees: yeah but the if statement is not going to get executed if the pid file does not exist.  previously this if had a fallback for that condition.

omron93: Also in #147 I would like to don't use  MONGODB_PID_FILE and use `pgrep` instead of it...

bparees: so should line 49 just be removed entirely?

omron93: No, if `mongod` is not running this call exits with error code 1 and error message `There doesn't seem to be a server running with dbpath: ...`

It very similar to kill or pkill...

bparees: hm.  i'm just worried about the case where the pid file got removed for some reason, but mongo is actually still running.  Now we wont kill it.

omron93: What could remove the pid file?

Should our scripts also try to fix this? If user removes it in mounted storage, it is a mistake of user...

Or in every case use also `pkill`?

bparees: not sure, but whoever wrote the original logic thought it was a condition worth worrying about.

omron93: And does some know who and why wrote it in this way?

bparees: i'm gonna bet on @rhcarvalho or @php-coder

omron93: In other parts of scripts, there is also only `mongod --shutdown`...

https://github.com/openshift/mongodb/blob/master/2.6/root/usr/bin/run-mongod#L94
https://github.com/openshift/mongodb/blob/master/2.6/root/usr/bin/run-mongod#L86

bparees: so let's just always call it then, and do an || true to avoid an error resulting in us exiting.  or do a set +e/set -e

omron93: https://github.com/openshift/mongodb/commit/d7efd424116c7fe087094f8887940eb8e96ac1c6

@mfojtik Do you remember why you've added this form of `cleanup()`

bparees: @mfojtik bump

mfojtik: @omron93 @bparees it might be something related to replica, but I don't remember the details now, sorry :-/

bparees: I reiterate my suggestion:

> so let's just always call it then, and do an || true to avoid an error resulting in us exiting. or do a set +e/set -e

omron93: IMPOV it would be better to use only pgrep a pkill instead of using || true or set +-e (it also removes dependency on PID file and it is more clear). What do you think about current implementation?

omron93: IMPOV it would be better to use only pgrep a pkill instead of using || true or set +-e (it also removes dependency on PID file and it is more clear). What do you think about current implementation?

bparees: i don't care how you do it as long as you always kill the process regardless of the existence of the pid file, and you don't error out.

omron93:
@rhcarvalho @bparees @mfojtik Rebased. Any other suggestions?

bparees:
[test]

bparees:
lgtm, will merge if it passes tests.

PR #156, opened 2016-05-06
Clean passing arguments to mongod in replSet

omron93:
Clean handling authentication in replSet
- `--auth` could be added to `mongo_common_args`. There is no reason why to have it in separate variable. Also in MongoDB 2.6+ `--auth` is redundant with `--keyFile` (https://docs.mongodb.com/v2.6/reference/configuration-options/#security.keyFile) so removing it for 2.6 to have script more clean.

Double-quote mongod options - based on this comment https://github.com/openshift/mongodb/pull/147#discussion_r61588528

### DIFF
--- a/2.4/root/usr/bin/run-mongod
+++ b/2.4/root/usr/bin/run-mongod
@@ -103,19 +103,18 @@ else
     run_mongod_supervisor
     trap 'cleanup' SIGINT SIGTERM
   fi
-  auth_args=" --auth"
-  # When initializing the MongoDB cluster, we have to run without
-  # authentication
-  if [ ! -v MONGODB_NO_AUTH ]; then
-    auth_args=""
+  # In MongoDB 2.4 --keyFile does not imply authentication, so enabling it explicitly
+  # When not initializing the MongoDB cluster, we have to run with authentication
+  if [ -v MONGODB_NO_AUTH ]; then
+    mongo_common_args+=" --auth"
   fi
   # Run `unset_env_vars` and `mongod` in a subshell because
   # MONGODB_ADMIN_PASSWORD should still be defined when the trapped call to
   # `cleanup` references it.
   (
     unset_env_vars
-    mongod $mongo_common_args --replSet ${MONGODB_REPLICA_NAME} \
-      --keyFile ${MONGODB_KEYFILE_PATH} ${auth_args}
+    mongod $mongo_common_args --replSet "${MONGODB_REPLICA_NAME}" \
+      --keyFile "${MONGODB_KEYFILE_PATH}"
   ) &
   wait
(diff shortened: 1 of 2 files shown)


rhcarvalho:
@omron93 why is the patch for 2.4 different than 2.6?

omron93:
@rhcarvalho It is because `mongod` also uses `--keyFile` and for MongoDB 2.6 and later this options implies `--auth`. So `--auth` is redundant 2.6 (for 3.0 and 3.2 as well).

So I removed it to make code more clean.

omron93:
@rhcarvalho It is because `mongod` also uses `--keyFile` and for MongoDB 2.6 and later this options implies `--auth`. So `--auth` is redundant 2.6 (for 3.0 and 3.2 as well).

So I removed it to make code more clean.

php-coder:
I would like to have a comment about this differences between 2.4/2.6 somewhere in the code, but I don't know where to put it. If later we will decide to merge common scripts together it will be more difficult to understand where and especially why scripts differ. Do anybody see a way of improving this?

(Just my thoughts that shouldn't affect merging process anyway.)

omron93:
It is a good idea. There is more differences, so I would not add this as a part of this PR...

It would great to have some file in repository and in to have some comments what differ among versions. Where to store it? Or to have it like a part of README?

php-coder:
Documentation becomes outdated very quickly, especially when it's separated from the code (comments in the file vs separate file).

Maybe explicitly check for the mongodb version in such cases? For example:

``` sh
# When initializing the MongoDB cluster, we have to run without authentication.
# For MongoDB 2.6 and later we don't need it because --keyFile implies --auth.
if [ -v MONGODB_NO_AUTH -a ${MONGODB_VERSION:-2.4} = '2.4' ]; then
  mongo_common_args+=' --auth'
fi
```

WDYT?

omron93:
It works also for 2.6+ to add `--auth` with `--keyFile`. I only wanted to remove redundant piece of code in 2.6 and future versions. In this is is even more complicated :-(

What about adding commend in 2.4 code
`# In MongoDB 2.4 --keyFile does not imply authentication, so enabling it explicitly` ?

php-coder:
> What about adding commend in 2.4 code
> # In MongoDB 2.4 --keyFile does not imply authentication, so enabling it explicitly ?

It definitely better than what we have right now in the PR.

omron93:
Added...

php-coder:
Thank you!

omron93:

php-coder:
No objection.

rhcarvalho:
LGTM @bparees PTAL

bparees:
lgtm

PR #157, opened 2016-05-06
Do not rely on hidden file in dbpath

omron93:
In current situation scripts check if database path contains file `.mongodb_datadir_initialized`. It allows user to mount database directory inside a container.
However if user mounts dbpath not created by openshift images this file does not exists!

This PR allows scripts to check if a user is present without relying on this hidden file. So it is possible to mount database from non container deployment.
This change is also part of removing dependencies of scripts on mongodb options - it necessary to allow using custom config file in #147

### DIFF
--- a/2.4/root/usr/bin/run-mongod
+++ b/2.4/root/usr/bin/run-mongod
@@ -72,26 +72,25 @@ if [ -z "${MONGODB_REPLICA_NAME-}" ]; then
     usage
   fi
   # Run the MongoDB in 'standalone' mode
-  if [ ! -f ${MONGODB_DATADIR}/.mongodb_datadir_initialized ]; then
-    # Create MongoDB users and restart MongoDB with authentication enabled
-    # At this time the MongoDB does not accept the incoming connections.
-    mongod $mongo_common_args & #--bind_ip 127.0.0.1 --quiet >/dev/null &
-    wait_for_mongo_up
-    mongo_create_admin
-    mongo_create_user
-    touch ${MONGODB_DATADIR}/.mongodb_datadir_initialized
-    # Restart the MongoDB daemon to bind on all interfaces
-    mongod $mongo_common_args --shutdown
-    wait_for_mongo_down
+  mongod $mongo_common_args &
+  wait_for_mongo_up
+  js_command="db.system.users.count({'user':'admin'})"
+  if [ "$(mongo admin --quiet --eval "$js_command")" == "1" ]; then
+    echo "=> Admin user is already created. Resetting password ..."
+    mongo_reset_admin
   else
-    echo "=> Database directory is already initialized. Skipping creation of users ..."
-    # Ensure passwords match environment variables
-    mongod $mongo_common_args &
-    wait_for_mongo_up
-    mongo_reset_passwords
-    mongod $mongo_common_args --shutdown
-    wait_for_mongo_down
+    mongo_create_admin
   fi
+  js_command="db.system.users.count({'user':'${MONGODB_USER}'})"
+  if [ "$(mongo ${MONGODB_DATABASE} --quiet --eval "$js_command")" == "1" ]; then
+    echo "=> MONGODB_USER user is already created. Resetting password ..."
+    mongo_reset_user
+  else
+    mongo_create_user
+  fi
+  # Restart the MongoDB daemon to bind on all interfaces
+  mongod $mongo_common_args --shutdown
+  wait_for_mongo_down
   unset_env_vars
   exec mongod $mongo_common_args --auth
 else
--- a/2.4/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.4/root/usr/share/container-scripts/mongodb/common.sh
@@ -227,10 +227,8 @@ function mongo_create_user() {
   fi
 }
 
-# mongo_reset_passwords sets the MongoDB passwords to match MONGODB_PASSWORD
-# and MONGODB_ADMIN_PASSWORD
-function mongo_reset_passwords() {
-  # Reset password of MONGODB_USER
+# mongo_reset_user sets the MongoDB MONGODB_USER's password to match MONGODB_PASSWORD
+function mongo_reset_user() {
   if [[ -n "${MONGODB_USER:-}" && -n "${MONGODB_PASSWORD:-}" && -n "${MONGODB_DATABASE:-}" ]]; then
     local js_command="db.changeUserPassword('${MONGODB_USER}', '${MONGODB_PASSWORD}')"
     if ! mongo ${MONGODB_DATABASE} --eval "${js_command}"; then
@@ -238,8 +236,10 @@ function mongo_reset_passwords() {
       exit 1
     fi
   fi
+}
 
-  # Reset password of admin
+# mongo_reset_admin sets the MongoDB admin password to match MONGODB_ADMIN_PASSWORD
+function mongo_reset_admin() {
   if [[ -n "${MONGODB_ADMIN_PASSWORD:-}" ]]; then
     local js_command="db.changeUserPassword('admin', '${MONGODB_ADMIN_PASSWORD}')"
     if ! mongo admin --eval "${js_command}"; then
--- a/2.6/root/usr/bin/run-mongod
+++ b/2.6/root/usr/bin/run-mongod
@@ -72,26 +72,25 @@ if [ -z "${MONGODB_REPLICA_NAME-}" ]; then
     usage
   fi
   # Run the MongoDB in 'standalone' mode
-  if [ ! -f ${MONGODB_DATADIR}/.mongodb_datadir_initialized ]; then
-    # Create MongoDB users and restart MongoDB with authentication enabled
-    # At this time the MongoDB does not accept the incoming connections.
-    mongod $mongo_common_args & #--bind_ip 127.0.0.1 --quiet >/dev/null &
-    wait_for_mongo_up
-    mongo_create_admin
-    mongo_create_user
-    touch ${MONGODB_DATADIR}/.mongodb_datadir_initialized
-    # Restart the MongoDB daemon to bind on all interfaces
-    mongod $mongo_common_args --shutdown
-    wait_for_mongo_down
+  mongod $mongo_common_args &
+  wait_for_mongo_up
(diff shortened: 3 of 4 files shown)


mfojtik:
[test]

2.4/root/usr/bin/run-mongod
```diff
@@ -74,26 +74,23 @@ if [ -z "${MONGODB_REPLICA_NAME-}" ]; then
...
   fi
   # Run the MongoDB in 'standalone' mode
-  if [ ! -f ${MONGODB_DATADIR}/.mongodb_datadir_initialized ]; then
-    # Create MongoDB users and restart MongoDB with authentication enabled
-    # At this time the MongoDB does not accept the incoming connections.
-    mongod $mongo_common_args & #--bind_ip 127.0.0.1 --quiet >/dev/null &
-    wait_for_mongo_up
+  mongod $mongo_common_args &
+  wait_for_mongo_up
+  if [ $(mongo admin --quiet --eval "db.system.users.find({'user':'admin'}).length()") == "1" ]; then
```
php-coder: I suggest to add quotes to the command and also use `collection.count()` instead of `collection.find().length()`

omron93: What is a difference `count() vs length()`?

`collection.find()` is used to return only user with exact name (and in 2.6 with specified exact database).

php-coder: > What is a difference count() vs length()?

In case of using `collection.find().length()` you're fetching results from MongoDB and after that use `length()` for counting how many rows were returned. You're ignoring results completely and don't use it. It's the same as using `SELECT COUNT(*)` vs `SELECT *` and counting rows manually.

I'm suggesting explicitly ask mongodb to count rows for us and return result. For a huge amount of data it could be faster (but it's not our case) and also it's more clean way IMHO.

P.S. Examples:

``` console
bash-4.2$ mongo admin -u admin -ptest --quiet --eval "db.system.users.find({'user':'admin'})" 
DBQuery: admin.system.users -> { "user" : "admin" }
bash-4.2$ mongo admin -u admin -ptest --quiet --eval "db.system.users.find({'user':'admin'}).length()" 
1
bash-4.2$ mongo admin -u admin -ptest --quiet --eval "db.system.users.count({'user':'admin'})"         
1
```

omron93: @php-coder You are right, I've forgotten about this method. It is better...

Both should be fixed now.

omron93:

omron93:
@rhcarvalho @php-coder @bparees What to do to make this PR merged?

bparees:
lgtm, @rhcarvalho or @php-coder any concerns?

2.4/root/usr/bin/run-mongod
```diff
@@ -74,26 +74,25 @@ if [ -z "${MONGODB_REPLICA_NAME-}" ]; then
...
-  if [ ! -f ${MONGODB_DATADIR}/.mongodb_datadir_initialized ]; then
-    # Create MongoDB users and restart MongoDB with authentication enabled
-    # At this time the MongoDB does not accept the incoming connections.
-    mongod $mongo_common_args & #--bind_ip 127.0.0.1 --quiet >/dev/null &
-    wait_for_mongo_up
+  mongod $mongo_common_args &
+  wait_for_mongo_up
+  js_command="db.system.users.count({'user':'admin'})"
+  if [ "$(mongo admin --quiet --eval "$js_command")" == "1" ]; then
+    echo "=> Admin user is already created. Reseting password ..."
```
rhcarvalho: s/Reseting/Resetting/g

2.4/root/usr/bin/run-mongod
```diff
@@ -74,26 +74,25 @@ if [ -z "${MONGODB_REPLICA_NAME-}" ]; then
...
   # Run the MongoDB in 'standalone' mode
-  if [ ! -f ${MONGODB_DATADIR}/.mongodb_datadir_initialized ]; then
-    # Create MongoDB users and restart MongoDB with authentication enabled
-    # At this time the MongoDB does not accept the incoming connections.
-    mongod $mongo_common_args & #--bind_ip 127.0.0.1 --quiet >/dev/null &
-    wait_for_mongo_up
+  mongod $mongo_common_args &
+  wait_for_mongo_up
+  js_command="db.system.users.count({'user':'admin'})"
+  if [ "$(mongo admin --quiet --eval "$js_command")" == "1" ]; then
```
rhcarvalho, line 78: What if `mongo admin` fails because e.g. the server is down? We'll end up calling `mongo_reset_admin` and then?

2.4/root/usr/bin/run-mongod
```diff
@@ -74,26 +74,25 @@ if [ -z "${MONGODB_REPLICA_NAME-}" ]; then
...
   fi
   # Run the MongoDB in 'standalone' mode
-  if [ ! -f ${MONGODB_DATADIR}/.mongodb_datadir_initialized ]; then
-    # Create MongoDB users and restart MongoDB with authentication enabled
-    # At this time the MongoDB does not accept the incoming connections.
-    mongod $mongo_common_args & #--bind_ip 127.0.0.1 --quiet >/dev/null &
-    wait_for_mongo_up
+  mongod $mongo_common_args &
+  wait_for_mongo_up
+  js_command="db.system.users.count({'user':'admin'})"
```
rhcarvalho, line 77: Not a problem, just a note -- we're changing from looking at an existing file to look at the existence of a specific user.
Seems sane.

2.4/root/usr/bin/run-mongod
```diff
@@ -74,26 +74,25 @@ if [ -z "${MONGODB_REPLICA_NAME-}" ]; then
     usage
   fi
   # Run the MongoDB in 'standalone' mode
-  if [ ! -f ${MONGODB_DATADIR}/.mongodb_datadir_initialized ]; then
-    # Create MongoDB users and restart MongoDB with authentication enabled
-    # At this time the MongoDB does not accept the incoming connections.
-    mongod $mongo_common_args & #--bind_ip 127.0.0.1 --quiet >/dev/null &
-    wait_for_mongo_up
+  mongod $mongo_common_args &
+  wait_for_mongo_up
```
rhcarvalho, line 76: The above two lines look good as well, since the two if branches did the same thing... :+1: @omron93

2.4/root/usr/bin/run-mongod
```diff
@@ -74,26 +74,25 @@ if [ -z "${MONGODB_REPLICA_NAME-}" ]; then
...
+    mongo_reset_admin
+  else
     mongo_create_admin
-    mongo_create_user
-    touch ${MONGODB_DATADIR}/.mongodb_datadir_initialized
-    # Restart the MongoDB daemon to bind on all interfaces
-    mongod $mongo_common_args --shutdown
-    wait_for_mongo_down
+  fi
+  js_command="db.system.users.count({'user':'${MONGODB_USER}'})"
```
rhcarvalho, line 84: What if `MONGODB_USER` is changed?

I think that was not properly supported before, only password changes.
So this seems good, if the value of `MONGODB_USER` is changed, it will create a new user (and the old user is left intact). :+1:

2.4/root/usr/bin/run-mongod
omron93: Thanks. Fixed.

2.4/root/usr/bin/run-mongod
omron93, line 78: It is the same as in current state.  What if database is not initialized and server is down - resetting passwd fails. Mongod should be running and it is running before this command. If `mongo admin` fails because server is down something went wrong so scripts fails in every case!

Now scripts check that mongod is running (wait_for_mongo_up) and then it assumes that it is running and do not check it every time - in that case mongod would fail in every line of code (each command should have `|| true`). So with your worry mongodb scripts would be completely wrong. And I hope it is not needed to check mongod status in every command.

2.4/root/usr/bin/run-mongod
omron93, line 84: You've answered yourself :-)

2.4/root/usr/bin/run-mongod
omron93, line 77: Yes. And we don't bring our implementation (adding file - it could fail in some time) and instead of it we rely on MongoDB.

2.4/root/usr/bin/run-mongod
rhcarvalho, line 78: Yeah, not to check in every command, just thinking aloud to make sure we're not accidentally changing behavior.

Checking if a file exists fails for different reasons than `mongo admin --quiet --eval ...`.
Perhaps I'm just thinking this far remembering debug sessions that @php-coder had when we saw strange behavior hard to correlate with an actual cause.

We have `set -e` etc to make it fail fast, I wouldn't like to add `|| true` every other line, because IMHO it makes it harder to understand problems when things blow up much further from the original error.

2.4/root/usr/bin/run-mongod
```diff
@@ -72,26 +72,25 @@ if [ -z "${MONGODB_REPLICA_NAME-}" ]; then
...
+  if [ "$(mongo ${MONGODB_DATABASE} --quiet --eval "$js_command")" == "1" ]; then
+    echo "=> MONGODB_USER user is already created. Reseting password ..."
```
rhcarvalho: @omron93 you missed the global replacement :)

2.4/root/usr/bin/run-mongod
omron93, line 78: @rhcarvalho So what do you suggest?

2.4/root/usr/bin/run-mongod
omron93: Sorry. Should be OK now :-)

2.4/root/usr/bin/run-mongod
rhcarvalho, line 78: To leave it as is.

If `mongo admin` errors out, it's unlikely that it will output "1". In any case, no matter what branch of the  `if-else` we go, either `mongo_create_admin` or `mongo_reset_admin` will fail and `exit 1`.

bparees:
[test]

PR #158, opened 2016-05-12
Remove creationTimestamp from template

rhcarvalho:
The field shouldn't be in the template, it is set when the
DeploymentConfig is created.


rhcarvalho:
[test] @bparees PTAL

bparees:
PR tester is broken right now (i'm working on it), but this looks safe.

PR #161, opened 2016-05-12
Remove ignore fields in template

rhcarvalho:
The fields belong to a PodSpec, but they are for a very long time part
of the DeploymentConfigSpec, and simply ignored when the template is
loaded into OpenShift.
- restartPolicy: we probably want to keep the default behavior, which is
  restart 'Always'. If a pod running mongod dies, we want it to be
  restarted. The 'Never' was never really used.
- dnsPolicy: the default is 'ClusterFirst', so even if we'd move it to
  the PodSpec, it'd be a no-op.

### DIFF
--- a/2.4/examples/replica/mongodb-clustered.json
+++ b/2.4/examples/replica/mongodb-clustered.json
@@ -170,9 +170,7 @@
               }
             ]
           }
-        },
-        "restartPolicy": "Never",
-        "dnsPolicy": "ClusterFirst"
+        }
       }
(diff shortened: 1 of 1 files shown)


rhcarvalho:
@bparees PTAL

Seems that we haven't noticed the misplaced keys/values for quite a long time... :)

I'm looking into ways to avoid this type of problem in the future.

bparees:
lgtm, thanks.

PR #162, opened 2016-06-01
Fix adding arguments in mongo_cmd_local.

omron93:
Execing command into container in tests in function `mongo_cmd_local` is not working properly. If `bash -c` does not take quoted string arguments are not passed to running command.

`mongo_cmd_local` could be used to connect to database which ofter requires adding additional arguments to mongo shell.


bparees:
[test]

bparees:
lgtm

PR #165, opened 2016-06-13
fix recursive handling of links

gabemontero:
@bparees PTAL

### DIFF
--- a/2.4/root/usr/libexec/fix-permissions
+++ b/2.4/root/usr/libexec/fix-permissions
@@ -1,6 +1,6 @@
 #!/bin/sh
 # Fix permissions on the given directory to allow group read/write of 
 # regular files and execute of directories.
-chgrp -R 0 $1
-chmod -R g+rw $1
+find $1 -exec chgrp 0 {} \;
+find $1 -exec chmod g+rw {} \;
 find $1 -type d -exec chmod g+x {} +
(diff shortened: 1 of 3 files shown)


2.4/root/usr/libexec/fix-permissions
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 # Fix permissions on the given directory to allow group read/write of 
 # regular files and execute of directories.
-chgrp -R 0 $1
-chmod -R g+rw $1
+find $1 -exec chgrp -R 0 {} \;
```
php-coder: Should we omit `-R` option here like we did for `chmod`?

gabemontero: yep it is unnecessary

php-coder:
Gabe, could you explain, why `find`+`chmod` is better than `chmod -R`? Is `find`+`chmod` will do the same as `chmod -R --no-dereference`? Just interesting to know.

gabemontero:
@bparees and I found that `chmod -R` was not identifying links.  See https://github.com/openshift/jenkins/issues/101

2.4/root/usr/libexec/fix-permissions
bparees: fyi i updated the jenkins fix-permissions to remove the -R there.  thanks for catching that @php-coder

gabemontero: cool - one less to correct - thx @bparees ; and yep thx @php-coder

gabemontero:
removal of -R pushed

bparees:
lgtm

PR #171, opened 2016-08-11
Revert "add jenkins status badge"

PI-Victor:
sorry @hhorak here's the revert, since we don't need the badge.
Reverts sclorg/mongodb-container#169


PI-Victor:
@bparees can you merge this to undo the pesky tag?

PR #176, opened 2016-08-25
Use mongodb replset addressing.

omron93:
This PR changes way of addressing replica set.

In current situation scripts connects to primary member of replset to access it. For this an IP address of primary has to be found.
Upstream suggests different way of addressing. "Use the following form: `<replSetName>/<hostname1><:port>,<hostname2><:port>,<...>`" (https://docs.mongodb.com/manual/reference/program/mongo/#cmdoption--host)

This method automatically selects where to connect (for example in case of some primary failure it automatically reconnects to the rest of replset). So we don't have to develop out methods (`mongo_primary_member_addr` function) to connect replset.
Also I hope code is also shorter and more clean.

@bparees @rhcarvalho @hhorak Please take a look and test. In my replication tests everything works fine.

### DIFF
--- a/2.4/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.4/root/usr/share/container-scripts/mongodb/common.sh
@@ -122,68 +122,51 @@ function mongo_initiate() {
   mongo admin --eval "${config};rs.initiate(config);${mongo_wait}"
 }
 
-# get the address of the current primary member
-function mongo_primary_member_addr() {
-  local rc=0
-
-  endpoints | grep -v "$(container_addr)" |
-  (
-    while read mongo_node; do
-      cmd_output="$(mongo admin -u admin -p "$MONGODB_ADMIN_PASSWORD" --host "$mongo_node:$CONTAINER_PORT" --eval 'print(rs.isMaster().primary)' --quiet || true)"
-
-      # Trying to find IP:PORT in output and filter out error message because mongo prints it to stdout
-      ip_and_port_regexp='[0-9]\+\.[0-9]\+\.[0-9]\+\.[0-9]\+:[0-9]\+'
-      if addr="$(echo "$cmd_output" | grep -x "$ip_and_port_regexp")"; then
-        echo -n "$addr"
-        exit 0
-      fi
-
-      echo >&2 "Cannot get address of primary from $mongo_node node: $cmd_output"
-    done
-
-    exit 1
-  ) || rc=$?
-
-  if [ $rc -ne 0 ]; then
-    echo >&2 "Cannot get address of primary node: after checking all nodes we don't have the address"
+# replset_addr return the address of the current replSet
+function replset_addr() {
+  local current_endpoints
+  current_endpoints="$(endpoints)"
+  if [ -z "${current_endpoints}" ]; then
+    echo >&2 "Cannot get address of replica set: no nodes are listed in service"
     return 1
   fi
+  echo "${MONGODB_REPLICA_NAME}/${current_endpoints//[[:space:]]/,}"
 }
 
 # mongo_remove removes the current MongoDB from the cluster
 function mongo_remove() {
-  local primary_addr
+  local host
   # if we cannot determine the IP address of the primary, exit without an error
   # to allow callers to proceed with their logic
-  primary_addr="$(mongo_primary_member_addr || true)"
-  if [ -z "$primary_addr" ]; then
+  host="$(replset_addr || true)"
+  if [ -z "$host" ]; then
     return
   fi
 
   local mongo_addr
   mongo_addr="$(mongo_addr)"
 
-  echo "=> Removing ${mongo_addr} on ${primary_addr} ..."
+  echo "=> Removing ${mongo_addr} from replica set ..."
   mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" \
-    --host "${primary_addr}" --eval "rs.remove('${mongo_addr}');" || true
+    --host "${host}" --eval "rs.remove('${mongo_addr}');" || true
 }
 
 # mongo_add advertise the current container to other mongo replicas
 function mongo_add() {
-  local primary_addr
+  local host
   # if we cannot determine the IP address of the primary, exit without an error
   # to allow callers to proceed with their logic
-  primary_addr="$(mongo_primary_member_addr || true)"
-  if [ -z "$primary_addr" ]; then
+  host="$(replset_addr || true)"
+  if [ -z "$host" ]; then
     return
   fi
 
   local mongo_addr
   mongo_addr="$(mongo_addr)"
 
-  echo "=> Adding ${mongo_addr} to ${primary_addr} ..."
+  echo "=> Adding ${mongo_addr} to replica set ..."
   mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" \
-    --host "${primary_addr}" --eval "rs.add('${mongo_addr}');"
+    --host "${host}" --eval "rs.add('${mongo_addr}');"
 }
 
 # mongo_create_admin creates the MongoDB admin user with password: MONGODB_ADMIN_PASSWORD
--- a/2.4/root/usr/share/container-scripts/mongodb/replset_supervisor.sh
+++ b/2.4/root/usr/share/container-scripts/mongodb/replset_supervisor.sh
@@ -62,7 +62,7 @@ if [[ "$1" == "initiate" ]]; then
   mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --quiet --host ${mongo_node} --eval "var done=false;while(done==false){var members=rs.status().members;for(i=0;i<members.length;i++){if(members[i].stateStr=='PRIMARY' && members[i].name!='$(mongo_addr)'){done=true}};sleep(500)};" &>/dev/null
 
   # Remove the initialization container MongoDB from cluster and shutdown
-  echo "=> The new PRIMARY member elected, shutting down current member ..."
+  echo "=> A new PRIMARY member was elected, shutting down local mongod ..."
   mongo_remove
 
   mongod -f ${MONGODB_CONFIG_PATH} --shutdown &>/dev/null
--- a/2.6/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/common.sh
@@ -121,68 +121,51 @@ function mongo_initiate() {
   mongo admin --eval "${config};rs.initiate(config);${mongo_wait}"
 }
 
-# get the address of the current primary member
-function mongo_primary_member_addr() {
-  local rc=0
-
-  endpoints | grep -v "$(container_addr)" |
-  (
-    while read mongo_node; do
-      cmd_output="$(mongo admin -u admin -p "$MONGODB_ADMIN_PASSWORD" --host "$mongo_node:$CONTAINER_PORT" --eval 'print(rs.isMaster().primary)' --quiet || true)"
-
-      # Trying to find IP:PORT in output and filter out error message because mongo prints it to stdout
-      ip_and_port_regexp='[0-9]\+\.[0-9]\+\.[0-9]\+\.[0-9]\+:[0-9]\+'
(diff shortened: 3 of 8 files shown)


2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -113,40 +113,22 @@ function mongo_initiate() {
...
-    done
-
-    exit 1
-  ) || rc=$?
-
-  if [ $rc -ne 0 ]; then
-    echo >&2 "Cannot get address of primary node: after checking all nodes we don't have the address"
+# replset_addr return the address of the current replSet
+function replset_addr() {
+  current_endpoints=$(endpoints)
```
php-coder: Do we have a reason of not defining this variable as local and not using double quotes?

I'd rather write it as:

``` bash
  local current_endpoints
  current_endpoints="$(endpoints)"
```

php-coder:
Let's say we have 3 members: node1, node2, node3 and node2 is a current primary. Currently the code finds out that node2 is primary, connects to it and executes command like "add node4 to the replica set". Will it continue to work if code connects to `rs0/node1:27017,node2:27017,node3:27017`? In this case it will connect to node2 or node1?

omron93:
> @php-coder  Let's say we have 3 members: node1, node2, node3 and node2 is a current primary. Currently the code finds out that node2 is primary, connects to it and executes command like "add node4 to the replica set". Will it continue to work if code connects to rs0/node1:27017,node2:27017,node3:27017? In this case it will connect to node2 or node1?

rs0/node1,node2,node3... means that mongo client internally tries to connect to connect each node (I was playing with to try figure out what is the order and I don't know the rule - simply first node in address don't have to be first checked) and it connects and this node is in replset with adequate name, the client get members of this replset and in the future actions it tries to connect to these new members (no other nodes from original address are examined).
Client then keeps monitoring the replicaset and update own host variable. 

So to answer your question: yes, it will work completely fine.

2.4/root/usr/share/container-scripts/mongodb/common.sh
bparees: does "endpoints" return the comma separated format that mongo needs?

php-coder: `endpoints` returns newline-separated values.

bparees: @php-coder yeah that's what i thought.  so does this actually work?

php-coder: @bparees Yes, it should, because later we'll use it like `echo $var | sed` and there we intentionally not using double quotes and content printed in the single line.

bparees: i've learned something today. thanks :)

bparees:
definitely would like @rhcarvalho to sign off on this one since he's doing a lot of mongo clustering work right now.

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: @php-coder Thanks, variable made local.

rhcarvalho:
@bparees I'm doing a code review now, though @php-coder should have spotted it all already :)

In any case, can we delegate functional testing to AOS QE? I don't have bandwidth for that + they've got a list of scenarios to validate against.

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -113,40 +113,23 @@ function mongo_initiate() {
...
-    echo >&2 "Cannot get address of primary node: after checking all nodes we don't have the address"
+# replset_addr return the address of the current replSet
+function replset_addr() {
+  local current_endpoints
+  current_endpoints="$(endpoints)"
+  if [ -z "${current_endpoints}" ]; then
+    echo >&2 "Cannot get address of replica set: no nodes are listed in service"
     return 1
   fi
+  echo "${MONGODB_REPLICA_NAME}/$(echo ${current_endpoints} | sed -e 's| |:27017,|g' -e 's|$|:27017|')"
```
rhcarvalho: :exclamation: reminder to self: look at this again, seems unnecessarily complex.

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -113,40 +113,23 @@ function mongo_initiate() {
...
 }
 
 # mongo_remove removes the current MongoDB from the cluster
 function mongo_remove() {
   local primary_addr
   # if we cannot determine the IP address of the primary, exit without an error
   # to allow callers to proceed with their logic
-  primary_addr="$(mongo_primary_member_addr || true)"
+  primary_addr="$(replset_addr || true)"
```
rhcarvalho: :+1: this is good. We indeed don't need to know which one is the primary to remove a member from the replica set.

However, the name of the variable is misleading now.
@omron93 please rename it, perhaps to `host`? (because that's the flag name where we'll use it)

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -164,7 +147,7 @@ function mongo_add() {
...
```
rhcarvalho: Same here, needs to be renamed.

2.4/root/usr/share/container-scripts/mongodb/initiate_replica.sh
```diff
@@ -66,7 +66,7 @@ echo "=> Waiting for the new PRIMARY to be elected ..."
 mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --quiet --host ${mongo_node} --eval "var done=false;while(done==false){var members=rs.status().members;for(i=0;i<members.length;i++){if(members[i].stateStr=='PRIMARY' && members[i].name!='$(mongo_addr)'){done=true}};sleep(500)};" &>/dev/null
 
 # Remove the initialization container MongoDB from cluster and shutdown
-echo "=> The new PRIMARY member is $(mongo_primary_member_addr), shutting down current member ..."
+echo "=> The new PRIMARY member elected, shutting down current member ..."
```
rhcarvalho: "**A** new PRIMARY member **was elected**, shutting down current member ..."

Still, "current member" is really strange... but we don't need to bike shed on that.

2.4/root/usr/share/container-scripts/mongodb/common.sh
rhcarvalho: [rant]

Funny that the format that the `mongo` CLI client takes differs from connection strings used by other clients...
- https://docs.mongodb.com/v2.4/reference/program/mongo/#cmdoption--host

```
--host <replSetName>/<hostname1><:port>,<hostname2><:port>,<...>
```

x
- https://docs.mongodb.com/v2.4/reference/connection-string/#connection-string-options

```
mongodb://db1.example.net,db2.example.net:2500/?replicaSet=test
```

[/rant]

rhcarvalho: Looking at the docs, we see that the PORT is optional. Since the point of the PR is to simplify matters, we can simplify even further by not using `sed` to insert the default port number in hard-coded form.

Let's keep in mind that the hard-coded port number is really an arbitrary limitation, but one that makes sense in terms of being pragmatic. One could run their mongod servers in any port using OpenShift/Docker. But handling arbitrary ports would require furthers complications to the Bash script.

All that is left then is to replace all spaces with comma?

``` bash
  echo "${MONGODB_REPLICA_NAME}/${current_endpoints//[[:space:]]/,}"
```

`dig` returns multiple IPs in separate lines, and the `[[:space:]]` pattern captures that (while also handling spaces, tabs, etc).

rhcarvalho:
@bparees @omron93, code review done.
- [x] simplify implementation of `replset_addr`
- [x] rename variable 2x
- [x] improve log message
- [x] apply all of the above to all copies/versions (2.4, ...)

Again, I haven't run the code, but I believe it should work. Will defer functional tests to AOS QE if possible.

Thanks!

bparees:
> In any case, can we delegate functional testing to AOS QE? I don't have bandwidth for that + they've got a list of scenarios to validate against.

yes.  @hhorak can you work with AOS QE to get some regression scenarios run on these changes?

omron93:
@rhcarvalho Thanks  for review.

All should be fixed/changed.

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -113,68 +113,51 @@ function mongo_initiate() {
...
-    echo >&2 "Cannot get address of primary node: after checking all nodes we don't have the address"
+# replset_addr return the address of the current replSet
+function replset_addr() {
+  local current_endpoints
+  current_endpoints="$(endpoints)"
+  if [ -z "${current_endpoints}" ]; then
+    echo >&2 "Cannot get address of replica set: no nodes are listed in service"
     return 1
   fi
+  echo "${MONGODB_REPLICA_NAME}/$(echo ${current_endpoints//[[:space:]]/,}"
```
rhcarvalho: No need to call `echo` twice and the subshell.

The original proposal was:

``` bash
  echo "${MONGODB_REPLICA_NAME}/${current_endpoints//[[:space:]]/,}"
```

That's a single command running the current shell process.

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -113,68 +113,51 @@ function mongo_initiate() {
...
+  host="$(replset_addr || true)"
+  if [ -z "$host" ]; then
     return
   fi
 
   local mongo_addr
   mongo_addr="$(mongo_addr)"
 
-  echo "=> Removing ${mongo_addr} on ${primary_addr} ..."
+  echo "=> Removing ${mongo_addr} on ${host} ..."
```
rhcarvalho: This message now will look a bit weird on the console output, as it will read something like:

```
=> Removing 81.114.31.36 on rs0/81.114.31.36,81.114.31.35,81.114.31.37
```

Maybe we should omit the host part and just say something like:

``` bash
  echo "=> Removing ${mongo_addr} from replica set ..."
```

WDYT?

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -113,68 +113,51 @@ function mongo_initiate() {
...
+  host="$(replset_addr || true)"
+  if [ -z "$host" ]; then
     return
   fi
 
   local mongo_addr
   mongo_addr="$(mongo_addr)"
 
-  echo "=> Adding ${mongo_addr} to ${primary_addr} ..."
+  echo "=> Adding ${mongo_addr} to ${host} ..."
```
rhcarvalho: Same here about console output.

rhcarvalho:
@omron93 nice stuff! Please update based on latest comments, and you get a LGTM from me :)

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: You are right. Fixed.

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: Looks better.

omron93:
@rhcarvalho @bparees Should be fixed.

rhcarvalho:
LGTM, pending squash and QE testing.

omron93:
Squashed.

wanghaoran1988:
I finished the testing today, no issues found, thanks

bparees:
[test]

bparees:
needs a rebase due to https://github.com/sclorg/mongodb-container/pull/177

omron93:
@bparees Rebased.

bparees:
[test]

bparees:
tests are failing after rebase..

rhcarvalho:
[test] again? The console output is so dirty I can't see what failed.

rhcarvalho:
@bparees looks like it was a test flake!

rhcarvalho:
Just for reference, this is the run that failed: https://ci.openshift.redhat.com/jenkins/job/mongodb/310/console

PR #130, opened 2016-02-25
Fix determination of the primary member

php-coder:
Exclude ourself from the list of endpoints and don't try to connect our own host.

@bparees @mfojtik PTAL.

Thank you @rhcarvalho for a session of debugging this case.

Fixes #125

PR #178, opened 2016-09-13
[WIP] Allow local replicaset without kubernetes.

omron93:
This PR allows user to create replset using only local docker.

If variable  `MONGODB_REPLICA_MEMBERS` is specified (it could be even empty) run-mongod does not use kubernetes to obtain pods IP addresses and read IPs from this environmental variable.

So creating replset:

```
docker run ... -e MONGODB_REPLICA_NAME=rs0 -e MONGODB_KEYFILE_VALUE=... -e MONGODB_REPLICA_MEMBERS="" centos/mongodb-24-centos7
docker run ... -e MONGODB_REPLICA_NAME=rs0 -e MONGODB_KEYFILE_VALUE=... -e MONGODB_REPLICA_MEMBERS="MEMBER1_IP" centos/mongodb-24-centos7
docker run ... -e MONGODB_REPLICA_NAME=rs0 -e MONGODB_KEYFILE_VALUE=... -e MONGODB_REPLICA_MEMBERS="MEMBER1_IP,MEMBER2_IP" centos/mongodb-24-centos7
docker run ... -e MONGODB_REPLICA_NAME=rs0 -e MONGODB_KEYFILE_VALUE=... -e MONGODB_REPLICA_MEMBERS="MAMBER1_IP,MEMBER2_IP,MEMBER3_IP" centos/mongodb-24-centos7 run-mongod initialize
```

Only initializing member requires IPs of all members, other does not require this (they know own IP) but more IPs members know the better (more robust in case of some problems).

Also test for testing local replset was added.

@php-coder Please, do you remember why did you add `grep -v "$(container_addr)"` to `mongo_primary_member_addr()` in #130 ? (why local mongod IP is excluded?)

### DIFF
--- a/2.4/root/usr/bin/run-mongod-replication
+++ b/2.4/root/usr/bin/run-mongod-replication
@@ -67,6 +67,7 @@ ${CONTAINER_SCRIPTS_PATH}/init-replset.sh "${1:-}" &
 
 # Run `mongod` in a subshell because MONGODB_ADMIN_PASSWORD should still be
 # defined when the trapped call to `cleanup` references it.
+set -m
 (
   # Make sure env variables don't propagate to mongod process.
   unset MONGODB_USER MONGODB_PASSWORD MONGODB_DATABASE MONGODB_ADMIN_PASSWORD
--- a/2.4/root/usr/bin/run-mongod-replication
+++ b/2.4/root/usr/bin/run-mongod-replication
@@ -36,6 +36,8 @@ function cleanup() {
   echo "=> Shutting down MongoDB server ..."
   pkill -INT mongod || :
   wait_for_mongo_down
+
+  ps -p ${supervisor_pid:-} && kill ${supervisor_pid}
   exit 0
 }
 
@@ -58,16 +60,19 @@ fi
 setup_keyfile
 
 # Initialize the replica set or add member in the background.
-${CONTAINER_SCRIPTS_PATH}/init-replset.sh "${1:-}" &
+${CONTAINER_SCRIPTS_PATH}/init-replset.sh "${1:-}" & supervisor_pid=$!
 
 # TODO: capture exit code of `init-replset.sh` and exit with an error if the
 # initialization failed, so that the container will be restarted and the user
 # can gain more visibility that there is a problem in a way other than just
 # inspecting log messages.
 
+# Don't pass signals to background processes - to succesfully remove local `mongod`
+# from replset in `cleanup()` mongod has to be running (to get addresses of other members).
+# Background processes are killed in `cleanup`.
+set -m
 # Run `mongod` in a subshell because MONGODB_ADMIN_PASSWORD should still be
 # defined when the trapped call to `cleanup` references it.
-set -m
 (
   # Make sure env variables don't propagate to mongod process.
   unset MONGODB_USER MONGODB_PASSWORD MONGODB_DATABASE MONGODB_ADMIN_PASSWORD
--- a/2.4/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.4/root/usr/share/container-scripts/mongodb/common.sh
@@ -86,8 +86,12 @@ function _wait_for_mongo() {
 # To get list of endpoints, you need to have headless Service named 'mongodb'.
 # NOTE: This won't work with standalone Docker container.
 function endpoints() {
-  service_name=${MONGODB_SERVICE_NAME:-mongodb}
-  dig ${service_name} A +search +short 2>/dev/null
+  if [ -n "${MONGODB_REPLICA_MEMBERS+x}" ]; then
+    echo "${MONGODB_REPLICA_MEMBERS} $(container_addr)"
+  else
+    service_name=${MONGODB_SERVICE_NAME:-mongodb}
+    dig ${service_name} A +search +short 2>/dev/null
+  fi
 }
 
 # build_mongo_config builds the MongoDB replicaSet config used for the cluster
--- a/2.4/test/run
+++ b/2.4/test/run
@@ -387,12 +387,13 @@ function wait_for_secondary() {
 }
 
 function wait_for_connection() {
-    local name
-    name=$1
     local host
-    host=$2
+    host=$1
 
-    while ! mongo_cmd_local ${name} "admin --host ${host} --eval 'quit()'"; do
+    DB=admin
+    CONTAINER_IP=${host}
+
+    while ! docker run --rm $IMAGE_NAME mongo "$DB" --host ${host} --eval 'quit()'; do
       sleep 5
     done
 }
@@ -410,6 +411,11 @@ function run_local_replication_test() {
 -e MONGODB_KEYFILE_VALUE=xxxxxxxxxxxx
 -e MONGODB_SMALLFILES=true"
 
+    # need to set these because `mongo_cmd` relies on global variables
+    USER=admin
+    PASS=adminPassword
+    DB=admin
+
     docker run -d --cidfile $CIDFILE_DIR/replset0 ${BASE_ARGS} -e MONGODB_REPLICA_MEMBERS="" $IMAGE_NAME run-mongod-replication
     docker run -d --cidfile $CIDFILE_DIR/replset1 ${BASE_ARGS} -e MONGODB_REPLICA_MEMBERS="$(get_container_ip replset0)" $IMAGE_NAME run-mongod-replication
     docker run -d --cidfile $CIDFILE_DIR/replset2 ${BASE_ARGS} -e MONGODB_REPLICA_MEMBERS="$(get_container_ip replset0) $(get_container_ip replset1)" $IMAGE_NAME run-mongod-replication
@@ -420,9 +426,9 @@ function run_local_replication_test() {
     insert_and_wait_for_replication "rs0/$(get_container_ip replset1),$(get_container_ip replset2)" "{a:5, b:10}"
 
     # Testing stored documents in each node
-    [ "$(mongo_cmd_local replset0 '-u admin -p adminPassword admin --eval "rs.slaveOk(); printjson(db.getSiblingDB(\"test_db\").data.count())"' | tail -n 1)" == "1" ]
-    [ "$(mongo_cmd_local replset1 '-u admin -p adminPassword admin --eval "rs.slaveOk(); printjson(db.getSiblingDB(\"test_db\").data.count())"' | tail -n 1)" == "1" ]
-    [ "$(mongo_cmd_local replset2 '-u admin -p adminPassword admin --eval "rs.slaveOk(); printjson(db.getSiblingDB(\"test_db\").data.count())"' | tail -n 1)" == "1" ]
+    [ "$(CONTAINER_IP=$(get_container_ip replset0) mongo_cmd 'rs.slaveOk(); printjson(db.getSiblingDB("test_db").data.count())' | tail -n 1)" == "1" ]
+    [ "$(CONTAINER_IP=$(get_container_ip replset1) mongo_cmd 'rs.slaveOk(); printjson(db.getSiblingDB("test_db").data.count())' | tail -n 1)" == "1" ]
+    [ "$(CONTAINER_IP=$(get_container_ip replset2) mongo_cmd 'rs.slaveOk(); printjson(db.getSiblingDB("test_db").data.count())' | tail -n 1)" == "1" ]
 
     # Stopping and removing second container
     docker stop $(get_cid replset1)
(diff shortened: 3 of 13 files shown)


2.4/root/usr/bin/run-mongod-replication
```diff
@@ -68,6 +68,7 @@ ${CONTAINER_SCRIPTS_PATH}/replset_supervisor.sh "${1:-}" "$$" &
 # Run `unset_env_vars` and `mongod` in a subshell because
 # MONGODB_ADMIN_PASSWORD should still be defined when the trapped call to
 # `cleanup` references it.
+set -m
```
php-coder, line 73: `set -o monitor`? And why do we need it at all? Maybe put a comment here about this too? Looks like a wizardry..

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -86,8 +86,12 @@ function _wait_for_mongo() {
 # To get list of endpoints, you need to have headless Service named 'mongodb'.
 # NOTE: This won't work with standalone Docker container.
 function endpoints() {
-  service_name=${MONGODB_SERVICE_NAME:-mongodb}
-  dig ${service_name} A +search +short 2>/dev/null
+  if [ -n "${MONGODB_REPLICA_MEMBERS+x}" ]; then
```
php-coder, line 89: This is look cryptic.. Why we can't use `[ -n "${MONGODB_REPLICA_MEMBERS:-}" ]` instead?

2.4/test/run
```diff
@@ -342,10 +342,121 @@ noprealloc = true" > $config_file
     echo "  Success!"
 }
 
+function insert_and_wait_for_replication() {
+    local host
+    host=$1
+    local data
+    data=$2
```
php-coder, line 343: IMHO it's better to add double quotes here (and everywhere.. )

2.4/test/run
```diff
@@ -342,10 +342,121 @@ noprealloc = true" > $config_file
...
+
+    # Storing document into replset and wait replication to finish
+    CONTAINER_IP="${host}" ADMIN_PASS="adminPassword" \
+    mongo_admin_cmd "db.getSiblingDB('test_db').data.insert(${data});
+        while(true) {
+          var status=rs.status();
+          var optime=status.members[0].optime;
+          var ok=true;
+          for(var i=1; i < status.members.length; i++) {
```
php-coder, line 353: Do you intentionally skip the first member, do you?

2.4/test/run
```diff
@@ -342,10 +342,121 @@ noprealloc = true" > $config_file
...
+function insert_and_wait_for_replication() {
+    local host
+    host=$1
+    local data
+    data=$2
+
+    # Storing document into replset and wait replication to finish
+    CONTAINER_IP="${host}" ADMIN_PASS="adminPassword" \
+    mongo_admin_cmd "db.getSiblingDB('test_db').data.insert(${data});
+        while(true) {
```
php-coder: Infinity loop? Maybe we can introduce a counter with sleep?

php-coder:
> @php-coder Please, do you remember why did you add grep -v "$(container_addr)" to mongo_primary_member_addr() in #130 ? (why local mongod IP is excluded?)

Yes, I remember. `mongo_add` and `mongo_remove` invoke `mongo_primary_member_addr` to get an IP of the primary. When `mongo_primary_member_addr` fails, they will fail too. With the algorithm that we used, `mongo_primary_member_addr` tried to connect to each of the hosts and sometimes it tried to connect to himself and failed. I excluded node own IP because it was 1) useless 2) error-prone

I'm not sure that whether we need it or not after your latest PR for changing the algorithm for finding the primary.

2.4/test/run
praiskup: Definitely not.  That is proper way of variable assignment.

2.4/root/usr/share/container-scripts/mongodb/common.sh
praiskup: What exactly is cryptic?  :)  I would care about the '+' vs ':+' difference ('-' vs ':-' resp.).

php-coder: I agree that `:-` also cryptic but it's used more often, looks more idiomatic and I don't need to open `man bash` again to find what `+` means in this context.

2.4/test/run
praiskup: I wouldn't separate that assignments into separate lines when you can use 'local X=$Y', though.

2.4/root/usr/share/container-scripts/mongodb/common.sh
praiskup: This is rather personal preference (the '+' vs '-'), so just nitpick.  In usual shell scripting the ':-' statement is used rather properly, here we have the ugly `nounset` so the ':-' is rather overused.

php-coder: It's not only personal preference but also the style that is used in this repository:

``` bash
$ git grep -c ':-}' | wc -l
16
$ git grep -c '+[a-z]\+}' | wc -l
0
```

All of these cases are cryptic but it's better to use the same approach everywhere in the repository instead of different styles.

praiskup: Well, I really fail to see what is cryptic on variable substitution (be that minus or plus) ;) it is like you said we are going to use just 'if' branches, and never else branches.

omron93: > Why we can't use [ -n "${MONGODB_REPLICA_MEMBERS:-}" ] instead?

@php-coder The if condition have to be true even for set but null value (for MONGODB_REPLICA_MEMBERS=""), therefore I used +.

Something different should be used?

2.4/test/run
omron93: > IMHO it's better to add double quotes here (and everywhere.. )

@praiskup @php-coder Could assigning without double quotes cause any troubles?

2.4/test/run
omron93: Yes and no:-)

First member value is stored `optime` variable. And values of every other member are compared to this first member value.

2.4/test/run
omron93: Good idea. I haven't found any easy `sleep` equivalent in JS, so using only counter (it should be enough).

2.4/root/usr/bin/run-mongod-replication
omron93: @php-coder It is now more explained in code

```
# Don't pass signals to background processes - to succesfully remove local `mongod`
# from replset in `cleanup()` mongod has to be running (to get addresses of other members).
# Background processes are killed in `cleanup`.
```

omron93:
> I'm not sure that whether we need it or not after your latest PR for changing the algorithm for finding the primary.

No it is not needed. Replset addressing allows to have IP of 'dead' pods during connections (mongo automatically selects right IPs).

rhcarvalho (changes requested):
- [ ] Pending a rebase
- [ ] Needs user documentation

omron93:
Rebased.

@rhcarvalho Where documentation should be placed? Is it OK to put it into README.md of each version or create a new README for localhost replication?

BTW I would like to postpone this PR because I would like to use initialization of replset without one initiator which terminates. So I want to wait for #184, how it evolve and if it will use usable scripts (common for initialization without post hook).

rhcarvalho:
> Where documentation should be placed?

We have some documentation for the image in the OpenShift docs [1], but since this is about a standalone usage of the image, I don't know of any other place than this repository itself.

[1] https://docs.openshift.com/enterprise/3.2/using_images/db_images/mongodb.html

The README(s) are what we can work with, or open for proposing a different way to organize the image documentation.

> BTW I would like to postpone this PR

No problem. Perhaps prefix the title with [WIP]?

omron93:
@rhcarvalho @php-coder So after petset PR is merged how to integrate this PR? Do you have some ideas how to unify all free way of replset initialization to avoid having three different set of scripts?

rhcarvalho:
Before diving into the implementation details, could I ask for some documentation like [examples/petset/README.md](https://github.com/sclorg/mongodb-container/blob/master/examples/petset/README.md)?

I think it would be useful to understand first the differences from the user's point of view, so that then we can see how to best support all use cases.

E.g., one point of discussion is what happens when a container is terminated.

omron93:
> Before diving into the implementation details, could I ask for some documentation like examples/petset/README.md?
> 
> I think it would be useful to understand first the differences from the user's point of view, so that then we can see how to best support all use cases.

OK. I will try to prepare something this week.

> E.g., one point of discussion is what happens when a container is terminated.

Container should be removed from replset configuration.

omron93:
@rhcarvalho @php-coder Added first draft of README for Docker replication.

Scripts don't work in this described way yet. First I want to know you feedback about usage. And how to structure code for different way of replset deployment.

Also I don't have as much time for containers now (sry), so hopefully next week I will update this PR...
Current implementation do not work - replication scripts does not wait properly for replication, so creating users fails... So another thing to fix.

examples/replset-docker/README.md
```diff
@@ -0,0 +1,81 @@
+# MongoDB Replication Example Using a Docker
+
+This [MongoDB replication](https://docs.mongodb.com/manual/replication/) example
+uses a [Docker engine](http://docker.com) to run replica set members. This platform
+do not provide any deployment automation and it is used mainly for developing and testing.
```
rhcarvalho, line 5: Why would one use Docker and not use at least Docker Compose?!

examples/replset-docker/README.md
```diff
@@ -0,0 +1,81 @@
...
+    docker inspect --format='{{.NetworkSettings.IPAddress}}' ${id}
+}
+
+REPLICATION_ARGS="
+-e MONGODB_DATABASE=db
+-e MONGODB_USER=user
+-e MONGODB_PASSWORD=password
+-e MONGODB_ADMIN_PASSWORD=adminPassword
+-e MONGODB_REPLICA_NAME=rs0
+-e MONGODB_KEYFILE_VALUE=xxxxxxxxxxxx
```
rhcarvalho, line 31: This one we need to write a note about / link to MongoDB's docs, and perhaps the script can generate it.

examples/replset-docker/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+You will need an Docker engine where you can run containers. If you want to avoid running all replset members on one host, you can also use [Docker Swarm](https://docs.docker.com/swarm/).
+
+## Example Working Scenarios
+
+This section describes how this example is designed to work.
+
+### Initial Deployment: 3-member Replica Set
+
+To create replica set with three members you can use this script:
```
rhcarvalho, line 17: If a script is needed, perhaps it is better/more convenient to have it as a file in the repo than a block-of-text in the README.

examples/replset-docker/README.md
```diff
@@ -0,0 +1,81 @@
...
+```console
+sh-4.2$ mongo $MONGODB_DATABASE -u $MONGODB_USER -p $MONGODB_PASSWORD --host $MONGODB_REPLICA_NAME/localhost
+MongoDB shell version: 3.2.6
+connecting to: sampledb
+rs0:PRIMARY>
+```
+
+Note: You can also use host version of mongo shell, but you have to substitute values of environmental variables yourself.
+
+During the lifetime of your deployment, one or more of those members might crash or fail. In this case, replica set member is removed. It some special cases, replica set configuration might not be updated -- see Known Limitations.
```
rhcarvalho, line 52: Why is it removed?

Wouldn't it be a more reasonable workflow that either:
1) set docker settings to automatically restart the container
2) operator manually restart the container
?

AFAIK in normal MongoDB deployments, you don't automatically remove members from the replica set config when they go down. Instead, you bring them back up.

examples/replset-docker/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+**Note**: for production usage, you should maintain as much separation between
+members as possible. It is recommended to run containers on different hosts.
```
rhcarvalho, line 55: It is very very unrealistic to tell people to use a single-host Docker installation, and then do that.
IDK if using Docker Swarm you can influence the choice of host the container will run in, but anyway that doesn't seem to be the focus of this example.

examples/replset-docker/README.md
rhcarvalho: Why not using Atomic Host?

examples/replset-docker/README.md
```diff
@@ -0,0 +1,81 @@
...
+REPLICATION_ARGS="
+-e MONGODB_DATABASE=db
+-e MONGODB_USER=user
+-e MONGODB_PASSWORD=password
+-e MONGODB_ADMIN_PASSWORD=adminPassword
+-e MONGODB_REPLICA_NAME=rs0
+-e MONGODB_KEYFILE_VALUE=xxxxxxxxxxxx
+-e MONGODB_SMALLFILES=true"
+
+docker run -d --name replset0 ${BASE_ARGS} --hostname initialize -e MONGODB_REPLICA_MEMBERS="" $IMAGE_NAME run-mongod-replication
```
rhcarvalho, line 34: Does this mean that `run-mongod-replication` is supposed to inspect the `hostname` to decide whether to initialize the replica set?

omron93: > Does this mean that run-mongod-replication is supposed to inspect the hostname to decide whether to initialize the replica set?

Yes. Or it could be possible (maybe better?) to pass "initiate" parameter to `run-mongod-replication` script.

The main difference is that this container does not terminate after initialization.

examples/replset-docker/README.md
omron93: > Why is it removed?
> Wouldn't it be a more reasonable workflow that either:
> 1) set docker settings to automatically restart the container

I was not aware of this. It would be good to mention this in README.

2) operator manually restart the container

It can. Nobody prevents him to do it.

The restarted container probably gets a new IP, so it have to be added to replica set again.

> AFAIK in normal MongoDB deployments, you don't automatically remove members from the replica set config when they go down. Instead, you bring them back up.

In normal MongoDB deployments member IS NOT added to replica set on startup. I've heard many times that deploying in containers is different :-)

I see no difference between some failure of container and terminating of container.

examples/replset-docker/README.md
omron93: You are right, I will remove this Note.

examples/replset-docker/README.md
omron93: @rhcarvalho It is not needed. This is only example and user can use different script to start replica set.

examples/replset-docker/README.md
omron93: IMPOV it should be documented in `run-mongod-replication` help command or in some shared place. Not in every README.

It is same as other environmental variables.

examples/replset-docker/README.md
omron93: > Why would one use Docker and not use at least Docker Compose?!

@rhcarvalho Because docker is IMPOV "standard" for developers, Docker Compose requires another program to install and things to read about.
It definitely would be good to mention Docker Compose and ideally prepare configuration for it to be able to easily start replication.

I would like to prepare same way for deploying replication as other database containers offers (mariadb, mysql, postgresql).

rhcarvalho:
I'm not particularly against this work, but the premise seems weak.

Why are we trying to support a MongoDB replication scheme for "[standard developers that don't want to read about things](https://github.com/sclorg/mongodb-container/pull/178#discussion_r85773021)" making choices that can hardly work even for setting up a development / throw away cluster?

If I wanted to play with a replica set to see how it works in my development machine, why would I use this instead of e.g., https://github.com/rueckstiess/mtools?

hhorak:
> Why are we trying to support a MongoDB replication scheme for "standard developers that don't want to read about things" making choices that can hardly work even for setting up a development / throw away cluster?

One of the basic requirements for the RHSCL containers is that they should provide the same feature set regardless of environment, be it OpenShift, Atomic Host, or pure RHEL/CentOS. People might always use k8s, but they simply don't do it in practice, especially for simple deployments and development. They use the docker daemon directly, not even swarm (which is not in RHEL/CentOS btw). Having the replication working without k8s will make it much more consumable for those users and it will be valuable especially during transition to container environment. And as a bonus, it will allow us to run tests of the replication for PRs.

omron93:
Closing in favor of #224

PR #177, opened 2016-08-26
Split main scripts for standalone server and replication.

omron93:
The aim of this PR is to split run-mongod into main replication and standalone script.

I would like to:
1) Make code more clean and readable
- current situation:
  - run-mongod initiate -> execed into initiate_replica.sh -> it run run-mongod in background -> this background run-mongod after some coding execked mongod
  - run-mongod (replicated) -> run in background replica_supervisor.sh
  - run-mongod -> run mongod in background -> create users -> exec mongod
  
  Main code is in one script and decision is made environmental variables. Also I don't like that if I want to see a flow of the script that I have to look at 3+ files to find it.
  - in this PR:
    - run-mongod -> run mongod in background -> create users -> exec mongod
  - run-mongod-replication -> run replset_supervisor.sh in background (according to first parameter (initiate) - initiates replicaset or tries to add current node to replset ) and exec mongod

2) Separating these scripts allows us to easily add other scenarios - for example add sharding support -> new script run-mongod-sharding ...

3) Make mongodb more similar to other databases (mysql and postgresql both have separated scripts for standalone and master/slave replication)

**I didn't add new code** (only restructured old one)
- run-mongod and run-mongod-replication are only splited (removed if statement) original run-mongod
- replset_supervisor.sh is joined replica_supervisor.sh and initiate_replica.sh and added if according to first argument of run-mongod-replication
- from common.sh removed unneeded function run_mongod_supervisor

@hhorak @bparees @rhcarvalho @php-coder Please take a look.

PR #180, opened 2016-09-15
Fix replication

### LINKED ISSUE #179: mongo replica tests are failing
https://ci.openshift.redhat.com/jenkins/job/origin_extended_image_tests/549/consoleFull

probably due to either:
https://github.com/sclorg/mongodb-container/commit/2e55696d2218ad38c4fe3b3db7c34ccccc3e95d4

or

https://github.com/sclorg/mongodb-container/commit/df104523417431ac013733b5478e483124b048c8

@omron93 @php-coder

php-coder:
This fixes MongoDB replication that was broken by df104523417431ac013733b5478e483124b048c8 (#177) commit.

Before this change we run the `replset_supervisor.sh` script in the background and after that run the mongodb daemon. Because `replset_supervisor.sh` script connect to mongo it was possible that database wasn't ready yet and script was failing. I updated the script for waiting when mongo will accept connections.

Replica set started working but only for a short time and later it stopped working again because mongo process was killed to early. To fix this I removed the code from `replset_supervisor.sh` script that kills mongo.

Fix #179 

PTAL @bparees 

@omron93 Why did we kill mongo from the `replset_supervisor.sh` script?

### DIFF
--- a/2.4/root/usr/bin/run-mongod-replication
+++ b/2.4/root/usr/bin/run-mongod-replication
@@ -63,7 +63,7 @@ fi
 setup_keyfile
 
 # Run the supervisor in background to manage replset
-${CONTAINER_SCRIPTS_PATH}/replset_supervisor.sh "${1:-}" "$$" &
+${CONTAINER_SCRIPTS_PATH}/replset_supervisor.sh "${1:-}" &
 
 # Run `unset_env_vars` and `mongod` in a subshell because
 # MONGODB_ADMIN_PASSWORD should still be defined when the trapped call to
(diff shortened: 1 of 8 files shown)


bparees:
lgtm, will wait for @omron93 to weigh in before merging.

bparees:
@omron93 i'm merging this to get our openshift tests unbroken but if you have concerns please let us know and we can revert/revisit the change.

omron93:
> @omron93 Why did we kill mongo from the replset_supervisor.sh script?

This `kill ${main_process_id}` line was there to call `cleanup()` in main script. I was wrong and this line was redundant. `cleanup()` is called anyway.

However I do not see a difference. Before calling the removed line, mongod was already down!  [1]
@php-coder So what changed? Why do you think previous implementation failed?

[1] https://github.com/sclorg/mongodb-container/blob/master/2.4/root/usr/share/container-scripts/mongodb/replset_supervisor.sh#L70

omron93:
> @php-coder So what changed? Why do you think previous implementation failed?

Sry, I got it. `wait_for_mongo_up` was missing. I am very sorry, I've removed it by mistake when I was removing running run-mongod  in background (`MONGODB_NO_SUPERVISOR=1 MONGODB_NO_AUTH=1 run-mongod mongod &`) during replset initialization.

PR #182, opened 2016-09-19
Minor improvements

php-coder:
- `run-mongod(usage)`: remove unused optional variables from the output
- `run-mongod(cleanup)`: don't try to remove instance from the replica set
- `run-mongod-replication(usage)`: add `MONGODB_KEYFILE_VALUE` to the list
  of required parameters
- `common.sh(mongo_create_user)`: don't try to show value of empty variable

PTAL @bparees @omron93

### DIFF
--- a/2.4/root/usr/bin/run-mongod
+++ b/2.4/root/usr/bin/run-mongod
@@ -12,9 +12,6 @@ function usage() {
   echo "  MONGODB_PASSWORD"
   echo "  MONGODB_DATABASE"
   echo "  MONGODB_ADMIN_PASSWORD"
-  echo "Optional variables:"
-  echo "  MONGODB_REPLICA_NAME"
-  echo "  MONGODB_INITIAL_REPLICA_COUNT"
   echo "MongoDB settings:"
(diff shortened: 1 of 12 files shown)


2.4/root/usr/bin/run-mongod
```diff
@@ -28,9 +25,6 @@ function unset_env_vars() {
 }
 
 function cleanup() {
-  if [ -n "${MONGODB_REPLICA_NAME-}" ]; then
-    mongo_remove
-  fi
```
php-coder, line 33: @omron93 Correct me if I'm wrong that this script is always gets executed for standalone mongodb installation. That's why I'm removing the code that is useful only for replicaset.

bparees:
lgtm.  @omron93 ?

2.4/root/usr/bin/run-mongod
omron93: You are right.

omron93:
@bparees @php-coder LGTM. Thanks for cleaning the code.

PR #185, opened 2016-09-22
Initialize replset without deploy hook.

omron93:
This PR is only to show possibility of creating replset without post deploy hook. It is only fast draft, only to get feedback from others.

@php-coder @bparees @rhcarvalho Please take a look and test. I am not sure that it will work (on my local openshift it works), that there is no problem... so I mainly want to get a feedback.

### DIFF
--- a/2.4/examples/replica/mongodb-clustered.json
+++ b/2.4/examples/replica/mongodb-clustered.json
@@ -84,27 +84,6 @@
         "name": "${MONGODB_SERVICE_NAME}"
       },
       "spec": {
-        "strategy": {
-          "type": "Recreate",
-          "recreateParams": {
-            "post": {
-              "failurePolicy": "Retry",
-              "execNewPod": {
-                "command": [
-                  "run-mongod-replication",
-                  "initiate"
-                ],
-                "containerName": "mongodb",
-                "env": [
-                  {
-                    "name": "MONGODB_INITIAL_REPLICA_COUNT",
-                    "value": "3"
-                  }
-                ]
-              }
-            }
-          }
-        },
         "triggers": [
           {
             "type": "ConfigChange"
--- a/3.2/root/usr/share/container-scripts/mongodb/common.sh
+++ b/3.2/root/usr/share/container-scripts/mongodb/common.sh
@@ -89,36 +89,13 @@ function endpoints() {
   dig ${service_name} A +search +short 2>/dev/null
 }
 
-# build_mongo_config builds the MongoDB replicaSet config used for the cluster
-# initialization.
-# Takes a list of space-separated member IPs as the first argument.
-function build_mongo_config() {
-  local current_endpoints
-  current_endpoints="$1"
-  local members
-  members="{ _id: 0, host: \"$(mongo_addr)\"},"
-  local member_id
-  member_id=1
-  local container_addr
-  container_addr="$(container_addr)"
-  local node
-  for node in ${current_endpoints}; do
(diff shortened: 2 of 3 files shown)


bparees:
what happens if a new pod comes up with a lower ip address and tries to initialize as the master?

what if the pod w/ the lowest ip wasn't registered as an endpoint yet?

omron93:
@bparees That is only one thing I am not sure about and I can't find anywhere how kubernetes do IP assigning and when it lists pod in service.

Making MONGODB_INITIAL_REPLICA_COUNT mandatory for this type of initializing should solve this.

rhcarvalho:
I'd ask the same questions as @bparees. I think this type of initialization has too many race conditions to work in practice.

The guarantees we have with petsets, and semantics of how things work is different.

Without petsets I'd go for multiple services and multiple DCs, but that's really not what @php-coder was working on, the intent for him was to get the petset example working because of persistent storage...

Even though I would too, like @omron93,  prefer to have code in a single place (copy-paste-per-version, I don't like you!), IMHO the conceptual differences are enough to justify a separate path of implementation.

omron93:
> I'd ask the same questions as @bparees. I think this type of initialization has too many race conditions to work in practice.
> Without petsets I'd go for multiple services and multiple DCs, but that's really not what @php-coder was working on, the intent for him was to get the petset example working because of persistent storage.

@rhcarvalho Would having mandatory `MONGODB_INITIAL_REPLICA_COUNT` fit into "documented limitations" mentined in https://github.com/sclorg/mongodb-container/pull/184#issuecomment-249013321? Do you see any problem with using this variable to determine who should initialize? 

> copy-paste-per-version, I don't like you!

I would also to prefer some solution to this. But I don't have any idea how to solve this and there is no result from discussion about this. However IMPOV with some tooling it is not so hard to deal with this.

> IMHO the conceptual differences are enough to justify a separate path of implementation.

Yes, of course.

rhcarvalho:
> @rhcarvalho Would having mandatory MONGODB_INITIAL_REPLICA_COUNT fit into "documented limitations" mentined in #184 (comment)? Do you see any problem with using this variable to determine who should initialize?

I don't see how that solves the problem @bparees raised.

If users would keep MONGODB_INITIAL_REPLICA_COUNT in sync with the number of replicas (so maybe the name should be REPLICA_COUNT, remove "INITIAL" ...), each pod would need to wait until it can see N other pods to determine who is to perform the initialization.
This "waiting" part has several rough edges to solve:
- what if `dig` returns more IPs than MONGODB_INITIAL_REPLICA_COUNT?
- what if `dig` takes an arbitrary long time (infinity?) to return as many IPs as MONGODB_INITIAL_REPLICA_COUNT?
- what if `dig` returns MONGODB_INITIAL_REPLICA_COUNT IPs, pod A thinks it's got the lowest IP, but then pod B gets restarted and gets an even lower IP... does it run initialization again?

We really get into solving a consensus problem if we are to guarantee that all N pods see the same list of IPs when they decide who does initialization.

With petsets and hostnames a pod is independent of all other pods to decide if it should perform initialization.

omron93:
@rhcarvalho I see. Sorry I was somehow missing fact that newly added pods run the same code with same environment. I am ashamed :-)

Now I see that is probably not possible.

Thanks. Closing since this PR is wrong.

rhcarvalho:
> I am ashamed :-)

Don't be! It takes some good discussions and cross-validation to get this moving forward. Thank you!

PR #186, opened 2016-09-23
Fix removing master from replset.

omron93:
This PR aims to fix 'stepping down server from PRIMARY role'.

When member is going to be removed from replset and this member is PRIMARY, it should check that this won't break replset - that replset is able to vote a new PRIMARY.
In current situation while removing PRIMARY it could happen that not all members of replset are up-to-date and replset user can lost some information.

Code which is now running after post deploy hook successfully initialize replset performs removing PRIMARY from replset. So removing this code from `init-replset.sh` and post-hook kills main scripts and this calls `cleanup()` (there the PRIMARY member is clearly removed).

Also this fixes https://bugzilla.redhat.com/show_bug.cgi?id=1260586

@php-coder @rhcarvalho @bparees Please take a look and review.

EDIT: Updated info after rebase to latest master.

### DIFF
--- a/3.2/root/usr/bin/run-mongod-replication
+++ b/3.2/root/usr/bin/run-mongod-replication
@@ -28,6 +28,33 @@ function usage() {
 }
 
 function cleanup() {
+  if [ "$(mongo admin -u admin -p ${MONGODB_ADMIN_PASSWORD} --quiet --eval 'rs.isMaster().ismaster')" == "true" ]; then
+
+    # Wait that some SECONDARY is synced with PRIMARY 
+    echo "=> Waiting for syncing SECONDARY ..."
+    replset_wait_sync
+
+    # Some commands will force MongoDB client to re-connect. This is not working
+    # well in combination with '--eval'. In that case the 'mongo' command will fail
+    # with return code 254.
+    echo "=> Giving up the PRIMARY role ..."
+    mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --quiet --eval "rs.stepDown(120);" &>/dev/null || true
+
+    # Wait till the new PRIMARY member is elected
+    echo "=> Waiting for the new PRIMARY to be elected ..."
+    mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --quiet \
+     --eval "var i = ${MAX_ATTEMPTS};
+      while(i > 0) {
+        var members=rs.status().members;
+        for(i=0;i<members.length;i++){
+          if(members[i].stateStr=='PRIMARY' && members[i].name!='$(mongo_addr)'){
+            quit(0)
+          }
+        };
+        sleep(${SLEEP_TIME}*1000)
+      };" &>/dev/null
+    echo "=> A new PRIMARY member was elected, removing from replset ..."
+  fi
   if [ -n "${MONGODB_REPLICA_NAME-}" ]; then
     mongo_remove
   fi
@@ -56,13 +83,15 @@ fi
 setup_keyfile
 
 # Initialize the replica set or add member in the background.
-${CONTAINER_SCRIPTS_PATH}/init-replset.sh "${1:-}" &
+${CONTAINER_SCRIPTS_PATH}/init-replset.sh "${1:-}" "$$" &
 
 # TODO: capture exit code of `init-replset.sh` and exit with an error if the
 # initialization failed, so that the container will be restarted and the user
 # can gain more visibility that there is a problem in a way other than just
 # inspecting log messages.
 
+# Don't pass signals to background processes. Background processes are killed in `cleanup`.
+set -m
 # Run `mongod` in a subshell because MONGODB_ADMIN_PASSWORD should still be
 # defined when the trapped call to `cleanup` references it.
 (
--- a/3.2/root/usr/share/container-scripts/mongodb/common.sh
+++ b/3.2/root/usr/share/container-scripts/mongodb/common.sh
@@ -130,6 +130,31 @@ function replset_addr() {
   echo "${MONGODB_REPLICA_NAME}/${current_endpoints//[[:space:]]/,}"
 }
 
+# replse_wait_sync wait for at least two members to be up to date (PRIMARY and one SECONDARY)
+function replset_wait_sync() {
+  local host
+  # if we cannot determine the IP address of the primary, exit without an error
+  # to allow callers to proceed with their logic
+  host="$(replset_addr || true)"
+  if [ -z "$host" ]; then
+    return 1
+  fi
+
+  mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --host ${host} \
+    --eval "var i = ${MAX_ATTEMPTS};
+    while(i > 0) {
+      var status=rs.status();
+      var primary_optime=status.members.filter(function(el) {return el.state ==1})[0].optime;
+      // Check that at least one member has same optime as PRIMARY (PRIMARY and one SECONDARY ~ >= 2)
+      if(status.members.filter(function(el) {return el.optime.ts.tojson() == primary_optime.ts.tojson()}).length >= 2)
+        quit(0);
+      else
+        sleep(${SLEEP_TIME}*1000);
+      i--;
+    };
+    quit(1);"
+}
+
 # mongo_remove removes the current MongoDB from the cluster
 function mongo_remove() {
   local host
--- a/3.2/root/usr/share/container-scripts/mongodb/init-replset.sh
+++ b/3.2/root/usr/share/container-scripts/mongodb/init-replset.sh
@@ -46,29 +46,6 @@ function initiate() {
   mongo_create_admin
   mongo_create_user "-u admin -p ${MONGODB_ADMIN_PASSWORD}"
 
-  echo "=> Waiting for replication to finish ..."
-  # TODO: Replace this with polling or a Mongo script that will check if all
-  #       members of the cluster are now properly replicated (user accounts are
-  #       created on all members).
-  sleep 10
-
-  # Some commands will force MongoDB client to re-connect. This is not working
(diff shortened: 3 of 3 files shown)


3.2/root/usr/bin/run-mongod-replication
```diff
@@ -34,6 +34,19 @@ function unset_env_vars() {
...
+
+    # Some commands will force MongoDB client to re-connect. This is not working
+    # well in combination with '--eval'. In that case the 'mongo' command will fail
+    # with return code 254.
+    echo "=> Giving up the PRIMARY role ..."
+    mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --quiet --eval "rs.stepDown(120);" &>/dev/null || true
+
+    # Wait till the new PRIMARY member is elected
+    echo "=> Waiting for the new PRIMARY to be elected ..."
+    mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --quiet --host ${mongo_node} --eval "var done=false;while(done==false){var members=rs.status().members;for(i=0;i<members.length;i++){if(members[i].stateStr=='PRIMARY' && members[i].name!='$(mongo_addr)'){done=true}};sleep(500)};" &>/dev/null
```
rhcarvalho: `mongo_node` is a variable defined in `replset_supervisor.sh`, undefined here.

rhcarvalho: I see this is just moving a chunk of code from one file to another. In any case, wouldn't it be nice to split the line to make the oneliner javascript code easier to read / more visible?

3.2/root/usr/share/container-scripts/mongodb/replset_supervisor.sh
```diff
@@ -53,25 +55,12 @@ if [[ "$1" == "initiate" ]]; then
...
-  echo "=> A new PRIMARY member was elected, shutting down local mongod ..."
-  mongo_remove
-
-  mongod -f ${MONGODB_CONFIG_PATH} --shutdown &>/dev/null
-  wait_for_mongo_down
 
   echo "=> Successfully initialized replSet"
 
+  # Exit this pod
+  kill ${main_process_id}
```
rhcarvalho: Hmm, sorry I didn't follow. Why are we going to kill `run-mongod-replication`? How does this relate to https://bugzilla.redhat.com/show_bug.cgi?id=1260586?

(BTW, if this PR intends to fix that bug, please add a link in the PR's description)

rhcarvalho (commented):
@omron93 please clarify what this is trying to solve and how it solves it in the description.

There might be some future here, but I didn't understand yet how this solves the problem, thus asking for clarification.

3.2/root/usr/share/container-scripts/mongodb/replset_supervisor.sh
omron93: We kill run-mongod-replication to terminate post-hook pod after it initiates replset and creates users. After we kill this pod, `cleanup()` is processed and this pod removed from replset (in master it is removed in `replset_supervisor.sh`)

I was not aware of this bug when I was preparing this PR. I wanted to fix general situation when removing PRIMARY from replset.

And the bug is about this as I understand. In my testing this fixes it.

3.2/root/usr/bin/run-mongod-replication
omron93: I agree.

@rhcarvalho   Can I do it in this PR? :-)

omron93: Thanks. Fixed.

omron93:
> @omron93 please clarify what this is trying to solve and how it solves it in the description.

https://github.com/sclorg/mongodb-container/pull/186#discussion_r80470367 helps?

> There might be some future here, but I didn't understand yet how this solves the problem, thus asking for clarification.

By " mainly wanted to show possible future improvements." in PR description I wanted to make clear that this PR is not ideal.

For example 

```
echo "=> Waiting for replication to finish ..."
  # TODO: Replace this with polling or a Mongo script that will check if all
  #       members of the cluster are now properly replicated (user accounts are
  #       created on all members).
  sleep 10
```

should be also removed from `replset_supervisor.sh` and waiting in `cleanup()` should be updated to handle this. But I did not want to make this PR more complicated.

(also I wanted to show that after moving code from `replset_supervisor.sh` to `cleanup()` in this PR code of `replset_supervisor.sh` has a similar structure to `replset_petset_supervisor.sh` in #184 )

3.2/root/usr/bin/run-mongod-replication
rhcarvalho: Would be a welcome improvement!

rhcarvalho:
> #186 (comment) helps?

When people (think we in 6 months from now) look at a PR, we shouldn't require them to go through each comment and follow links to understand what's going on.

It is much better when the PR title gives a good short description of what it is about, and then the PR description explains the motivation, why we are doing and what we are doing. Also link there relevant information, such as related GitHub issues and bugzillas.

That's why I asked for clarification in the description, not in a comment or a link.

omron93:
> That's why I asked for clarification in the description, not in a comment or a link.

It is updated.

3.2/root/usr/bin/run-mongod-replication
omron93: Done. (bun only for code changed in this PR)

omron93:
Rebased and added function to wait that at least one SECONDARY is synced before PRIMARY removing.

@rhcarvalho @php-coder @bparees Any comments please?

php-coder:
@omron93 Don't wait my feedback, please. At this time I'm in an another team and working on a different area of the OpenShift, so I don't have time to watch for MongoDB and other images as I did before.

rhcarvalho:
@omron93, my situation is similar to @php-coder.

bparees:
[test]

bparees:
@omron93 lgtm.

omron93:
@php-coder @rhcarvalho Thanks for your work.

Anyone doing the same job instead of you?

bparees:
@omron93 we haven't identified a specific owner within the Developer Experience team for the mongo petset work that @php-coder and @rhcarvalho did, but to be honest I think the expectation at this point is for the SCL team to own this sort of thing.

omron93:
> but to be honest I think the expectation at this point is for the SCL team to own this sort of thing.

@bparees I understand. I mainly wanted to know whom to ask for OpenShift related code and code reviews?

3.2/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -130,6 +130,30 @@ function replset_addr() {
...
+  host="$(replset_addr || true)"
+  if [ -z "$host" ]; then
+    return 1
+  fi
+
+  mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --host ${host} \
+    --eval "var i = ${MAX_ATTEMPTS};
+    while(i > 0) {
+      var status=rs.status();
+      var primary_optime=rs.status().members.filter(function(el) {return el.state ==1})[0].optime;
```
hhorak: Any reason to not use `status` variable instead of calling `rs.status()` again?

hhorak: Calling `rs.status()` more times seems to me like opening gate to race condition, but I might miss something.

3.2/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -130,6 +130,30 @@ function replset_addr() {
...
+      if(rs.status().members.filter(function(el) {return el.optime.ts.tojson() == primary_optime.ts.tojson()}).length > 2)
```
hhorak: I read `...length > 2` like we need to find the master and one another who have the same `optime`, right? It would be better to add comment with short explanation, so reader does not need to figure out it..

3.2/root/usr/bin/run-mongod-replication
```diff
@@ -28,6 +28,32 @@ function usage() {
...
+    echo "=> Waiting for the new PRIMARY to be elected ..."
+    mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --quiet --eval "var done=false;
+      while(done==false){
+        var members=rs.status().members;
+        for(i=0;i<members.length;i++){
+          if(members[i].stateStr=='PRIMARY' && members[i].name!='$(mongo_addr)'){
+            done=true
+          }
+        };
+        sleep(500)
```
hhorak: This will make the process sleep 0.5s even after new primary is elected, I'd suggest to not sleep in this case any more, to avoid possible race condition which could potentially happen during that period (that the original master would be re-elected during that time). Many things might happen during 500ms in DB world, so I'd rather be super-paranoid :)

hhorak:
Just few formal comments in line, otherwise it all makes sense to me.

bparees:
> @bparees I understand. I mainly wanted to know whom to ask for OpenShift related code and code reviews?

for now just tag me, when i get overwhelmed i'll find another victim on the devex team :)

3.2/root/usr/share/container-scripts/mongodb/common.sh
omron93: @hhorak You are right. I overlooked this.

Fixed.

3.2/root/usr/share/container-scripts/mongodb/common.sh
omron93: There should be `length >= 2` (PRIMARY and at least on SECONDARY ~ `>= 2`)
Comment added.

3.2/root/usr/bin/run-mongod-replication
omron93: Right. Fixed.

Also removed infinite loop and added `MAX_ATTEMPTS` and `SLEEP_TIME`.

hhorak:
LGTM.

Issue #170, opened 2016-08-11
mongo 3.2 lacks most of the tools


jorgemoralespou:
To do initial loads into mongodb there is a need for mongoimport. As the package is no in rh-mongodb32-mongo-tools there is a need for:
- Have the tools in the mongo image
- Have an additional mongo-tools image that can be used to import data in the mongo db. This image can be used as sidecar, as post-deployment or as a job or runonce pod to load data into the mongo database.

hhorak:
> As the package is no in rh-mongodb32-mongo-tools there is a need for...
> I don't get it, I believe `mongoimport` is part of `rh-mongodb32-mongo-tools` package. But this package is not installed in the image, right?

Anyway, I see the point, but every solution has some side effects.

> Have the tools in the mongo image

The problem with mongo-tools is that the binaries are really big (statically-linked golang binaries with ~7MB each, so ~70MB in total).

Slightly less invasive but also less perfect would be to include only mongoimport in the image (remove unneeded binaries to safe image size; eventually by having something like mongo-tools-import sub-pkg)

> Have an additional mongo-tools image that can be used to import data in the mongo db. This image can be used as sidecar, as post-deployment or as a job or runonce pod to load data into the mongo database.

I have "some" idea about how complex this would be in k8s or openshift, but I'd like to see another opinions, whether this is the way we could go and whether it is worth doing for the image size decreasing. @bparees, any opinion?

bparees:
The sidecar would be fairly painful for users i think, they'd have to update the pod (or DC) to add the sidecar container, then ssh into it and use the tools, then update again to remove the sidecar container.

they're also likely to hit resource limits due to needing another container (especially in online) and having to redeploy my DB to add a tools container sounds very bad.  If i were an admin faced with that, i'd instead ssh into the DB container and curl down the tools i needed and use them that way.

jorgemoralespou:
@bparees why would the sidecar need to be removed and the
dbericat@redhat.com redeployed? Couldn't just be used for a post-deployment
hook?
Couldn't the template have both containers and just guarantee that the db
is running?
Or the container be used in a job to do the load/import?

bparees:
> @bparees why would the sidecar need to be removed and the dbericat@redhat.com redeployed? 

@jorgemoralespou  i was under the impression the admin wanted to use the command line tools by hand to do things.  In which case they need the sidecar container to rsh into, and they'd want to remove it afterwards so it is not consuming quota resources.

[...]

PR #187, opened 2016-09-23
Add mongo-tools for mongodb 3.2

omron93:
Install rh-mongodb32-mongo-tools in 3.2 image. Older mongodb versions already includes tools.

See #170 for discussion.


bparees:
[test]

PR #188, opened 2016-09-23
Minor improvements

php-coder:
- `run-mongod-replication(unset_env_vars)`: inline function
- `run-mongod(unset_env_vars)`: inline function
- `run-mongod-replication`: fail fast if `MONGODB_KEYFILE_VALUE` isn't defined

PTAL @bparees @omron93


omron93:
Thanks. Looks very good.

@bparees Please merge.

bparees:
squashed and merged.

PR #189, opened 2016-09-26
Allow passing arguments to mongo_add and mongo_remove

omron93:
This PR adds ability to pass arguments and behavior of `mongo_add()` and `mongo_remove()` functions. To allow to use them in more situations...

Also added `--quiet` of `mongo` command to not pollute output by mongo shell info (lines with shell version and where connecting).

@php-coder @rhcarvalho Please take a look.

### DIFF
--- a/2.4/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.4/root/usr/share/container-scripts/mongodb/common.sh
@@ -133,40 +133,42 @@ function replset_addr() {
   echo "${MONGODB_REPLICA_NAME}/${current_endpoints//[[:space:]]/,}"
 }
 
-# mongo_remove removes the current MongoDB from the cluster
+# mongo_remove by default removes the local mongod server from the replset
+# optionally takes identifier of replset to connect and address of server to remove
 function mongo_remove() {
   local host
   # if we cannot determine the IP address of the primary, exit without an error
   # to allow callers to proceed with their logic
-  host="$(replset_addr || true)"
+  host="${1:-$(replset_addr || true)}"
+  local mongo_addr
+  mongo_addr="${2:-$(mongo_addr)}"
+
   if [ -z "$host" ]; then
     return
   fi
 
-  local mongo_addr
-  mongo_addr="$(mongo_addr)"
-
   echo "=> Removing ${mongo_addr} from replica set ..."
   mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" \
-    --host "${host}" --eval "rs.remove('${mongo_addr}');" || true
+    --host "${host}" --quiet --eval "printjson(rs.remove('${mongo_addr}'));" || true
 }
 
-# mongo_add advertise the current container to other mongo replicas
+# mongo_add by default add the local mongod server to the replset
+# optionally takes identifier of replset to connect and address of server to add
 function mongo_add() {
   local host
   # if we cannot determine the IP address of the primary, exit without an error
   # to allow callers to proceed with their logic
-  host="$(replset_addr || true)"
+  host="${1:-$(replset_addr || true)}"
+  local mongo_addr
+  mongo_addr="${2:-$(mongo_addr)}"
+
   if [ -z "$host" ]; then
     return
   fi
 
-  local mongo_addr
-  mongo_addr="$(mongo_addr)"
-
   echo "=> Adding ${mongo_addr} to replica set ..."
   mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" \
-    --host "${host}" --eval "rs.add('${mongo_addr}');"
+    --host "${host}" --quiet --eval "printjson(rs.add('${mongo_addr}'));"
 }
 
 # mongo_create_admin creates the MongoDB admin user with password: MONGODB_ADMIN_PASSWORD
--- a/2.6/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/common.sh
@@ -132,40 +132,42 @@ function replset_addr() {
   echo "${MONGODB_REPLICA_NAME}/${current_endpoints//[[:space:]]/,}"
 }
 
-# mongo_remove removes the current MongoDB from the cluster
+# mongo_remove by default removes the local mongod server from the replset
+# optionally takes identifier of replset to connect and address of server to remove
 function mongo_remove() {
   local host
   # if we cannot determine the IP address of the primary, exit without an error
   # to allow callers to proceed with their logic
-  host="$(replset_addr || true)"
+  host="${1:-$(replset_addr || true)}"
+  local mongo_addr
+  mongo_addr="${2:-$(mongo_addr)}"
+
   if [ -z "$host" ]; then
     return
   fi
 
-  local mongo_addr
-  mongo_addr="$(mongo_addr)"
-
   echo "=> Removing ${mongo_addr} from replica set ..."
   mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" \
-    --host "${host}" --eval "rs.remove('${mongo_addr}');" || true
+    --host "${host}" --quiet --eval "printjson(rs.remove('${mongo_addr}'));" || true
 }
 
-# mongo_add advertise the current container to other mongo replicas
+# mongo_add by default add the local mongod server to the replset
+# optionally takes identifier of replset to connect and address of server to add
 function mongo_add() {
   local host
   # if we cannot determine the IP address of the primary, exit without an error
   # to allow callers to proceed with their logic
-  host="$(replset_addr || true)"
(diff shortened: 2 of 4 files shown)


2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -134,39 +134,41 @@ function replset_addr() {
...
 # mongo_remove removes the current MongoDB from the cluster
+# optionally takes address of replset to connect and address of server to remove
 function mongo_remove() {
   local host
   # if we cannot determine the IP address of the primary, exit without an error
   # to allow callers to proceed with their logic
-  host="$(replset_addr || true)"
+  host="${1:-$(replset_addr || true)}"
+  local mongo_addr
+  mongo_addr="${2:-(mongo_addr)}"
```
php-coder: Did you forgot `$` here? `${2:-(mongo_addr)}` vs `${2:-$(mongo_addr)}`

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -134,39 +134,41 @@ function replset_addr() {
...
     return
   fi
 
-  local mongo_addr
-  mongo_addr="$(mongo_addr)"
-
   echo "=> Adding ${mongo_addr} to replica set ..."
   mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" \
-    --host "${host}" --eval "rs.add('${mongo_addr}');"
+    --host "${host}" --quiet --eval "rs.add('${mongo_addr}');"
```
php-coder: Can you add `printjson()` here? Like this: `--eval "rs.add(printjson('${mongo_addr}'));"` Because otherwise I sill won't be able to use this function in my PR.

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: Thanks.  Fixed.

rhcarvalho:
UNIX philosophy applied to Bash scripting, I would rather have a function do less and do it well, than to do more and have several flags and complicated usage.

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -134,39 +134,41 @@ function replset_addr() {
 }
 
 # mongo_remove removes the current MongoDB from the cluster
+# optionally takes address of replset to connect and address of server to remove
 function mongo_remove() {
   local host
   # if we cannot determine the IP address of the primary, exit without an error
   # to allow callers to proceed with their logic
```
rhcarvalho, line 141: I think this comment (and another instance of it in `mongo_add`) is outdated since `primary_addr` was changed into `replset_addr`...

rhcarvalho:
This PR by itself feels like a change just for the sake of making code more complex.

Instead of "To allow to use them in more situations..." as a motivation, it would be good to have one such "situation" driving the change, and then we could discuss the implementation / change.

omron93:
> UNIX philosophy applied to Bash scripting, I would rather have a function do less and do it well, than to do more and have several flags and complicated usage.

Yes. This function adds/removes MEMBER to REPLSET.
I don't think it too complicated to be able to specify the replset and what to add there. Or should be there several almost same functions for every possible combination of repset identifiers and what we would like to add.
Almost every UNIX simple program (cat, echo,...) takes some input - is parameterized. And still comply to this simple rule. 

> Instead of "To allow to use them in more situations..." as a motivation, it would be good to have one such "situation" driving the change, and then we could discuss the implementation / change.

The "situation" is #184 . 

Imagine situation that upstream change functions to manipulate with replset (in recent versions they changed function to user manipulation). IMPOV lesser/single place to update the better.

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: OK. Substituted address to identifier.

IMPOV, "address of replset" is not "address of primary member of replset". It some identifier how to connect to replset.

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: @php-coder Fixed. I hope it should be `printjson(rs.add('${mongo_addr}'))`?

rhcarvalho:
If this is necessary in #184, let's make the move there.

2.4/root/usr/share/container-scripts/mongodb/common.sh
php-coder: @omron93 Ah, yes, of course! Thank you!

php-coder:
LGTM. Even if it will break something, we catch and fix it shortly.

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -134,39 +134,41 @@ function replset_addr() {
 }
 
 # mongo_remove removes the current MongoDB from the cluster
+# optionally takes identifier of replset to connect and address of server to remove
```
rhcarvalho, line 137: This says "removes the current MongoDB from the cluster". That is inaccurate.
If we are to make the function generic, let's revisit the wording here please.

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -134,39 +134,41 @@ function replset_addr() {
...
 function mongo_remove() {
   local host
   # if we cannot determine the IP address of the primary, exit without an error
   # to allow callers to proceed with their logic
-  host="$(replset_addr || true)"
+  host="${1:-$(replset_addr || true)}"
+  local mongo_addr
+  mongo_addr="${2:-$(mongo_addr)}"
```
rhcarvalho, line 144: Different than flags to UNIX tools, arguments here are not so convenient to use.
E.g., it is not easy to provide what you want to remove ($2) without providing the replica set address ($1).

Undocumented behavior: is it valid to call `mongo_remove` to remove something that is not an item in the replica set members?

When do we want to specify the host to be something that is not `replset_addr`?

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: I hope comments are more clear. @rhcarvalho  Please check.

2.4/root/usr/share/container-scripts/mongodb/common.sh
omron93: > E.g., it is not easy to provide what you want to remove ($2) without providing the replica set address ($1).

@rhcarvalho You can use `""` as a first argument. I know that this is not elegant, but it is possible.

> Undocumented behavior: is it valid to call mongo_remove to remove something that is not an item in the replica set members?

Yes, IMPOV it is valid but should return the appropriate error.

> When do we want to specify the host to be something that is not replset_addr?

#184 uses only hostnames to addressing.

rhcarvalho: > You can use "" as a first argument. I know that this is not elegant, but it is possible.

In this case it would have a different behavior, it would set host to the empty string (and not call `replset_addr`)...

> > When do we want to specify the host to be something that is not replset_addr?
> 
> #184 uses only hostnames to addressing.

And you were arguing there that we should use the replica set addressing :-)
And I agree with you!

That's why I would defer this to be worked on #184, not as a separate PR since it has no clear benefits without that other one (which is work in progress...).

omron93: > In this case it would have a different behavior, it would set host to the empty string (and not call replset_addr)...

@rhcarvalho Really? `:-` is "If parameter is unset or null"

> And you were arguing there that we should use the replica set addressing :-)
> And I agree with you!

Yes i think we should. I am trying to convice @php-coder. But now it is better for me to have more generic function than do not use this function.

> That's why I would defer this to be worked on #184, not as a separate PR since it has no clear benefits without that other one (which is work in progress...).

I agree.
Too late? :-)

rhcarvalho: > Really? :- is "If parameter is unset or null"

Dear God, I should tatoo the contents of `man bash` in my forearms...

php-coder:
@rhcarvalho @bparees What is blocking the merge?

bparees:
@php-coder i don't have time to look at this right now.  If you and @rhcarvalho are in agreement it's ok, i will merge it.

rhcarvalho:
@bparees,

@omron93 and I [agreed](https://github.com/sclorg/mongodb-container/pull/189#discussion_r80481766) this can be worked on #184, if @php-coder will need this code.

Otherwise this PR is "coding for the future".

omron93:
> @omron93 and I agreed this can be worked on #184, if @php-coder will need this code.

However not sure if this is necessary prolonging.

PR #190, opened 2016-09-29
Improve error reporting

php-coder:
- show detailed error message when MongoDB home directory isn't writable.
  Before:

``` console
$ oc logs po/mongodb-0
/usr/share/container-scripts/mongodb/common.sh: line 250: /var/lib/mongodb/keyfile: Permission denied
```

After:

``` console
$ oc logs po/mongodb-0
ERROR: Couldn't create /var/lib/mongodb/keyfile
CAUSE: current user doesn't have permissions for writing to /var/lib/mongodb directory
DETAILS: current user id = 1000040000, user groups: 0 1000040000
DETAILS: directory permissions: drwxrwxr-x owned by 1000:1000
```
- print errors into stderr

PTAL @bparees @omron93 @rhcarvalho

### DIFF
--- a/2.4/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.4/root/usr/share/container-scripts/mongodb/common.sh
@@ -44,7 +44,7 @@ function cache_container_addr() {
     fi
     sleep $SLEEP_TIME
   done
-  echo "Failed to get Docker container IP address." && exit 1
+  echo >&2 "Failed to get Docker container IP address." && exit 1
 }
 
 # wait_for_mongo_up waits until the mongo server accepts incomming connections
@@ -174,14 +174,14 @@ function mongo_add() {
 # $2 - host where to connect (localhost by default)
 function mongo_create_admin() {
   if [[ -z "${MONGODB_ADMIN_PASSWORD:-}" ]]; then
-    echo "=> MONGODB_ADMIN_PASSWORD is not set. Authentication can not be set up."
+    echo >&2 "=> MONGODB_ADMIN_PASSWORD is not set. Authentication can not be set up."
     exit 1
   fi
 
   # Set admin password
   local js_command="db.addUser({user: 'admin', pwd: '${MONGODB_ADMIN_PASSWORD}', roles: ['dbAdminAnyDatabase', 'userAdminAnyDatabase' , 'readWriteAnyDatabase','clusterAdmin' ]});"
   if ! mongo admin ${1:-} --host ${2:-"localhost"} --eval "${js_command}"; then
-    echo "=> Failed to create MongoDB admin user."
+    echo >&2 "=> Failed to create MongoDB admin user."
     exit 1
   fi
 }
@@ -193,22 +193,22 @@ function mongo_create_admin() {
 function mongo_create_user() {
   # Ensure input variables exists
   if [[ -z "${MONGODB_USER:-}" ]]; then
-    echo "=> MONGODB_USER is not set. Failed to create MongoDB user"
+    echo >&2 "=> MONGODB_USER is not set. Failed to create MongoDB user"
     exit 1
   fi
   if [[ -z "${MONGODB_PASSWORD:-}" ]]; then
-    echo "=> MONGODB_PASSWORD is not set. Failed to create MongoDB user: ${MONGODB_USER}"
+    echo >&2 "=> MONGODB_PASSWORD is not set. Failed to create MongoDB user: ${MONGODB_USER}"
     exit 1
   fi
   if [[ -z "${MONGODB_DATABASE:-}" ]]; then
-    echo "=> MONGODB_DATABASE is not set. Failed to create MongoDB user: ${MONGODB_USER}"
+    echo >&2 "=> MONGODB_DATABASE is not set. Failed to create MongoDB user: ${MONGODB_USER}"
     exit 1
   fi
 
   # Create database user
   local js_command="db.getSiblingDB('${MONGODB_DATABASE}').addUser({user: '${MONGODB_USER}', pwd: '${MONGODB_PASSWORD}', roles: [ 'readWrite' ]});"
   if ! mongo admin ${1:-} --host ${2:-"localhost"} --eval "${js_command}"; then
-    echo "=> Failed to create MongoDB user: ${MONGODB_USER}"
+    echo >&2 "=> Failed to create MongoDB user: ${MONGODB_USER}"
     exit 1
   fi
 }
@@ -218,7 +218,7 @@ function mongo_reset_user() {
   if [[ -n "${MONGODB_USER:-}" && -n "${MONGODB_PASSWORD:-}" && -n "${MONGODB_DATABASE:-}" ]]; then
     local js_command="db.changeUserPassword('${MONGODB_USER}', '${MONGODB_PASSWORD}')"
     if ! mongo ${MONGODB_DATABASE} --eval "${js_command}"; then
-      echo "=> Failed to reset password of MongoDB user: ${MONGODB_USER}"
+      echo >&2 "=> Failed to reset password of MongoDB user: ${MONGODB_USER}"
       exit 1
     fi
   fi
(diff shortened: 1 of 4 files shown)


2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -244,7 +244,15 @@ function setup_keyfile() {
...
-    echo "ERROR: You have to provide the 'keyfile' value in MONGODB_KEYFILE_VALUE"
+    echo >&2 "ERROR: You have to provide the 'keyfile' value in MONGODB_KEYFILE_VALUE"
+    exit 1
+  fi
+  MONGODB_KEYFILE_DIR="$(dirname "$MONGODB_KEYFILE_PATH")"
+  if [ ! -w "$MONGODB_KEYFILE_DIR" ]; then
+    echo >&2 "ERROR: Couldn't create ${MONGODB_KEYFILE_PATH}"
+    echo >&2 "CAUSE: current user doesn't have permissions for writing to ${MONGODB_KEYFILE_DIR} directory"
+    echo >&2 "DETAILS: current user id = $(id -u), user groups: $(id -G)"
+    echo >&2 "DETAILS: directory permissions: $(stat -c '%A owned by %u:%g' "${MONGODB_KEYFILE_DIR}")"
```
rhcarvalho: Are there other places where we write "CAUSE" and "DETAILS"?

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -44,7 +44,7 @@ function cache_container_addr() {
     fi
     sleep $SLEEP_TIME
   done
-  echo "Failed to get Docker container IP address." && exit 1
+  echo >&2 "Failed to get Docker container IP address." && exit 1
```
rhcarvalho, line 47: :+1:  to writing to stderr

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -244,7 +244,15 @@ function setup_keyfile() {
     exit 0
   fi
   if [ -z "${MONGODB_KEYFILE_VALUE-}" ]; then
-    echo "ERROR: You have to provide the 'keyfile' value in MONGODB_KEYFILE_VALUE"
+    echo >&2 "ERROR: You have to provide the 'keyfile' value in MONGODB_KEYFILE_VALUE"
+    exit 1
+  fi
+  MONGODB_KEYFILE_DIR="$(dirname "$MONGODB_KEYFILE_PATH")"
```
rhcarvalho: Maybe make this a local variable?

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -244,7 +244,15 @@ function setup_keyfile() {
...
+  if [ ! -w "$MONGODB_KEYFILE_DIR" ]; then
+    echo >&2 "ERROR: Couldn't create ${MONGODB_KEYFILE_PATH}"
+    echo >&2 "CAUSE: current user doesn't have permissions for writing to ${MONGODB_KEYFILE_DIR} directory"
```
rhcarvalho: How about suggesting how to fix the problem or pointing to docs?

rhcarvalho (commented):
May be good for detecting / debugging ill configured PV. I have a question about the format and suggestions.

2.4/root/usr/share/container-scripts/mongodb/common.sh
php-coder: Oh, indeed. I've just pushed updated version.

2.4/root/usr/share/container-scripts/mongodb/common.sh
php-coder: No, as far I know. But I'm using "CAUSE" in #184

2.4/root/usr/share/container-scripts/mongodb/common.sh
php-coder: There are many causes that can lead to this error and I don't think that it's possible to cover all of them. For example, SELinux problems not covered here (actually, I tried to use `id -Z` and `stat -c %C` but `id -Z` shows that there is no selinux context inside of container).

omron93:
@php-coder Nice improvements, thanks.

I suggest add a similar test as for keyfile dir also for MONGODB_DATADIR and run this test somewhere in `run-mongod` or `run-mongod-replication`. To fail clearly if mounted data directory is not writable... WDYT?

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -244,7 +244,16 @@ function setup_keyfile() {
     exit 0
   fi
   if [ -z "${MONGODB_KEYFILE_VALUE-}" ]; then
-    echo "ERROR: You have to provide the 'keyfile' value in MONGODB_KEYFILE_VALUE"
+    echo >&2 "ERROR: You have to provide the 'keyfile' value in MONGODB_KEYFILE_VALUE"
+    exit 1
+  fi
+  local keyfile_dir
+  keyfile_dir="$(dirname "$MONGODB_KEYFILE_PATH")"
+  if [ ! -w "$keyfile_dir" ]; then
```
omron93, line 252: When this could happen? When is HOME directory mounted?

php-coder: Yes, exactly. I faced it when I added PV.

php-coder:
> I suggest add a similar test as for keyfile dir also for MONGODB_DATADIR and run this test somewhere in run-mongod or run-mongod-replication. To fail clearly if mounted data directory is not writable... WDYT?

@omron93 It make sense if `MONGODB_DATADIR` and `$HOME` is different but right now they point to the same directory. I don't think that we need this additional check now. But I wouldn't mind if you create separate PR add add it, if you think different.

rhcarvalho:
@php-coder I understood the suggestion is to check just $MONGODB_DATADIR (not $HOME, even if they are the same...).
Sounds like a good idea, to fail fast instead of in the middle of the database operation.

php-coder:
@rhcarvalho I didn't get the idea: are you suggesting to change the definition of `MONGODB_KEYFILE_PATH` to use `MONGODB_DATADIR` instead of `HOME`?

rhcarvalho:
@php-coder the suggestion from @omron93  was to in addition to checking $MONGODB_KEYFILE_PATH, also check $MONGODB_DATADIR (in a different place).

2.4/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -244,7 +244,16 @@ function setup_keyfile() {
     exit 0
```
php-coder, line 244: @omron93 The `exit 0` here looks very suspicious for me. As far I understand it should be replaced by `return`, right? Because if we execute `exit` inside of the function it will exit the script and other code won't be executed.

php-coder:
@rhcarvalho Ok, now it's clear. I'm not going to add this change to the current PR.

php-coder:
@bparees @omron93 @rhcarvalho Do you have any objections against merging this?

2.4/root/usr/share/container-scripts/mongodb/common.sh
rhcarvalho: Not only here, looking at `common.sh` there are many occurrences of `exit`... feels puzzling.

php-coder: Yes, but others are `exit 1` that's signalized about fatal errors while this case isn't an error.

rhcarvalho: Exiting really depends on the caller... if the call is in a subshell, it will exit the subshell.
In the case of setup_keyfile, I see it is called in the top level, I totally agree this `exit 0` is suspicious.

And so are the `exit 1`. They should probably be `return 1`, and exiting would be caused by the `errexit` shell option that we have set.

PR #191, opened 2016-10-04
Kill mongod processes ignoring errors

rhcarvalho:
Problems with previous approach:
- pgrep might find processes, but there is no guarantee pkill will not
  fail for any other reason (e.g. there is a race condition in which
  mongod is terminating, pgrep finds a PID, but pkill doesn't). If pkill
  returns non-zero exit code, the rest of the cleanup function is not
  executed.
- could be fixed in a different way, but for now pgrep was
  unnecessarily spitting out PIDs to stdout.

### DIFF
--- a/2.4/root/usr/bin/run-mongod
+++ b/2.4/root/usr/bin/run-mongod
@@ -21,9 +21,7 @@ function usage() {
 
 function cleanup() {
   echo "=> Shutting down MongoDB server ..."
-  if pgrep mongod; then
-    pkill -2 mongod
-  fi
+  pkill -INT mongod || :
   wait_for_mongo_down
   exit 0
 }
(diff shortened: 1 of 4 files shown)


rhcarvalho:
@php-coder @omron93 PTAL

php-coder:
LGTM.

rhcarvalho:
just curious, @bparees are we merging PRs without merge commits by design or accident?
noticed that when doing my branch cleanup steps...

bparees:
@rhcarvalho something was off about this PR, despite it being only a single commit, git was asking me to "squash+merge" instead of just merge.  I didn't feel like investigating so i just went with it.  Perhaps that's the cause of whatever you're seeing?

rhcarvalho:
@bparees it's probably the GitHub UI trying to influence us then.
This is what I observed:
https://github.com/sclorg/mongodb-container/commits/master

The last 3 commits/PRs went straight into master (with different commit hashes than in the PR), unlike what we have done more often with commit(s) + merge commit.

Branch cleanup-wise, with the squash merge, git won't list merged branches with `git branch --merged`... I often use that command to find branches that I can get rid of.

PR #193, opened 2016-10-06
Optimize Dockerfile for cache usage

rhcarvalho:
Saves lots of developer time running `make` / `docker build`, by
reordering instructions to better use the cache when we make changes to
scripts, and also reduce number of layers leading to a smaller time to
build.

For instance, a developer might run the command below multiple times in
a row while making changes to the scripts that are copied into the
image:

```
make VERSION=3.2 SKIP_SQUASH=1
```

### DIFF
--- a/3.2/Dockerfile
+++ b/3.2/Dockerfile
@@ -11,15 +11,27 @@ MAINTAINER SoftwareCollections.org <sclorg@redhat.com>
 #  * $MONGODB_DATABASE - Name of the database to create
 #  * $MONGODB_ADMIN_PASSWORD - Password of the MongoDB Admin
 
-ENV MONGODB_VERSION=3.2 \
-    HOME=/var/lib/mongodb
-
 LABEL io.k8s.description="MongoDB is a scalable, high-performance, open source NoSQL database." \
       io.k8s.display-name="MongoDB 3.2" \
       io.openshift.expose-services="27017:mongodb" \
       io.openshift.tags="database,mongodb,rh-mongodb32"
 
+ENV MONGODB_VERSION=3.2 \
+    # Set paths to avoid hard-coding them in scripts.
+    HOME=/var/lib/mongodb \
+    CONTAINER_SCRIPTS_PATH=/usr/share/container-scripts/mongodb \
+    MONGODB_PREFIX=/opt/rh/rh-mongodb32/root/usr \
+    # Incantations to enable Software Collections on `bash` and `sh -i`.
+    ENABLED_COLLECTIONS=rh-mongodb32 \
+    BASH_ENV="\${CONTAINER_SCRIPTS_PATH}/scl_enable" \
+    ENV="\${CONTAINER_SCRIPTS_PATH}/scl_enable" \
+    PROMPT_COMMAND=". \${CONTAINER_SCRIPTS_PATH}/scl_enable"
+
 EXPOSE 27017
+
+
+ENTRYPOINT ["container-entrypoint"]
+CMD ["run-mongod"]
 
 # Due to the https://bugzilla.redhat.com/show_bug.cgi?id=1206151,
 # the whole /var/lib/mongodb/ dir has to be chown-ed.
@@ -33,26 +45,11 @@ RUN yum install -y centos-release-scl && \
     chmod g+w -R /var/opt/rh/rh-mongodb32/lib/mongodb && \
     chmod -R g+rwx /var/lib/mongodb
 
-# Get prefix path and path to scripts rather than hard-code them in scripts
-ENV CONTAINER_SCRIPTS_PATH=/usr/share/container-scripts/mongodb \
-    MONGODB_PREFIX=/opt/rh/rh-mongodb32/root/usr \
-    ENABLED_COLLECTIONS=rh-mongodb32
-
-# When bash is started non-interactively, to run a shell script, for example it
-# looks for this variable and source the content of this file. This will enable
-# the SCL for all scripts without need to do 'scl enable'.
-ENV BASH_ENV=${CONTAINER_SCRIPTS_PATH}/scl_enable \
-    ENV=${CONTAINER_SCRIPTS_PATH}/scl_enable \
-    PROMPT_COMMAND=". ${CONTAINER_SCRIPTS_PATH}/scl_enable"
+VOLUME ["/var/lib/mongodb/data"]
 
 ADD root /
 
 # Container setup
 RUN touch /etc/mongod.conf && chown mongodb:0 /etc/mongod.conf && /usr/libexec/fix-permissions /etc/mongod.conf
 
-VOLUME ["/var/lib/mongodb/data"]
-
 USER 184
-
-ENTRYPOINT ["container-entrypoint"]
-CMD ["run-mongod"]
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,43 @@
 
 This guide aims to help developers working on this repository.
 
+## Making changes to runtime scripts
+
+The Docker images created from this repository contain several runtime scripts
+that are executed from the image entrypoint / default command. These scripts are
+located under `${IMAGE_VERSION}/root/`.
+
+As a suggestion, start making changes to files of a single image version, e.g.,
+the latest one, and then run tests for that image:
+
+```bash
+make test VERSION=3.2 SKIP_SQUASH=1
+```
+
+By default, the last step in the image build process is to squash the output
+image into less layers, optimizing image size for distribution. However, unless
+specifically testing that this squash process is working, you don't need to
+squash images in development, and can save some precious time with consecutive
+rebuilds as you work in the image contents.
+
+The `SKIP_SQUASH=1` option disables image squashing, so that building and
+rebuilding images locally should take less time.
+
+Setting `VERSION=3.2` equally saves development time. Instead of building every
+image version, only the version you are working on will be built.
+
+The tests run via `make test` cover some basic usage patterns of the images.
+Often times, some manual testing is required while making changes to the scripts
+that are copied into the images. For instance, you can build and rebuild a
+specific image (without running tests) with this command:
+
+```bash
+make VERSION=3.2 SKIP_SQUASH=1
+```
+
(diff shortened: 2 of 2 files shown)


rhcarvalho:
Tired of slow builds while working on #184, I've made these changes to the Dockerfile.

@hhorak @omron93 @bparees @php-coder PTAL

If we have agreement on the changes for the 3.2 Dockerfile I'll apply similar changes the other Dockerfiles before merge.

php-coder:
LGTM.

3.2/Dockerfile
```diff
@@ -33,26 +30,25 @@ RUN yum install -y centos-release-scl && \
     chmod g+w -R /var/opt/rh/rh-mongodb32/lib/mongodb && \
     chmod -R g+rwx /var/lib/mongodb
 
-# Get prefix path and path to scripts rather than hard-code them in scripts
-ENV CONTAINER_SCRIPTS_PATH=/usr/share/container-scripts/mongodb \
+ENV MONGODB_VERSION=3.2 \
+    HOME=/var/lib/mongodb \
+    # Set paths to avoid hard-coding them in scripts.
+    CONTAINER_SCRIPTS_PATH=/usr/share/container-scripts/mongodb \
     MONGODB_PREFIX=/opt/rh/rh-mongodb32/root/usr \
```
rhcarvalho: I was tempted to remove this variable that is not referenced anywhere. If somebody can confirm it has no use, we can get rid of it.

3.2/Dockerfile
```diff
@@ -11,9 +11,6 @@ MAINTAINER SoftwareCollections.org <sclorg@redhat.com>
 #  * $MONGODB_DATABASE - Name of the database to create
 #  * $MONGODB_ADMIN_PASSWORD - Password of the MongoDB Admin
```
rhcarvalho, line 12: Observation: these variables are referenced in the scripts, not in the Dockerfile, and this list is incomplete. Perhaps this information fits better in the README file? It's already there anyway.

omron93: These comments are there from first commit. Personally I agree with you, so if there is no reason from OpenShift why there were added... it could be removed.

To note: other images contains similar comments, so it could be good to be consistent.

3.2/Dockerfile
omron93: @rhcarvalho remove it...

3.2/Dockerfile
```diff
@@ -33,26 +30,25 @@ RUN yum install -y centos-release-scl && \
...
-    ENABLED_COLLECTIONS=rh-mongodb32
+    # Incantations to enable Software Collections on `bash` and `sh -i`.
+    ENABLED_COLLECTIONS=rh-mongodb32 \
+    BASH_ENV="\${CONTAINER_SCRIPTS_PATH}/scl_enable" \
+    ENV="\${CONTAINER_SCRIPTS_PATH}/scl_enable" \
+    PROMPT_COMMAND=". \${CONTAINER_SCRIPTS_PATH}/scl_enable"
 
-# When bash is started non-interactively, to run a shell script, for example it
-# looks for this variable and source the content of this file. This will enable
-# the SCL for all scripts without need to do 'scl enable'.
```
omron93, line 43: IMPOV These comments are useful - normally I would not be sure what these env. vars do...

omron93:
@rhcarvalho Why not to move `ENV` before `RUN`? (it would same build time even more)

3.2/Dockerfile
rhcarvalho: I didn't "remove" them, just made it less verbose and added a IMPOV important information: it only works for `bash` and `sh -i`. We have hit hard walls forgetting about the `-i` some times...

See https://github.com/sclorg/s2i-python-container/pull/108 for a PoC that we could remove this requirement to wrap everything on a shell.

rhcarvalho:
> Why not to move ENV before RUN? (it would same build time even more)

Yes, it wouldn't change anything for now.

Because the first RUN instruction doesn't depend on any of the variables set in the ENV instruction, I put the ENV after RUN so that we can add/remove/edit variables without invalidating the cache for the `yum install`, which is definitely the one that benefits the most from a cache hit.

omron93:
> Because the first RUN instruction doesn't depend on any of the variables set in the ENV instruction, I put the ENV after RUN so that we can add/remove/edit variables without invalidating the cache for the yum install, which is definitely the one that benefits the most from a cache hit.

It depends what we think that changes more often ;-)

rhcarvalho:
> It depends what we think that changes more often ;-)

Yes, I agree, that's not science.

They are probably similarly (un)likely to change. Since they don't depend on each other, my logic was to put the most "expensive" instruction first, as I explained.

omron93:
> my logic was to put the most "expensive" instruction first

You are right. LGTM

3.2/Dockerfile
```diff
@@ -33,26 +30,25 @@ RUN yum install -y centos-release-scl && \
     chmod g+w -R /var/opt/rh/rh-mongodb32/lib/mongodb && \
     chmod -R g+rwx /var/lib/mongodb
 
-# Get prefix path and path to scripts rather than hard-code them in scripts
-ENV CONTAINER_SCRIPTS_PATH=/usr/share/container-scripts/mongodb \
+ENV MONGODB_VERSION=3.2 \
+    HOME=/var/lib/mongodb \
+    # Set paths to avoid hard-coding them in scripts.
+    CONTAINER_SCRIPTS_PATH=/usr/share/container-scripts/mongodb \
```
bparees: why not put all the ENV definitions first, instead of after the RUN, if you want to optimize cache reuse?  it also puts the "MONGODB_VERSION" variable front and center so it doesn't get overlooked during updates.

bparees (changes requested):

3.2/Dockerfile
rhcarvalho: https://github.com/sclorg/mongodb-container/pull/193#issuecomment-251918023

bparees: we generally put ENV front and center in the dockerfiles because it contains critical information people go looking for when trying to understand the env defined for an image.  I view it like comments.  I think they're better suited to be lumped w/ the labels.

and i disagree that these are more likely to change than the packages you're installing.  the env variables are almost never going to change.  the packages will, next time someone asks us to add another common mongo utility.  (or we need to chmod something else, or add some other step to the RUN block)

rhcarvalho:
@bparees PTAL

rhcarvalho:
Once we get to an agreement with the 3.2 Dockerfile, I still need to change the other image versions and RHEL.

CONTRIBUTING.md
```diff
@@ -2,6 +2,43 @@
...
+Often times, some manual testing is required while making changes to the scripts
+that are copied into the images. For instance, you can build and rebuild a
+specific image (without running tests) with this command:
+
+```bash
+make VERSION=3.2 SKIP_SQUASH=1
+```
+
+When you are done with your changes, you probably want to [change all other
+images without copy-paste](#changing-all-images-without-copy-paste).
```
bparees: s/images/versions/

bparees (changes requested):
one minor nit and lgtm.

CONTRIBUTING.md
rhcarvalho: updated

3.2/Dockerfile
```diff
@@ -1,28 +1,29 @@
+# MongoDB image for OpenShift.
```
omron93: @bparees It is still true? Does PM still reference these images as OpenShift or RHSCL?

omron93:
(note: ci.centos.org is down for 48 hours, so test won't end before 2016-10-12 13:00 UTC)

3.2/Dockerfile
bparees: i think both, but you're right that we can probably remove the openshift reference at this point since this is intended to be a general rhscl docker image, not specific to openshift (though i'm guessing 95% of the time it's being used on openshift)

omron93:
@rhcarvalho RHEL7 tests are failing because in result image 

```
bash-4.2$ ls -l /var/lib/mongodb/ -a
total 4
drwxrwxr-x.  3 mongodb root   17 Oct 10 09:37 .
drwxr-xr-x. 14 root    root 4096 Oct 10 09:38 ..
drwxr-xr-x.  2 root    root    6 Oct 10 13:40 data
```

rhcarvalho:
> RHEL7 tests are failing because in result image

There are no changes to the RHEL Dockerfile so far ;)

Doesn't that mean that tests do not pass with current master?

rhcarvalho:
@omron93 @hhorak could you help me understand why the RHEL7 check is failing? There are no changes to the RHEL Dockerfiles (yet). I want to see it green before pushing changes to the RHEL images to match the 3.2 CentOS.

3.2/Dockerfile
rhcarvalho: Removing it.

omron93:
> could you help me understand why the RHEL7 check is failing? There are no changes to the RHEL Dockerfiles (yet). I want to see it green before pushing changes to the RHEL images to match the 3.2 CentOS.

@rhcarvalho There was an problem in RHEL7 testing - it tested centos version (only in RHEL environment). And in this PR it failed.

For me, local centos testing also failed.

omron93:
From mongod log

```
2016-10-11T07:01:30.925+0000 I STORAGE  [initandlisten] exception in initAndListen: 98 Unable to create/open lock file: /var/lib/mongodb/data/mongod.lock errno:13 Permission denied Is a mongod instance already running?, terminating
```

see https://github.com/sclorg/mongodb-container/pull/193#issuecomment-252622123

rhcarvalho:
@omron93 how to re-run the tests? I want to see the RHEL7 check green, to avoid wasting time trying to fix the test execution later.

rhcarvalho:
Alright, [`VOLUME`](https://docs.docker.com/engine/reference/builder/#/volume) needs to come after the `RUN` instruction that creates directories and changes permissions. Updated.

bparees:
@rhcarvalho sorry, merged this prematurely.

PR #195, opened 2016-10-07
Document how to apply changes without copy-paste

rhcarvalho:
This is the beginning of a contribution guide.

Overtime the collaborators of this repo have developed their own ways around improving the repository, so to help share ideas and solutions to common problems, I started with documenting how I normally apply changes avoiding copy-paste.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,45 @@
+# Contribution guide
+
+This guide aims to help developers working on this repository.
+
+## Changing all images without copy-paste
+
+We keep the multiple files for the different image versions in sync. Here is a
+tip on how to use `git` commands to automatically apply patches targeting one
+image to all the others, without incurring in manually editing each copy of each
+modified file.
+
+Supposing that changes were made to files in the 3.2 directory, we can then
+apply those changes to all other images:
+
+```bash
+for version in 2.4 2.6 3.0-upg; do
+  git diff -- 3.2 | git apply -p2 --directory ${version} --reject
+done
+```
+
+Depending on the changes and the surrounding context, the patch may not apply
+cleanly. In that case, Git will create `*.rej` files with the changes that could
+not be applied.
+
+You can show them all with:
+
+```bash
+find -name '*.rej' -exec cat {} \;
+```
+
+Fix the differences manually, then delete the `*.rej` files.
+
+Notes:
+
+- Sometimes it may be useful to ignore part of the context passing the `-C<n>`
+  flag to `git apply`.
+- Read `git help diff` and `git help apply` for details about the flags used
+  here and further insights on how to combine them.
+
+**Review the changes** before committing to make sure the patch to all versions
+make sense.
+
+Sometimes it is also useful to use a graphical diff tool such as
+[Meld](http://meldmerge.org/) for a final verification, or working side-by-side
+on the parts that actually differ and that cannot the copied over automatically.


rhcarvalho:
@omron93 @php-coder @bparees PTAL

CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
...
+We keep the multiple files for the different image versions in sync. Here is a
+tip on how to use `git` and `sed` to automatically apply patches targeting one
+image to all the others, without incurring in manual copy-paste.
+
+Supposing that changes were made to files in the 3.2 directory, we can then
+apply those changes to all other images:
+
+```bash
+for version in 2.4 2.6 3.0-upg; do
+  git diff -- 3.2 | sed "s|3.2/|${version}/|g" | git apply
```
php-coder: Just an idea: you can replace `git apply` by `patch -p1 --dry-run` to check whether your changes will be applied without conflicts.

php-coder:
Maybe we should talk a bit about Bash code style that we're using?

praiskup:
There is `--check` and `--reject` in git-apply.

rhcarvalho:
> Maybe we should talk a bit about Bash code style that we're using?

Yes, I intentionally left that out for a future topic/PR, one thing at a time.

I wanted to start with sharing just this snippet that gets used by myself with for almost every change, and possibly learn from other collaborators better ways to do it.

CONTRIBUTING.md
rhcarvalho: If it doesn't apply I get an error from `git apply` and I'm fine with that.

How does running `patch -p1 --dry-run` and `git-apply --check` improve this work flow?
@php-coder @praiskup do you actually run those commands with those flags?

I see `--reject` could be useful to partially apply patches.

php-coder: Yep, I used it some time ago.

php-coder:
> Yes, I intentionally left that out for a future topic/PR, one thing at a time.

Ok. So, it's a good start then!

CONTRIBUTING.md
praiskup: I use `--reject` a lot.  The point is that I get the `*.rej` files, and I open them by:
`vim affected/file.suffix{,.rej}` (or so..), `vsplit`, `b2`, ...

Anyway, I don't consider this is "no-cut-and-paste" approach.

Personally, I don't consider this auto-C&P much better than what we have now, because that is several seconds to do manually anyway.  Much worse is to review the code later in web-ui.
So I allways fall-back to cli and do `some-smart-diff-tool versionA versionB`.

rhcarvalho:
@php-coder @praiskup I've updated based on your feedback, thank you!

I've mentioned some more alternatives for when the patch doesn't apply, and suggested using Meld for side-by-side editing and diff'ing. Of course people are free to use whatever they feel like, including copy-paste.

I just want to give an advice instead of leaving contributors in the dark.

praiskup:
It would be fine if these instruction could be shared (at least) by C&Ping it elsewhere.
Also consider:

```
$ git diff diff [--cached]  | git apply -p2 --directory VERSION --reject
$ find -name '*.rej'
... Dockerfile.rej
```

rhcarvalho:
> `git diff [--cached]  | git apply -p2 --directory VERSION --reject`

@praiskup beautiful. Learning this already paid off the PR :)
Thank you!

rhcarvalho:
Updated with @praiskup's awesome suggestion.

@bparees PTAL

CONTRIBUTING.md
```diff
@@ -0,0 +1,45 @@
...
+find -name '*.rej' -exec cat {} \;
+```
+
+Fix the differences manually, then delete the `*.rej` files.
+
+Notes:
+
+- Sometimes it may be useful to ignore part of the context passing the `-C<n>`
+  flag to `git apply`.
+- Read `git diff --help` and `git apply --help` for details about the flags used
```
php-coder: We can save 2 letters to type if will use `git help diff` and `git help apply` :)

bparees:
thanks @rhcarvalho

PR #196, opened 2016-10-07
Some bash improvements

rhcarvalho:
- The pkill change is the same as recently applied to the `cleanup`
  function in the `run-mongod` script (#191).
- Quoting and bracketing following [Google's style guide](https://google.github.io/styleguide/shell.xml) (that we've been loosely following).

### DIFF
--- a/2.4/root/usr/bin/run-mongod-replication
+++ b/2.4/root/usr/bin/run-mongod-replication
@@ -33,9 +33,7 @@ function cleanup() {
     mongo_remove
   fi
   echo "=> Shutting down MongoDB server ..."
-  if pgrep mongod; then
-    pkill -2 mongod
-  fi
+  pkill -INT mongod || :
   wait_for_mongo_down
   exit 0
 }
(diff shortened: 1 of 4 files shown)


php-coder:
If you're going to fix all of quoting issues then `git grep '$[^({@1]' 3.2/root/ | grep -v '#'` could be helpful :)

php-coder:
LGTM.

praiskup:
@php-coder I agree that this is issue.  Could you write automatic check that will show us (except for white-listed uses) usage of unquoted shell constructs?  The standard way to do this is `make syntax-check` ...

For inspiration, GNU libtool has (for us too pedantic) checker.

praiskup:
FWIW Google should call that "Bash" guidelines

php-coder:
> Could you write automatic check that will show us (except for white-listed uses) usage of unquoted shell constructs?

@praiskup I'm not ready yet :)

rhcarvalho:
The intention here was neither to solve all existing occurrences nor to write an automated checker.

My bad for not making that clear upfront.

These changes came from:
1. realizing I missed one occurrence of `pkill mongod` in #191 
2. "forward-porting" some of the quoting and bracketing improvements @php-coder made in #184

bparees:
lgtm

PR #38, opened 2015-06-09
Make mongo work with docker exec -ti bash -c mongo

mnagy:
@mfojtik @hhorak @jhadvig

PR #198, opened 2016-10-11
Remove call to mongo with no arguments

rhcarvalho:
This hangs the tests forever if run with Docker 1.12 (tests work fine
with e.g., Docker 1.10).

The code is unnecessary now, as it seems to be a left-over of the
original way to check that SCL was enabled (introduced in #38). Now we
already have specific tests for SCL enablement, executed just two lines
above the line we're removing.

### DIFF
--- a/2.4/test/run
+++ b/2.4/test/run
@@ -45,11 +45,6 @@ function mongo_admin_cmd() {
     docker run --rm $IMAGE_NAME mongo admin --host $CONTAINER_IP -u admin -p "$ADMIN_PASS" --eval "${@}"
 }
 
-function mongo_cmd_local() {
-    local name="$1" ; shift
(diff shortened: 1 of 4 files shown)


rhcarvalho:
@bparees PTAL

rhcarvalho:
Without this, tests hang midway:

``` console
$ make test VERSION=3.2 SKIP_SQUASH=1
...
+ docker exec d354f25167e6a45e1523427b03612b54122b9a41ffef7c61e576c51b18b02b4a bash -c 'mongo '
MongoDB shell version: 3.2.6
connecting to: test

```

I could isolate this to the Docker version and the way how it passes or not stdin in exec calls.

bparees:
lgtm

PR #200, opened 2016-10-13
3.2: remove smallfiles and noprealloc options

php-coder:
Inspired by @omron93 comment in #184: [`Mongodb 3.2 uses wiredTiger storage engine - these options are for old mmapv1 engine`](https://github.com/sclorg/mongodb-container/pull/184#discussion_r82174964)

If this PR will be merged after #184 then it should be updated to also update `run-mongod-pet` script.

PTAL @bparees @rhcarvalho @omron93

### DIFF
--- a/3.2/root/usr/share/container-scripts/mongodb/mongodb.conf.template
+++ b/3.2/root/usr/share/container-scripts/mongodb/mongodb.conf.template
@@ -10,9 +10,9 @@ dbpath = ${MONGODB_DATADIR}
 # only for mounted data directory from older MongoDB server
 noprealloc = true
 
-# Set MongoDB to use a smaller default data file size. Default: true
+# Set MongoDB to use a smaller default data file size.
 # only for mounted data directory from older MongoDB server
-smallfiles = ${MONGODB_SMALLFILES}
+smallfiles = true
 
 # Runs MongoDB in a quiet mode that attempts to limit the amount of output.
 # Default: true
--- a/3.2/root/usr/share/container-scripts/mongodb/mongodb.conf.template
+++ b/3.2/root/usr/share/container-scripts/mongodb/mongodb.conf.template
@@ -6,9 +6,9 @@ port = ${CONTAINER_PORT}
 # Default: /var/lib/mongodb/data
 dbpath = ${MONGODB_DATADIR}
 
-# Disable data file preallocation. Default: true
+# Disable data file preallocation.
 # only for mounted data directory from older MongoDB server
-noprealloc = ${MONGODB_NOPREALLOC}
+noprealloc = true
 
 # Set MongoDB to use a smaller default data file size. Default: true
 # only for mounted data directory from older MongoDB server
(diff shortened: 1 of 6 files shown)


3.2/root/usr/share/container-scripts/mongodb/mongodb.conf.template
```diff
@@ -6,13 +6,13 @@ port = ${CONTAINER_PORT}
 # Default: /var/lib/mongodb/data
 dbpath = ${MONGODB_DATADIR}
 
-# Disable data file preallocation. Default: true
+# Disable data file preallocation.
 # only for mounted data directory from older MongoDB server
-noprealloc = ${MONGODB_NOPREALLOC}
+noprealloc = true
```
rhcarvalho, line 11: If we are removing the env var, why do we keep the value in the configuration template?

php-coder: In the original comment @omron93 said that "Config file can contain it (it takes effect only if different storage engine is specified or data from old engine are present)."

php-coder:
@omron93 Please, take a look :)

3.2/root/usr/share/container-scripts/mongodb/mongodb.conf.template
rhcarvalho: Well, if we remove the user's ability to set that via env var, shouldn't we just leave the configuration to use the default values?

IIUC, this setting only applies to users migrating from an older version to 3.2 with existing data (in the old storage format).

omron93:
@php-coder Mentioned comment was addressed specifically to petSet. It is OpenShift specific and petSet is addressing 3.2 versions, so I was asking if OpenShift support older data in this new feature.

In standalone mode (which it supported from 2.4 version) why not to let to use old data? For example when it is upgrading from older version. He may for some reason want to use old storage engine or do some changes before migration to WiredTiger...
I hope #178 will be merged so replication won't be openshift specific too.

These are reasons why I leaved env. variables in 3.2 in #127 And in mentioned comment I only wanted to know what OpenShift allows to users ,...

omron93:
@php-coder You answer seems reasonable - https://github.com/sclorg/mongodb-container/pull/184/files#r83625592 Custom config file is a good alternative.

So LGTM.

3.2/root/usr/share/container-scripts/mongodb/mongodb.conf.template
omron93: User can use custom config file. So +1 for removing config option completely.

rhcarvalho:
@bparees I think there was still a discussion about `noprealloc = true` in the configuration template.
Pretty strange to have it hard coded in the config.

bparees:
@rhcarvalho looked resolved to me.  the resolution was "replace the file if you want to change it, which you have to do anyway if you want to change the engine which is the only way to make that setting matter"

omron93:
> have to do anyway if you want to change the engine which is the only way to make that setting matter

@bparees It is not true. You can mount data from old storage engine and old engine is set automatically.

PR #201, opened 2016-10-14
Apply some improvements to old replica set scripts

rhcarvalho:
Based on work on new scripts to support the replication example using
a PetSet (#184), these little improvements make the different scripts a bit
easier to compare, and who knows one day unify.

Notes:
- MONGODB_SERVICE_NAME has probably always been used, just never listed.
- The name 'supervisor' is at least confusing, for the script doesn't
  "supervise" anything.

### DIFF
--- a/2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
@@ -4,10 +4,9 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-source ${CONTAINER_SCRIPTS_PATH}/common.sh
+source "${CONTAINER_SCRIPTS_PATH}/common.sh"
 
-# Initiate replicaset and exit
-if [[ "$1" == "initiate" ]]; then
+function initiate() {
   current_endpoints=$(endpoints)
   if [ -n "${MONGODB_INITIAL_REPLICA_COUNT:-}" ]; then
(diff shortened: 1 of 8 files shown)


rhcarvalho:
[test][openshift] @bparees @php-coder PTAL

php-coder:
LGTM.

2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
```diff
@@ -4,10 +4,9 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-source ${CONTAINER_SCRIPTS_PATH}/common.sh
+source "${CONTAINER_SCRIPTS_PATH}/common.sh"
```
omron93, line 7: Are double quotes necessary?

omron93:
@rhcarvalho LTGM. Thank you very much for this.

2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
rhcarvalho: Depends on the value stored in `CONTAINER_SCRIPTS_PATH`...
Quoting gives a bit more predictable behavior, and is what is suggested in https://google.github.io/styleguide/shell.xml#Quoting

Note, I didn't try to go change all the quoting issues, this is just a small thing taken from #184.

PR #205, opened 2016-10-17
3.2: remove smallfiles and preallocation options from default configuration

php-coder:
We all agreed that we can remove this options completely from the default configuration. If user has data in old format he should provide custom configuration file.

Follow-up to #200

PTAL @bparees 
CC @rhcarvalho @omron93


bparees:
[test]

PR #206, opened 2016-10-18
Replica set example that uses PetSet (v2)

php-coder:
This is effectively the same as https://github.com/sclorg/mongodb-container/pull/184 but without all of our comments. I hope that new PR will allow us to overcome problems with unicorns that are shown for the old PR sometimes.

PTAL @bparees

### DIFF
--- a/3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
+++ b/3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
@@ -1,0 +1,159 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+source "${CONTAINER_SCRIPTS_PATH}/common.sh"
+
+# This is a full hostname that will be added to replica set
+# (for example, "replica-2.mongodb.myproject.svc.cluster.local")
+readonly MEMBER_HOST="$(hostname -f)"
+
+# Description of possible statuses: https://docs.mongodb.com/manual/reference/replica-states/
+readonly WAIT_PRIMARY_STATUS='
+  while (rs.status().startupStatus || (rs.status().hasOwnProperty("myState") && rs.status().myState != 1)) {
+    printjson(rs.status());
+    sleep(1000);
+  };
+  printjson(rs.status());
+'
+readonly WAIT_PRIMARY_OR_SECONDARY_STATUS="
+  var mbrs;
+  while (!mbrs || mbrs.length == 0 || !(mbrs[0].state == 1 || mbrs[0].state == 2)) {
+    printjson(rs.status());
+    sleep(1000);
+    mbrs = rs.status().members.filter(function(el) {
+      return el.name.indexOf(\"${MEMBER_HOST}:\") > -1;
+    });
+  };
+  print(mbrs[0].stateStr);
+"
+
+# Outputs available endpoints (hostnames) to stdout.
+# This also includes hostname of the current pod.
+#
+# Uses the following global variables:
+# - MONGODB_SERVICE_NAME (optional, defaults to 'mongodb')
+function find_endpoints() {
+  local service_name="${MONGODB_SERVICE_NAME:-mongodb}"
+
+  # Extract host names from lines like this: "10 33 0 mongodb-2.mongodb.myproject.svc.cluster.local."
+  dig "${service_name}" SRV +search +short | cut -d' ' -f4 | rev | cut -c2- | rev
+}
+
+# TODO: unify this and `mongo_initiate` from common.sh
+# Initializes the replica set configuration. It is safe to call this function if
+# a replica set is already configured.
+#
+# Arguments:
+# - $1: host address[:port]
+#
+# Uses the following global variables:
+# - MONGODB_REPLICA_NAME
+# - MONGODB_ADMIN_PASSWORD
+# - WAIT_PRIMARY_STATUS
+function initiate() {
+  local host="$1"
+
+  if mongo --eval "quit(db.isMaster().setName == '${MONGODB_REPLICA_NAME}' ? 0 : 1)" --quiet; then
+    info "Replica set '${MONGODB_REPLICA_NAME}' already exists, skipping initialization"
+    return
+  fi
+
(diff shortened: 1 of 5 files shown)


bparees:
this one opens for me.  will wait for [test] success.

examples/petset/README.md
```diff
@@ -0,0 +1,183 @@
...
+previous number, it is guaranteed that the replica set may elect a new PRIMARY
+in case one of the pods that was deleted had that role. For example, that is the
+case when going from 5 to 3 members.
+
+On the other hand, scaling down to a lower number will temporarily render the
+replica set to have only SECONDARY members and be in read-only mode. That would
+be the case when scaling from 5 down to 1 member.
+
+The next step is to update the replica set configuration to
+[remove members](https://docs.mongodb.com/manual/tutorial/remove-replica-set-member/)
```
omron93, line 165: @php-coder Is this possible?

After 5->1 scale down, no PRIMARY is elected. Am I right? If so, is it possible to change replset config on non-PRIMARY member?

3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
```diff
@@ -0,0 +1,159 @@
...
+      }
+      sleep(${SLEEP_TIME});
+    }
+    printjson(ret);
+    quit(1);
+  "
+
+  # TODO: replace this with a call to `replset_addr` from common.sh, once it returns host names.
+  local endpoints
+  endpoints="$(find_endpoints | paste -s -d,)"
```
omron93, line 111: @php-coder Why do you use hostnames for addressing? Any benefit than using `endpoins()` and `replset_addr()` from common.sh?

examples/petset/README.md
php-coder: This paragraph relates to any scale down procedure and not to the paragraph about case of scaling down from 5 to 1. In other words, when user scales down he/she has to modify replica set configuration manually.

3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
php-coder: It seems more natural. Also idea of using `endpoints()` came too late, when I didn't want to do improvements like this (because of possible breakage).

omron93:
@php-coder @rhcarvalho It somehow missed me that #184 is ready for review :-)

examples/petset/README.md
omron93: What about adding some note that after big scale down it won't be possible?

php-coder: It's already there: "On the other hand, scaling down to a lower number will temporarily render the replica set to have only SECONDARY members and be **in read-only mode**. That would be the case when scaling from 5 down to 1 member."

omron93: > It's already there: "On the other hand, scaling down to a lower number will temporarily render the replica set to have only SECONDARY members and be in read-only mode. That would be the case when scaling from 5 down to 1 member."

Yes. And in next paragraph there is that user have to update config manually. But it is not possible in this case!
Also I think that "replica set to have only SECONDARY members and be in read-only mode" might not be true too. It depends on what is a official OpenShift way how user applications should address replicaset?

3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
omron93: No technical differences?

php-coder: AFAIU @rhcarvalho said that there is no differences.

omron93: One big difference is that code could be shared.

Also using hostnames could be little bit slower :-)

php-coder: If you sure that it will work and give us some benefits, why are you commenting here and not providing PR? ;-)

omron93: I wanted to know why it was chosen. PR was merged before I managed to ask.

rhcarvalho: The reason we agreed to go with this alternative implementation + TODO was to minimize changes that could break the example.

Eliminating the duplication is desired, thus the TODO.

Worth noting that while here accessing the replica set via IPs will work, we definitely want to have host names in the replica set configuration and NOT IP addresses. The reason is that IPs can change over time, while host names are stable.

omron93: > we definitely want to have host names in the replica set configuration and NOT IP addresses. The reason is that IPs can change over time, while host names are stable.

I know this and understand benefits.

djohnston89:
Does anyone have a working OpenShift template/ Docker image for this PetSet clustering implementation that might work with OpenShift Container Platform version 3.3/3.4?

bparees:
@djohnston89 see https://github.com/sclorg/mongodb-container/tree/master/examples/petset

it uses the centos/mongo-32-centos7 image.

djohnston89:
@bparees unfortunately I apparently don't have privileges to create a stateful set. 
'Cannot create stateful set "mongodb". User "****" cannot create apps.statefulsets in project "******"'

veronicaalange:
Is there a workaround anyone has found for "cannot create apps.statefulsets in project " ?

bparees:
@djohnston89 @legracen it probably means you're running an older version of openshift when it was called the PetSet resource and not the StatefulSet resource.  I'd suggest you try a newer version, you shouldn't need any special permissions.

Barring that, you can try going back to the older version of the template that defined PetSet resources:
https://github.com/sclorg/mongodb-container/blob/cf99d38bfcff2657fd1effba7c2bf69e6bc96dfd/examples/petset/mongodb-petset-persistent.yaml

I'm not sure if it has bit-rotted in any other ways, though.

PR #218, opened 2017-01-17
switch api to statefulset

bparees:
fixes https://github.com/openshift/origin/issues/12526

PR #219, opened 2017-01-18
Docs: StatefulSet -> PetSet

php-coder:
Follow-up to #218 where we've updated a template only.

PTAL @bparees

### DIFF
--- a/examples/petset/mongodb-petset-persistent.yaml
+++ b/examples/petset/mongodb-petset-persistent.yaml
@@ -100,9 +100,6 @@ objects:
           # this label will be used for count running pods
           labels:
             name: "${MONGODB_SERVICE_NAME}"
-          annotations:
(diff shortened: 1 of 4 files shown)


examples/petset/mongodb-petset-persistent.yaml
```diff
@@ -100,9 +100,6 @@ objects:
           # this label will be used for count running pods
           labels:
             name: "${MONGODB_SERVICE_NAME}"
-          annotations:
-            # 'false' pauses a petset after creation of each pet, to avoid it we set it to 'true'
-            pod.alpha.kubernetes.io/initialized: "true"
```
php-coder, line 105: It's enabled by default now: https://kubernetes.io/docs/tasks/manage-stateful-set/upgrade-pet-set-to-stateful-set/#differences-between-alpha-petsets-and-beta-statefulsets

bparees:
lgtm, thanks @php-coder

PR #220, opened 2017-01-23
Timezone for container

git001:
The default timezone is utc. Some customers need to have there local timezone. It would be nice to be able to set the timezone at the creation time

### DIFF
--- a/2.4/examples/replica/mongodb-clustered.json
+++ b/2.4/examples/replica/mongodb-clustered.json
@@ -51,6 +51,11 @@
       "description": "The value of the MongoDB Key (see for details: https://docs.mongodb.org/v2.4/tutorial/generate-key-file/)",
       "generate": "expression",
       "from": "[a-zA-Z0-9]{255}"
+    },
+    {
+      "name": "TIMEZONE",
+      "description": "The time zone for the mongodb cluster",
+      "value": "UTC"
     }
   ],
   "objects": [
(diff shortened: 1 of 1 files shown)


omron93:
Thanks. It looks good to me. Please update petset template also (https://github.com/sclorg/mongodb-container/blob/master/examples/petset/mongodb-petset-persistent.yaml).

@bparees What about OpenShift POV - it is fine with its policies (to set timezone)?

[test]

git001:
done #223

2.4/examples/replica/mongodb-clustered.json
```diff
@@ -163,6 +168,10 @@
                   {
                     "name": "MONGODB_KEYFILE_VALUE",
                     "value": "${MONGODB_KEYFILE_VALUE}"
+                  },
+                  {
```
omron93, line 172: This part is missing in #223. @git001 Please add it (TIMEZONE has no effect without this).

bparees:
I don't know that we have a policy, we've discussed similar changes elsewhere a few times and usually conclude it should just be configured by the user.  Since this is just enabling it to be configured by the user (and providing a default), it seems fine.

omron93:
> I don't know that we have a policy, we've discussed similar changes elsewhere a few times and usually conclude it should just be configured by the user

Thanks. I meant your previous decisions about timezone = "policy" :-)

omron93:
[test]

omron93:
Replication using `ReplSet` is removed. Closing, discuss in #223 instead.

PR #221, opened 2017-01-23
Apply "Fix removing master from replset #186" to all versions.

omron93:
This PR apply changes merged in https://github.com/sclorg/mongodb-container/pull/186 to all versions (PR was merged with changes only applied to 3.2 version).

Description of original PR:

> This PR aims to fix 'stepping down server from PRIMARY role'.
>
> When member is going to be removed from replset and this member is PRIMARY, it should check that this won't break replset - that replset is able to vote a new PRIMARY.
In current situation while removing PRIMARY it could happen that not all members of replset are up-to-date and replset user can lost some information.
>
> Code which is now running after post deploy hook successfully initialize replset performs removing PRIMARY from replset. So removing this code from init-replset.sh and post-hook kills main scripts and this calls cleanup() (there the PRIMARY member is clearly removed).

> Also this fixes https://bugzilla.redhat.com/show_bug.cgi?id=1260586

@hhorak @bparees Please take a look and merge.

### DIFF
--- a/2.4/root/usr/bin/run-mongod-replication
+++ b/2.4/root/usr/bin/run-mongod-replication
@@ -30,6 +30,33 @@ function usage() {
 }
 
 function cleanup() {
+  if [ "$(mongo admin -u admin -p ${MONGODB_ADMIN_PASSWORD} --quiet --eval 'rs.isMaster().ismaster')" == "true" ]; then
+
+    # Wait that some SECONDARY is synced with PRIMARY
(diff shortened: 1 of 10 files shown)


hhorak:
[test]

hhorak:
Generally 2.4 version should already stay untouched, since it is no longer supported. But since it is already done this time and it fixes the mongodb24-related BZ, I'd be in favor merging this as it is. It looks good to me. LGTM after tests pass.

PR #222, opened 2017-01-24
Add simple openshift tests to this repository

omron93:
Tests for images are now packed together with source in `test/run` script. I would like add some simple scripts to test images in OpenShift environment.

It is quite easy to deploy OpenShift with `oc cluster up` and having these simple tests would allow developers/users to easily test changes or to add this testing to CI.

This PR adds new make target `test-openshift` - by it images are built and `test/run-openshift` is run (if exists), it is similar how `make test` is handled.

Also new template for replicaset is added to `examples` (it is a "copy" of template from `2.4/examples/` with parametrized image name).

Now `run-openshift` deploys OpenShift and created application from `examples/replset/mongodb-clustered.json` template. And it tests replicating data to all members and scaling up and down. The script should contains basically only instructions which user would use to create and administer application in OpenShift from command line.
Extending this scripts to test also petset is very welcomed (or providing hints to do it). Also maybe some things can be done simpler, so I would like to know it.

@hhorak @bparees Please review.

When I tested this it showed me some bugs what should be fixed in replication scripts. So in current implementation do not pass this every time. Also I hope adding this test to CI would make time needed to merge changes in replication scripts shorter.

### DIFF
--- a/3.2/root/usr/share/container-scripts/mongodb/test-functions.sh
+++ b/3.2/root/usr/share/container-scripts/mongodb/test-functions.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# insert_and_wait_for_replication insert data in host and wait all replset members
+# applied recent oplog entry
+function insert_and_wait_for_replication() {
+  local host
+  host=$1
+  local data
+  data=$2
+
+  # Storing document into replset and wait replication to finish
+  local script
+  script="db.getSiblingDB('test_db').data.insert(${data});
+    for (var i = 0; i < 60; i++) {
+      var status=rs.status();
+      var optime=status.members[0].optime;
+      var ok=true;
+      for(var j=1; j < status.members.length; j++) {
+        if(tojson(optime) != tojson(status.members[j].optime)) {
+          ok=false;
+        }
+      };
+      if(ok == true) {
+        print('INFO: All members of replicaset are synchronized');
+        quit(0);
+      }
+      sleep(1000);
+    }
+    print('ERROR: Members of replicaset are not synchronized');
+    printjson(rs.status());
+    quit(1);"
+
+  mongo admin --host "${host}" -u admin -p "${MONGODB_ADMIN_PASSWORD}" --eval "${script}"
+}
+
+# wait_replicaset_members waits till replset has specified number of members
+function wait_replicaset_members() {
+  local host
+  host=$1
+  local count
+  count=$2
+
+  local script
+  script="for (var i = 0; i < 60; i++) {
+    var ret = rs.status().members.length;
+    if (ret == ${count}) {
+      print('INFO: Replicaset has expected number of members');
+      quit(0);
+    }
+    sleep(1000);
+  }
+  print('ERROR: Wrong count of members in replicaset');
+  printjson(rs.status());
+  quit(1);"
+
+  mongo admin --host "${host}" -u admin -p "${MONGODB_ADMIN_PASSWORD}" --eval "${script}"
+}
--- a/3.2/test/run-openshift
+++ b/3.2/test/run-openshift
@@ -1,0 +1,87 @@
+#!/bin/bash
+#
+# Test the MongoDB image in OpenShift.
+#
+# IMAGE_NAME specifies a name of the candidate image used for testing.
+# The image has to be available before this script is executed.
+#
+
+set -exo nounset
+
+function prepare_oc_client() {
+  # Prepare oc command
+  if [[ ! -d ../openshift-client/ ]]; then
+    OC_RELEASE=$(curl -s https://api.github.com/repos/openshift/origin/releases | grep "browser_download_url" | grep "openshift-origin-client" | grep "linux-64bit" | sort | tail -n 1 | sed -e 's|"browser_download_url": ||' | tr -d ' "')
+
+    curl -L ${OC_RELEASE} | tar xz
+    mv ./openshift-origin-client-tools-* ../openshift-client
+  fi
+  export PATH=${PATH}:`pwd`/../openshift-client/
+}
+prepare_oc_client
+
+function cleanup() {
+  echo "Stopping and removing OpenShift cluster..."
+  oc delete project --all
+  oc cluster down
+}
+trap cleanup EXIT SIGINT
+
+function print_logs() {
+  oc get all
+  while read pod_info; do
+    pod=$(echo ${pod_info} | tr -s ' ' | cut -f1 -d' ')
+    echo "INFO: printing logs for pod ${pod}"
+    oc logs ${pod}
+  done < <(oc get pods --no-headers=true)
+}
+trap print_logs ERR
+
+# Start OpenShift cluster
+oc cluster up
+
+# Prepare image - add tag different than :latest
+# With :latest default imagePullPolicy is Always
+# https://docs.openshift.com/enterprise/3.2/dev_guide/managing_images.html#image-pull-policy
+docker tag ${IMAGE_NAME} ${IMAGE_NAME}:test-openshift
+IMAGE_NAME=${IMAGE_NAME}:test-openshift
+
+#
+# General functions
+#
+
+# wait_for_ready_pods wait till number of pods with label get ready
+function wait_for_ready_pods() {
+  local label="$1"
+  local count="$2"
+  for i in $(seq 30); do
+    if [[ "$(oc get pods --no-headers=true -l${label} | grep "1/1" | wc -l)" -eq ${count} ]]; then
+      return 0
+    fi
+    echo "Waiting for ${count} ready pods labeled with '${label}'..."
+    sleep 20
+  done
+  return 1
+}
+
+# Deploy MongoDB clustered application
+USER=user
+PASS=pass
+ADMIN_PASS=adminPass
+DB=db
+
+oc new-project petset-example
+oc new-app --file=../examples/petset/mongodb-petset-persistent.yaml -p MONGODB_USER=${USER} -p MONGODB_PASSWORD=${PASS} -p MONGODB_DATABASE=${DB} -p MONGODB_ADMIN_PASSWORD=${ADMIN_PASS} -p MONGODB_IMAGE=${IMAGE_NAME} -p VOLUME_CAPACITY=500M
+
+wait_for_ready_pods "app=mongodb-petset-replication" 3
+wait_for_ready_pods "openshift.io/deployer-pod-for.name" 0
+
+host="rs0/$(oc get endpoints mongodb --no-headers | tr -s ' ' | cut -f2 -d' ')"
+
+docker exec origin kubectl run --attach --restart=Never mongodb-test --image ${IMAGE_NAME} --env="MONGODB_ADMIN_PASSWORD=${ADMIN_PASS}" --command -- bash -c "set -x
+. /usr/share/container-scripts/mongodb/common.sh
+. /usr/share/container-scripts/mongodb/test-functions.sh
+wait_for_mongo_up '${host}'
+wait_replicaset_members '${host}' 3
+insert_and_wait_for_replication '${host}' '{a:5, b:10}'"
+


bparees:
we're really moving away from the replicaset example, you should base your tests on the petset(statefulset) replication instead.

hhorak:
That looks fantastic, I think we can use even most basic testing for start and see how it will work. One problem I see is that such test wouldn't run in parallel, right? Since one instance would `oc cluster down` other instance of the test.. Or is that wrong expectation?

omron93:
> we're really moving away from the replicaset example, you should base your tests on the petset(statefulset) replication instead.

@bparees I wanted to test both. But running petset example is harder for me. I've tried it (see the latest commit), but 
`oc new-app --file=../examples/petset/mongodb-petset-persistent.yaml -p MONGODB_USER=${USER},MONGODB_PASSWORD=${PASS},MONGODB_DATABASE=${DB},MONGODB_ADMIN_PASSWORD=${ADMIN_PASS},MONGODB_IMAGE=${IMAGE_NAME},VOLUME_CAPACITY=500M`
failed with `error: object does not implement the Object interfaces`. Do you have any idea what I do wrong?

I know that petset is preferred now. Problem is that I would like to support replication on other platforms than OpenShift. For example allow this with docker - #178 And I am not sure if using petset approach is applicable for it - dealing with stable network identifiers is hard IMPOV... In that case local code for adding/removing members used in replication would be the same. That is why I don't totally deprecate replicaset example and replicaset code (even it would not be marked as supported in OpenShift).
@bparees  Actually, it would be good to know what do you thing about replication using docker and staff like this? Before creating PRs...:-)

omron93:
> I think we can use even most basic testing for start and see how it will work.

Currently this OpenShift test is failing sometimes/for some versions. (mainly for pre 3.2 versions with slower initialization)
It seems to me that it is not a problem of the test but of replication scripts, so it would be good to have it running to be able to fix the code.
@bparees So am I working with OpenShift the right way? (to consider the result of this test valuable)

> One problem I see is that such test wouldn't run in parallel, right? Since one instance would oc cluster down other instance of the test.. Or is that wrong expectation?

@hhorak It is right expectation for coreservices Jenkins (RHEL testing).
In ci.centos.org every test is run on new VM, so this does not influence other tests. So it is not problem there... And I want to start with testing only CentOS versions. However it would be possible to modify RHEL ci configuration to use newly provisioned machines for OpenShift test - in case we would decide it is beneficial...

bparees:
> failed with error: object does not implement the Object interfaces. Do you have any idea what I do wrong?

your oc cluster up is probably too old.  the api name recently changed from petset to statefulset, the template was updated accordingly.

omron93:
> your oc cluster up is probably too old. the api name recently changed from petset to statefulset, the template was updated accordingly.

I've update from v1.3.2 to latest v1.4.1 and got the same error.
@bparees Any other idea please?

bparees:
> I've update from v1.3.2 to latest v1.4.1 and got the same error.

so you have a 1.4.1 cluster and you're using the version of the template that references statefulsets, not petsets?

omron93:
> so you have a 1.4.1 cluster and you're using the version of the template that references statefulsets, not petsets?

Exactly.

(I've pushed it to this PR. So you can try it by `make test-openshift SKIP_SQUASH=1 VERSION=3.2`)

bparees:
@omron93 sorry, i think the statefulset change may have happened in 1.5.

try using this older version of the template, from before we switched it to statefulset:
https://github.com/sclorg/mongodb-container/blob/cf99d38bfcff2657fd1effba7c2bf69e6bc96dfd/examples/petset/mongodb-petset-persistent.yaml

omron93:
> try using this older version of the template, from before we switched it to statefulset:
https://github.com/sclorg/mongodb-container/blob/cf99d38bfcff2657fd1effba7c2bf69e6bc96dfd/examples/petset/mongodb-petset-persistent.yaml

@bparees Thanks, it is "working". OpenShift created new app... but pods can't write to mounted persistent volume. There is any special setting about permissions?

EDIT: I've solved problem with permissions. (code is pushed to this PR)

I have another question. Should it be possible to access MongoDB database outside of OpenShift? I've tried it during `run-openshift` (to wait for PRIMARY election) and it seems to me that it is not possible with petset deployment.

> @omron93 sorry, i think the statefulset change may have happened in 1.5.

1.5 is not stable, is it? Why StatefulSet is already in example in `examples/petset/`?

bparees:
> I have another question. Should it be possible to access MongoDB database outside of OpenShift? I've tried it during run-openshift (to wait for PRIMARY election) and it seems to me that it is not possible with petset deployment.

it should be, you would need to add a route of course, which the template does not define.

omron93:
> it should be, you would need to add a route of course, which the template does not define.

Not sure if I understand it right - route is binding of hostname to the service.
I think that problem is that network identifiers used in addressing inside replset are not resolvable outside of OpenShift cluster.
So if you use addressing for replicasets (https://docs.mongodb.com/manual/reference/connection-string/) - not single node! - it is not working with petset.

@bparees Could you please try it for petset deployment?
Use `mongo --host mongodb://<USER>:<PASS>@<HOST>/admin?replicaSet=rs0` or `mongo admin -u <USER> -p<PASS> --host rs0/<HOST>`.

bparees:
if the mongo instances are reporting their container hostname to clients and expecting the clients use those hostnames to reach them, then I agree, that's going to be broken.

you'd need a route to each pod and to configure the mongo instances to use those routes as their hostname, I guess.

omron93:
> if the mongo instances are reporting their container hostname to clients and expecting the clients use those hostnames to reach them, then I agree, that's going to be broken.

Mongo instances have their hostnames used in "internal" replicaset config (for petest deployment, for replicaset deployment there are containers IPs). And clients updates host configuration where they connects from this "internal" replicaset config (this allows to keep connections although replicaset members are added/removed,...).

> you'd need a route to each pod and to configure the mongo instances to use those routes as their hostname, I guess.

@bparees So how I should connect to the mongodb petset cluster?
(In this test I want to insert something to mongodb cluster and there is no easy way how to do it) I am sorry that I ask about everything, but I am not so experienced OpenShift user :-(

bparees:
A couple options....  you could oc rsh into one of the pods and run the mongo commands from there

or you could spin up another pod and the commands in that other pod.

bparees:
(or just use oc exec to run the mongo command within one of the other pods)

omron93:
> A couple options.... you could oc rsh into one of the pods and run the mongo commands from there
or you could spin up another pod and the commands in that other pod.

@bparees Thanks, I've managed to run it in new pod.

PR is squashed and replset testing is removed. Please review.

hack/run-openshift
```diff
@@ -0,0 +1,183 @@
...
+      return 0
+    fi
+    echo "Waiting for ${count} ready pods labeled with '${label}'..."
+    sleep 20
+  done
+  return 1
+}
+
+# create_volumes reates Kubernetes volumes with given name prefix, specified capacity and accessMode
+function create_volumes() {
```
bparees: oc cluster up, as of 1.5, actually creates volumes automatically, so if you move to the 1.5 alpha release you can drop this.

hack/run-openshift
```diff
@@ -0,0 +1,183 @@
...
+# Deploy MongoDB clustered application
+USER=user
+PASS=pass
+ADMIN_PASS=adminPass
+DB=db
+
+create_volumes 3 "mongo-data" "500M" "ReadWriteOnce"
+
+oc new-project petset-example
+oc new-app --file=https://raw.githubusercontent.com/sclorg/mongodb-container/cf99d38bfcff2657fd1effba7c2bf69e6bc96dfd/examples/petset/mongodb-petset-persistent.yaml -p MONGODB_USER=${USER} -p MONGODB_PASSWORD=${PASS} -p MONGODB_DATABASE=${DB} -p MONGODB_ADMIN_PASSWORD=${ADMIN_PASS} -p MONGODB_IMAGE=${IMAGE_NAME} -p VOLUME_CAPACITY=500M
```
bparees: i'd reference the latest file (especially once you move to the new oc cluster up which should have the new statefulset api) to ensure the template doesn't get broken in the future.

hack/run-openshift
omron93: Good to know, thanks.

hack/run-openshift
omron93: @bparees Thanks, updated.

Now tests will use latest release (even unstable).

hack/run-openshift
```diff
@@ -0,0 +1,153 @@
...
+# IMAGE_NAME specifies a name of the candidate image used for testing.
+# The image has to be available before this script is executed.
+#
+
+set -exo nounset
+
+function prepare_oc_client() {
+  # Prepare oc command
+  if [[ ! -d ../openshift-client/ ]]; then
+    OC_RELEASE=$(curl -s https://api.github.com/repos/openshift/origin/releases | grep "browser_download_url" | grep "openshift-origin-client" | grep "linux-64bit" | sort | tail -n 1 | sed -e 's|"browser_download_url": ||' | tr -d ' "')
```
bparees: i'm getting a rate limiting failure hitting that url, i suspect that may cause flake issues for you...

bparees:
aside from the rate limiting question, lgtm.

hack/run-openshift
omron93: @bparees So when you run this, curl failed?... It haven't happen to me and I am not sure I understand it right.

Is there any other way how to get `oc` command?

omron93:
I am getting error after switching to oc v1.5 and auto PV creating. It seems content of PV is not cleaned after `oc cluster down` and it is used after running test again (and it fails).

`sudo rm -rf /var/lib/origin/openshift.local.pv` worked for me
@bparees Is there any other way how to clean content of PV before/after `oc cluster down`?

hack/run-openshift
bparees: yes, possibly because i was behind a proxy since the rate limiting is by IP.

getting the oc command isn't the problem, it's the call to the github api that you're using to determine what the latest release is that could be problematic if you're rate limited.

it's probably ok to leave it for now, but if you start seeing issues you may just want to hardcode the oc binary to download (and bump it as needed)

bparees:
@omron93 i'll have to defer to @csrwng on issues with oc cluster down not cleaning up/reusing PV dirs.

csrwng:
@omron93 we currently don't cleanup those directories for you. If you remove a claim, the directory is recycled and its contents are removed. However, for cleaning up everything after your cluster is down, I don't have any suggestions besides what you're already doing.

omron93:
>we currently don't cleanup those directories for you. If you remove a claim, the directory is recycled and its contents are removed. However, for cleaning up everything after your cluster is down, I don't have any suggestions besides what you're already doing.

@csrwng How can I "remove a claim" please? Removing whole project before `cluster down` is not removing a claim?
(problem is that removing `/var/lib/origin/openshift.local.pv` requires root)

[test]

omron93:
[test]

omron93:
[test]

csrwng:
@omron93 deleting a pvc is removing a claim. And yes, if you delete a project that contains a pvc, that should remove it.

omron93:
> If you remove a claim, the directory is recycled and its contents are removed.

I do `oc cluster up`, run application which uses persistent volumes, then run `oc delete project --all` and `oc cluster down`. But after repating this several times, application fails, because content of PV is not removed before using it.
@csrwng Am I doing something wrong?

Also after shutting down the cluster, so files in `openshift.local.volumes/pods` are still mounted (one example - `/var/lib/origin/openshift.local.volumes/pods/50a216b8-ffd8-11e6-a090-28d244c2c17c/volumes/kubernetes.io~secret/router-token-6rcz3`) so file can't be removed without unmounting.
Is this expected behavior?

omron93:
Rebased. Also functions for working with replicaset were moved to image. It makes creating test easier - accessing replicaset requires running tests from container in Kubernetes, so it is easier that scripts are already there.

@bparees Please review.

omron93:
It is possible to run `make test-openshift` (see sclorg/container-common-scripts#2) by adding [test-openshift] in comment (it should not be triggered automatically by new commit).

This is now enabled only for mongodb now, I will enable it for other images later (after all will use sclorg/container-common-scripts repository).

omron93:
[test-openshift]

omron93:
[test-openshift]

omron93:
[test-openshift]

omron93:
[test-openshift]

omron93:
[test-openshift]

3.2/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -259,3 +259,59 @@ function setup_keyfile() {
 function info() {
   printf "=> [%s] %s\n" "$(date +'%a %b %d %T')" "$*"
 }
+
+# insert_and_wait_for_replication insert data in host and wait all replset members
```
bparees: since the two scripts you're adding are only used from test code, it seems a bit confusing to put them in the image itself.  can you move them to a common test util script instead/

bparees (changes requested):

3.2/root/usr/share/container-scripts/mongodb/common.sh
omron93: The reason why I moved them into the image it, that running tests which access replicaset requires running it inside of kubernetes. So it have to be run in container (discussed above - https://github.com/sclorg/mongodb-container/pull/222#issuecomment-275657513), which also contains `mongo` client. And tested image itself seems to me to be the best candidate to run tests in.

Therefore I moved them to image itself, to avoid declaring them through `kubectl run`. And also to be easily re-run tests by somewhere else.

But if it seems confusing I can move them somewhere...

 >  can you move them to a common test util script instead

Do you this "common test util script" in image or only separated in own file?

omron93:
[test-openshift]

3.2/root/usr/share/container-scripts/mongodb/common.sh
bparees: ok, i didn't appreciate that you needed the script content inside the image itself, so that makes sense.

i'd still suggest either a separate file, or at least some comments on the functions indicating what they are used for/that they are for test purposes only.

omron93: @bparees Moved to separate file.

[test-openshift]

omron93:
[test-openshift]

omron93:
[test-openshift]

omron93:
@bparees If this PR is OK, please merge.

PR #224, opened 2017-02-03
Add example for replication with Docker.

omron93:
This PR allows user to create replset using only local docker. Also test for testing local replset is added.

This is replacement of #178.

Creating of replset would be (taken from README):
```
cat > variables <<EOF
MONGODB_DATABASE=db
MONGODB_USER=user
MONGODB_PASSWORD=password
MONGODB_ADMIN_PASSWORD=adminPassword
MONGODB_REPLICA_NAME=rs0
MONGODB_KEYFILE_VALUE=xxxxxxxxxxxx
MONGODB_SMALLFILES=true
MONGODB_SERVICE_NAME=mongodb"
EOF
source variables

IMAGE_NAME=centos/mongodb-32-centos7

network_name="mongodb-replset"
docker network create ${network_name}

docker run -d --cidfile $CIDFILE_DIR/replset0 --name=replset-0 --hostname=replset-0 --network ${network_name} --network-alias mongodb --env-file=variables $IMAGE_NAME run-mongod-replication
docker exec replset-0 bash -c "while ! [ -f /tmp/initialized ]; do sleep 1; done"
docker run -d --cidfile $CIDFILE_DIR/replset1 --name=replset-1 --hostname=replset-1 --network ${network_name} --network-alias mongodb --env-file=variables $IMAGE_NAME run-mongod-replication
docker exec replset-1 bash -c "while ! [ -f /tmp/initialized ]; do sleep 1; done"
docker run -d --cidfile $CIDFILE_DIR/replset2 --name=replset-2 --hostname=replset-2 --network ${network_name} --network-alias mongodb --env-file=variables $IMAGE_NAME run-mongod-replication
docker exec replset-2 bash -c "while ! [ -f /tmp/initialized ]; do sleep 1; done"
```

I am not sure about these two things:
1) remove members from replset config after container exits?
2) have one initializing pod which exits after replset is initialized?

(it is similar to difference petset vs. replset scripts in this repo)  

@bparees @hhorak What are you opinions?

IMPOV:
1) We are adding containers dynamically. So it seems to me logical also to remove them automatically. Absence of dynamic removing removes a lot of advantages of containers (quick reacting to changing situation), because it is required to manually update replset internal configuration. Also somehow specify unique IP for every starting container.
2) I am not sure about this. Docker allows setting restart policy for containers, so user could specify to automatically restart failed container. In case "container making initialization" equals "one member of replset" - it would not be possible to use automatic restarting (starting this container for second time will fail). So maybe I vote for exiting container which do initialization.

**UPDATE:**

Answer to both questions is "do it same as `PetSet` is Kubernetes`. So container only for initialization and no removing from mongodb replicaset configuration.
It is possible to use same code for docker replication as used in OpenShift template using `PetSet`. To do this, two things have to be changed:

- DNS of "pure" docker does not support  `SRV` records (was used in `init-replset.sh`). It was used during connecting to replicaset. Using A records is the same, because  SRV records are used immediately after they are obtained (so they are immediately resolved to IP addresses too)

- Container is using `hostname -f` to get its identification which is added to replicaset configuration. But this command is not in RHEL base image, so I added it. @bparees Are RHEL based images also tested and used in OpenShift? I thought yes...

### DIFF
--- a/2.6/Dockerfile.rhel7
+++ b/2.6/Dockerfile.rhel7
@@ -30,7 +30,7 @@ CMD ["run-mongod"]
 RUN yum install -y yum-utils && \
     yum-config-manager --enable rhel-server-rhscl-7-rpms && \
     yum-config-manager --enable rhel-7-server-optional-rpms && \
-    INSTALL_PKGS="bind-utils gettext iproute rsync tar v8314 rh-mongodb26-mongodb rh-mongodb26" && \
+    INSTALL_PKGS="bind-utils gettext iproute rsync tar hostname v8314 rh-mongodb26-mongodb rh-mongodb26" && \
     yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all
--- a/2.6/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/common.sh
@@ -73,7 +73,8 @@ function replset_addr() {
   local current_endpoints
   current_endpoints="$(endpoints)"
   if [ -z "${current_endpoints}" ]; then
-    echo >&2 "Cannot get address of replica set: no nodes are listed in service"
+    info "Cannot get address of replica set: no nodes are listed in service!"
+    info "CAUSE: DNS lookup for '${MONGODB_SERVICE_NAME:-mongodb}' returned no results."
     return 1
   fi
   echo "${MONGODB_REPLICA_NAME}/${current_endpoints//[[:space:]]/,}"
--- a/2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
@@ -10,18 +10,6 @@ source "${CONTAINER_SCRIPTS_PATH}/common.sh"
 # (for example, "replica-2.mongodb.myproject.svc.cluster.local")
 readonly MEMBER_HOST="$(hostname -f)"
 
-# Outputs available endpoints (hostnames) to stdout.
-# This also includes hostname of the current pod.
-#
-# Uses the following global variables:
-# - MONGODB_SERVICE_NAME (optional, defaults to 'mongodb')
-function find_endpoints() {
-  local service_name="${MONGODB_SERVICE_NAME:-mongodb}"
-
-  # Extract host names from lines like this: "10 33 0 mongodb-2.mongodb.myproject.svc.cluster.local."
-  dig "${service_name}" SRV +search +short | cut -d' ' -f4 | rev | cut -c2- | rev
-}
-
 # Initializes the replica set configuration.
 #
 # Arguments:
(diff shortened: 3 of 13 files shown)


bparees:
supporting dynamic removal makes sense though my understanding is that in practice no one does that, it's always a carefully planned operation undertaken by an admin.

for the restart, can the container not be smart enough to realize initialization should be skipped when it attempts it and fails?

omron93:
> supporting dynamic removal makes sense though my understanding is that in practice no one does that, it's always a carefully planned operation undertaken by an admin.

Aren't containers for different purpose? Quick and automated reacting to for example changing workload.

Automatic removing don't do anything what admin do not want. If I was an admin I would kill the container/mongod server, I would want to remove it from configuration too.
Or if terminating of planned by admin and occurs by some failure, it is hard to restore container/mongod server with same environment (IP address), so keeping server in configuration adds more work...

In petset deployment I understand that it makes sense - or better, to keep only few old members is configuration (to prevent replset unreachability during big scale down - more unreachable members that reachable one)
@bparees Did you have some customer questionare to know what they expect?

> for the restart, can the container not be smart enough to realize initialization should be skipped when it attempts it and fails?

It works good with mounted volume - information about replicaset is stored in local file.

Without local files it is more complicated. It could be possible to try to connect to the rest of containers. But there might be some network issue, container have to wait and retry... if it is a first run, other containers have to wait even longer (to initializer finishes retrying to reconnect + time to cluster initialization).
Not sure if having one container which initializes and can't be restarted is not better and safer way.

bparees:
> Aren't containers for different purpose? Quick and automated reacting to for example changing workload.

that can be one purpose, it's certainly not the only (or, i'd say, even the primary) one.  Regardless, containers don't make magic happen if the underlying software/best practices don't support it.  So for stateless apps, sure, it makes scaling up/down easier.  For an app that needs some orchestration around scaling up/down, containers don't solve that problem for you.  Adding/removing replicas can have a (temporary) impact on performance while data is replicated to new places, so it's not something that is going to be undertaken lightly in production.

> @bparees Did you have some customer questionare to know what they expect?

unfortunately not.  We don't have much customer feedback on clustered DB scenarios so far.  I think most customers are still more focused on migrating the stateless frontends to containers right now.

> Not sure if having one container which initializes and can't be restarted is not better and safer way.

there's nothing explicitly wrong with it, it just makes for a slightly more complicated template+topology.

omron93:
> there's nothing explicitly wrong with it, it just makes for a slightly more complicated template+topology.

@bparees And regarding replication using docker? (all this discussion)

In kubernetes petsets are good and more robust than replset, I understand. However in the world of Docker, "simulating" petset by manual assigning `--hostname` and `network-alias` for every container is painful and hard to automate in my point of view.

bparees:
> And regarding replication using docker? (all this discussion)

I don't really think users manually docker running a bunch of containers to construct a mongo replset is a realistic use case or one we're interested in promoting and supporting, but you guys own the image so I abstain from voting on the implementation path.

omron93:
[test]

omron93:
[test]

omron93:
[test]

omron93:
@bparees It is rebased. README and first PR comment is updated... Please review.

[test-openshift]

omron93:
[test-openshift]

bparees:
> But this command is not in RHEL base image, so I added it. @bparees Are RHEL based images also tested and used in OpenShift? I thought yes...

they are, but replication isn't a supported product feature so it may not get tested for the rhel images (we run some tests against the centos image on a nightly basis).

As long as this doesn't break statefulset replication (and if anything it sounds like you're fixing it for rhel), i abstain from otherwise having an opinion.

omron93:
Rebased.

> As long as this doesn't break statefulset replication (and if anything it sounds like you're fixing it for rhel), i abstain from otherwise having an opinion.

@bparees  OK, I understand. Anyway, thanks you very much for review.

omron93:
Rebased. [test-openshift]

@hhorak Please review.

hhorak:
I think we're ready to merge here, although I'm not sure whether it will influence the s2i PR #239. It looks like the replication features won't be easy to be just added as a separate file in `/usr/share/container-scripts/mongodb/pre-init/`, and it will stay a bit more separated, at least in the initial version of #239 .. @omron93, what do you think?

goern:
Could you guys have a look at https://github.com/fusor/ansible-playbook-bundle figure if that is a way to deploy a replset (so that we dont have to run multiple docker commands)?

omron93:
> Could you guys have a look at https://github.com/fusor/ansible-playbook-bundle figure if that is a way to deploy a replset (so that we dont have to run multiple docker commands)?

@goern Thanks for the tip.

Primary way to deploy replicaset is Kubernetes.
Aim of this PR is to be possible run replicaset mainly for development only by Docker. There are quite a lot alternatives how to deploy multicontainer app - Docker Composer, Nulecule, Ansible Container,... and each of them requires own "configuration". Anyway, example "configurations" how to run this image in any of these orchestrations solutions is really welcomed.

omron93:
> I think we're ready to merge here, although I'm not sure whether it will influence the s2i PR #239. It looks like the replication features won't be easy to be just added as a separate file in /usr/share/container-scripts/mongodb/pre-init/, and it will stay a bit more separated, at least in the initial version of #239 .. @omron93, what do you think?

@hhorak I hope merging any of these PRs should not cause problems. Rebase will be needed but nothing else.

I hope in latest versions of #239 using of customization scripts with replication is defined better. Please review again.

PR #227, opened 2017-02-09
Replica set fixes

jim-minter:
@bparees here is a set of fixes and simplifications that should improve our handling of mongo replica sets.

### DIFF
--- a/2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
@@ -11,13 +11,6 @@ source "${CONTAINER_SCRIPTS_PATH}/common.sh"
 readonly MEMBER_HOST="$(container_addr)"
 
 # Description of possible statuses: https://docs.mongodb.com/manual/reference/replica-states/
-readonly WAIT_PRIMARY_STATUS='
-  while (rs.status().startupStatus || (rs.status().hasOwnProperty("myState") && rs.status().myState != 1)) {
-    printjson(rs.status());
-    sleep(1000);
-  };
-  printjson(rs.status());
-'
 readonly WAIT_PRIMARY_OR_SECONDARY_STATUS="
   var mbrs;
   while (!mbrs || mbrs.length == 0 || !(mbrs[0].state == 1 || mbrs[0].state == 2)) {
@@ -39,7 +32,6 @@ readonly WAIT_PRIMARY_OR_SECONDARY_STATUS="
 # Uses the following global variables:
 # - MONGODB_REPLICA_NAME
 # - MONGODB_ADMIN_PASSWORD
-# - WAIT_PRIMARY_STATUS
 # - MONGODB_INITIAL_REPLICA_COUNT
 function initiate() {
   local host="$1"
@@ -78,7 +70,10 @@ function initiate() {
   local config="{_id: '${MONGODB_REPLICA_NAME}', $(replset_config_members "${current_endpoints}")}"
 
   info "Initiating MongoDB replica using: ${config}"
-  mongo admin --eval "rs.initiate(${config});${WAIT_PRIMARY_STATUS}" --quiet
+  mongo --eval "quit(rs.initiate(${config}).ok ? 0 : 1)" --quiet
+
+  info "Waiting for PRIMARY status ..."
+  mongo --eval "while (!rs.isMaster().ismaster) { sleep(100); }" --quiet
 
   info "Creating MongoDB users ..."
   mongo_create_admin
--- a/2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
@@ -149,10 +149,8 @@ function add_member() {
   info "Successfully joined replica set"
 }
 
-info "Waiting for local MongoDB to accept connections on ${MEMBER_HOST} ..."
-# connect using the host to ensure networking is working, otherwise
-# the add_member call will fail.
-wait_for_mongo_up ${MEMBER_HOST} &>/dev/null
+info "Waiting for local MongoDB to accept connections ..."
+wait_for_mongo_up &>/dev/null
 
 # Initialize replica set only if we're the first member
 if [ "$1" == "initiate" ]; then
--- a/2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
@@ -113,7 +113,7 @@ function add_member() {
       if (ret.code == 103) {
         quit(0);
       }
-      sleep(${SLEEP_TIME});
+      sleep(${SLEEP_TIME}*1000);
     }
     printjson(ret);
     quit(1);
--- a/2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
@@ -10,19 +10,6 @@ source "${CONTAINER_SCRIPTS_PATH}/common.sh"
 # (for example, "replica-2.mongodb.myproject.svc.cluster.local")
 readonly MEMBER_HOST="$(container_addr)"
 
-# Description of possible statuses: https://docs.mongodb.com/manual/reference/replica-states/
-readonly WAIT_PRIMARY_OR_SECONDARY_STATUS="
-  var mbrs;
-  while (!mbrs || mbrs.length == 0 || !(mbrs[0].state == 1 || mbrs[0].state == 2)) {
-    printjson(rs.status());
-    sleep(1000);
-    mbrs = rs.status().members.filter(function(el) {
-      return el.name.indexOf(\"${MEMBER_HOST}:\") > -1;
-    });
-  };
-  print(mbrs[0].stateStr);
-"
-
 # Initializes the replica set configuration. It is safe to call this function if
 # a replica set is already configured.
 #
@@ -92,7 +79,6 @@ function initiate() {
 # - MAX_ATTEMPTS
 # - SLEEP_TIME
 # - MONGODB_ADMIN_PASSWORD
-# - WAIT_PRIMARY_OR_SECONDARY_STATUS
 function add_member() {
   local host="$1"
   info "Adding ${host} to replica set ..."
@@ -128,18 +114,8 @@ function add_member() {
     return 1
   fi
 
-  info "Successfully added to replica set"
   info "Waiting for PRIMARY/SECONDARY status ..."
-
-  local rs_status_out
-  rs_status_out="$(mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --host "${replset_addr}" --eval "${WAIT_PRIMARY_OR_SECONDARY_STATUS}" --quiet || :)"
-
-  if ! echo "${rs_status_out}" | grep -xqs '\(SECONDARY\|PRIMARY\)'; then
-    info "ERROR: failed waiting for PRIMARY/SECONDARY status. Command output was:"
-    echo "${rs_status_out}"
-    echo "==> End of the error output <=="
-    return 1
-  fi
+  mongo --eval "while (!rs.isMaster().ismaster && !rs.isMaster().secondary) { sleep(100); }" --quiet
 
   info "Successfully joined replica set"
 }
--- a/2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
@@ -10,8 +10,7 @@ source "${CONTAINER_SCRIPTS_PATH}/common.sh"
 # (for example, "replica-2.mongodb.myproject.svc.cluster.local")
 readonly MEMBER_HOST="$(container_addr)"
 
-# Initializes the replica set configuration. It is safe to call this function if
-# a replica set is already configured.
+# Initializes the replica set configuration.
 #
 # Arguments:
 # - $1: host address[:port]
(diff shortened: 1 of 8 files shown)


jim-minter:
xref https://github.com/openshift/origin/issues/12588

2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
```diff
@@ -128,31 +79,27 @@ function add_member() {
...
-  local rs_status_out
-  rs_status_out="$(mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --host "${replset_addr}" --eval "${WAIT_PRIMARY_OR_SECONDARY_STATUS}" --quiet || :)"
-
-  if ! echo "${rs_status_out}" | grep -xqs '\(SECONDARY\|PRIMARY\)'; then
-    info "ERROR: failed waiting for PRIMARY/SECONDARY status. Command output was:"
-    echo "${rs_status_out}"
-    echo "==> End of the error output <=="
-    return 1
-  fi
+  while [[ "$(mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --eval 'rs.status().myState' --quiet || :)" != "2" ]]; do
```
omron93: After adding a member, new member can become even PRIMARY (in rare case).

2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
```diff
@@ -69,16 +47,10 @@ function initiate() {
...
-  if mongo --eval "quit(db.isMaster().setName == '${MONGODB_REPLICA_NAME}' ? 0 : 1)" --quiet; then
-    info "Replica set '${MONGODB_REPLICA_NAME}' already exists, skipping initialization"
-    return
-  fi
-
   local config="{_id: '${MONGODB_REPLICA_NAME}', $(replset_config_members "${current_endpoints}")}"
 
   info "Initiating MongoDB replica using: ${config}"
-  mongo admin --eval "rs.initiate(${config});${WAIT_PRIMARY_STATUS}" --quiet
+  mongo --eval "rs.initiate(${config}); while (!rs.isMaster().ismaster) { sleep(100); }" --quiet
```
omron93: Thanks, it is more "elegant" than removed code.

What about removing infinite loop?

3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
```diff
@@ -72,42 +45,21 @@ function initiate() {
...
-      sleep(${SLEEP_TIME});
-    }
-    printjson(ret);
-    quit(1);
-  "
-
   # TODO: replace this with a call to `replset_addr` from common.sh, once it returns host names.
   local endpoints
-  endpoints="$(find_endpoints | paste -s -d,)"
+  endpoints="$(find_endpoints | grep -v $MEMBER_HOST | paste -s -d,)"
```
omron93: Why using only local mongod address?

Connecting to replset with only this address won't work - this member is not initialized and connection with only this address will fail.

3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
```diff
@@ -118,31 +70,28 @@ function add_member() {
   local replset_addr
   replset_addr="${MONGODB_REPLICA_NAME}/${endpoints}"
 
-  if ! mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --host "${replset_addr}" --eval "${script}" --quiet; then
+  if ! mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --host "${replset_addr}" --eval "while (!rs.add('${host}').ok) { sleep(100); }" --quiet; then
```
omron93, line 76: Using '$(replset_addr)` from `common.sh` will work here and would solve the problem introduced by checking for DNS few lines above.

3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
```diff
@@ -118,31 +70,28 @@ function add_member() {
...
-  local rs_status_out
-  rs_status_out="$(mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --host "${replset_addr}" --eval "${WAIT_PRIMARY_OR_SECONDARY_STATUS}" --quiet || :)"
-
-  if ! echo "${rs_status_out}" | grep -xqs '\(SECONDARY\|PRIMARY\)'; then
-    info "ERROR: failed waiting for PRIMARY/SECONDARY status. Command output was:"
-    echo "${rs_status_out}"
-    echo "==> End of the error output <=="
-    return 1
-  fi
+  while [[ "$(mongo admin -u admin -p "${MONGODB_ADMIN_PASSWORD}" --eval 'rs.status().myState' --quiet || :)" != "2" ]]; do
```
omron93: No really sure why `WAIT_PRIMARY_STATUS` contained checking that `rs.status()` contains `myState` property. Are you sure, that `myState` is in the output also early after adding member to replicaset?

Maybe using `rs.isMaster()` would be safer.

omron93 (changes requested):
@jim-minter Thanks for this PR. Also I very appreciate that you keep petset and replset code same as much as possible, thanks.

Version 2.4 is deprecated, so no changes changes should be made (already some changes was not done for this version).

Also how is the `MONGODB_SERVICE_NAME` service working? (how kubernetes knows which pods should be listed). Also "internal" service is not used anywhere.

omron93:
I hope I understand how "internal" service work.

It should be used for mongodb addressing in our scripts, shouldn't it?

2.4/root/usr/bin/run-mongod
```diff
@@ -40,7 +40,7 @@ if ! [[ -v MONGODB_USER && -v MONGODB_PASSWORD && -v MONGODB_DATABASE && -v MONG
 fi
 
 # Run the MongoDB in 'standalone' mode
-mongod $mongo_common_args &
+mongod $mongo_common_args --bind_ip 127.0.0.1 &
```
bparees: comment this on why we're starting bound to localhost only first.

examples/petset/mongodb-petset-persistent.yaml
```diff
@@ -104,6 +120,7 @@ objects:
           containers:
             - name: mongo-container
               image: "${MONGODB_IMAGE}"
+              imagePullPolicy: IfNotPresent
```
bparees: this is the default, shouldn't need to specify it.

jim-minter: I've reminded myself: imagePullPolicy defaults to Always if tag is latest, otherwise IfNotPresent.  Given that the naming convention for this image is mongodb-32-centos7:latest, I happened to find this useful for debugging, but I agree it's not appropriate to check in.  Will revert.

2.4/root/usr/bin/run-mongod
jim-minter: Ack.

3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
jim-minter: grep -v *excludes* the local address :-)

2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
jim-minter: ok, will fix

3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
jim-minter: I don't think that's true, or I don't understand your point.  With stateful sets, the replica DNS name is stable, not the IP address, therefore we are using the DNS name here.

2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
jim-minter: I'm not convinced that setting an arbitrary timeout is better than having an infinite loop.  In the statefulset case, which I think is the case that people should really be using, the net effect of an infinite loop vs crashing out is the same: the 1st pod never becomes ready, the others don't deploy, the issue is clear from the pod logs.

3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
jim-minter: Yes, I think using rs.isMaster() is neater and possibly safer.  Although FWIW in the case that rs.status() doesn't contain the myState property, rs.status().myState evaluates to null.

jim-minter:
@omron93 internally the replicas use the ${MONGODB_SERVICE_NAME}-internal service, which is set to include running pods even if their readiness probe is not returning true, to find eachother.

A mongodb client can use ${MONGODB_SERVICE_NAME} to connect to any running replica, from which it can then discover the DNS names of all available replicas including the primary.

jim-minter:
@omron93 @bparees I have removed all changes to the 2.4 tree from this patchset. All changes requested above pushed.

3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
omron93: I see. Sorry.

3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
```diff
@@ -158,3 +107,5 @@ if [ "${MEMBER_ID}" = '0' ]; then
 else
   add_member "${MEMBER_HOST}"
 fi
+
+>/tmp/initialized
```
omron93, line 111: @jim-minter Why not to do the same for replicaset too? (to allow users to update their replicaset example if they want)

3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
omron93: DNS names are resolved to IP addresses before connection, so there is no difference. And IP addresses are used from same "source" as DNS names.

I wanted to remove ` # TODO: replace this with a call to `replset_addr` from common.sh, once it returns host names.` Because I think IP addresses and DNS names serves same in this usecase.

omron93 (approved):

3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
omron93: @jim-minter And why `MEMBER_HOST` is excluded?

3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
jim-minter: It is probably feasible to do so, but in practice I think all the code using k8s replicasets should be deprecated/removed from this repo - I can't think of a reason why anyone would/should use it now that statefulsets are available.  Indeed this repo doesn't even actually ship a template which enables their use - the only one was the mongodb-clustered.json under 2.4.

3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
jim-minter: You're right.  Thinking about it, it's not a necessary change - I will revert this line.

3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
jim-minter: I think there is a difference.  My intention is that the mongo rs config, once set, stays the same (unless the user changes the number of replicas).  The rs config stores the DNS names, so that if a pod dies and is rescheduled (and its IP address potentially changes in the process), the rs config need not be updated.  Mongo re-resolves the DNS name at each new connection attempt, so the potential IP address change is transparent to mongo.

omron93: I understand, this is difference: petset vs. replset. I think DNS does not brings no advantage when addresing petset (from mongodb clients).

IP addresses are not stored anywhere. The are got by the same way and in same time as DNS names (by `dig` command from service), so they points to the same containers while addressing mongodb replicaset. Only when using DNS names, another DNS query have to be done (immediately after getting DNS names from service) to get IP addresses.

However it is not important :-)

omron93:
@jim-minter Would not be better in readinessProbe to try to connect to mongdb replset? (`mongo admin -u admin -p $MONGODB_ADMIN_PASSWORD --host "${MONGODB_REPLICA_NAME}/$(hostname -f)" --eval "quit()"`)

This really tests that `mongod` is running and reachable part of replicaset.

[test]

3.2/root/usr/share/container-scripts/mongodb/init-petset-replset.sh
omron93: By the way: I figured out that this change is necessary for #224 (because Docker does not support SRV DNS records).

As we started to talk about this: @jim-minter Do you think this change might cause some troubles?

omron93:
[test]

bparees:
@jim-minter looks like there's one last comment/question from @omron93 above.

jim-minter:
> Would not be better in readinessProbe to try to connect to mongdb replset?

@omron93, I think probably not, because the mongo client is specially designed to handle non-responsive server endpoints.  I don't think that altering the readiness post instantiation really serves any purpose.

omron93:
> I think probably not, because the mongo client is specially designed to handle non-responsive server endpoints. I don't think that altering the readiness post instantiation really serves any purpose.

I would help in case that local server became inaccessible. So it would allow transition from ready to unready.  But hopefully this should not happend without `mongod` crashing.

Also readinessProbe would "fail" if PRIMARY is not elected ~ replicaset is not accessible : some problem occurred, or after a big scale down majority of members is not running.
It this how readiness should work?

But maybe it is too complicated and I'm not sure if it could cause some troubles in Kubernetes.

LGTM, thanks @jim-minter.

Issue #228, opened 2017-02-21
Replication support


omron93:
Replication using `ReplicaSet` was described in https://github.com/sclorg/mongodb-container/tree/master/2.4/examples/replica . 2.4 is now deprecated, so it is not documented nowhere. All other versions contains ReplicaSet code.

Replication using `StatefulSet` was added only for 3.2.

@bparees Is is possible to remove `ReplicaSet` code and add `StatefulSet` support also for 2.6 and 3.0-upg? Or `ReplicaSet` should still be a way how to deploy replication for <3.2 versions?

bparees:
yes we can do that.

PR #230, opened 2017-03-09
Remove ReplicaSet code

omron93:
As described in #228, OpenShift replication using ReplicaSet is not supported.
So this PR
- removes old implementations (code for ReplicaSet) and add new one (targeting StatefulSet) support for 2.6 and 3.0-upg versions
(Basically it is a copy of files for 3.2 - only two env. variables were added for older storage engine)
- adds OpenShift replication tests for 2.6, 3.0-upg versions [test-openshift]
- updates description of run-mongod-pet scripts (removed notes about differences with run-mongod-replication)
- removes unused functions from common.sh file

@bparees Please take a look.

Also what about keeping PetSet in a names? For me scripts are general (I hope used in #224 too) and basic thing what they do is mongodb replication (that they use Kubernetes PetSet is implementation detail). No other different implementation exists in this repo, so not reason to distinguish.
-> so old names would fit better IMPOV (`run-mongod-replication`, `init-replset.sh`) and maybe keeping old `run-mongod-pet` as symlink for compatibility. @bparees Thoughts?


omron93:
[test-openshift]

omron93:
Replication test failed because Persistent volume was reused but content was not removed - also mentioned in https://github.com/sclorg/mongodb-container/pull/222#issuecomment-283879717. Added hack to fix this (fixes it in CentOS CI where test is run in new environment every time, but probably not for "local" machines).

[test-openshift]

bparees:
renaming the scripts is fine, just don't change the args (or include backwards compatibility for the args) since existing templates reference them, eg:
https://github.com/sclorg/mongodb-container/blob/master/examples/petset/mongodb-petset-persistent.yaml#L126

omron93:
> renaming the scripts is fine, just don't change the args (or include backwards compatibility for the args) since existing templates reference them, eg:
https://github.com/sclorg/mongodb-container/blob/master/examples/petset/mongodb-petset-persistent.yaml#L126

Ok. Renamed. @bparees Generally, backward compatibility (e.g. symlinks with old names) have to be kept forever, right?

[test-openshift]

omron93:
@bparees Please review and merge.

bparees:
>  Generally, backward compatibility (e.g. symlinks with old names) have to be kept forever, right?

approximately forever, yes :(

bparees:
lgtm

omron93:
Squashed

[test-openshift]

omron93:
Tests are passing and LGTM from @bparees . Merging.

PR #231, opened 2017-03-13
Update default config file to YAML format

### LINKED ISSUE #197: 3.2: consider updating default config file to YAML format
Quote from https://docs.mongodb.com/v3.2/reference/configuration-options/:

> Changed in version 2.6: MongoDB 2.6 introduces a YAML-based configuration file format. The 2.4 configuration file format remains for backward compatibility.

I'm suggesting to update default config file to use YAML format.

omron93:
Update default config file to YAML format which was introduced in MongoDB 2.6.

Few changes:
 - HTTP interface is disable by default from 2.6 (https://docs.mongodb.com/manual/core/security-mongodb-configuration/), so no option about this is necessary 
 - In new format, there are no options "no" prefix (eg noprealloc) -> `MONGODB_NOPREALLOC` have to be negationed

@bparees Please take a look.

Resolves: #197

### DIFF
--- a/2.6/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/common.sh
@@ -10,7 +10,11 @@ set -o pipefail
 export MONGODB_DATADIR=/var/lib/mongodb/data
 export CONTAINER_PORT=27017
 # Configuration settings.
-export MONGODB_NOPREALLOC=${MONGODB_NOPREALLOC:-true}
+if [[ "${MONGODB_NOPREALLOC:-}" == "false" ]]; then
+  export MONGODB_PREALLOC=${MONGODB_PREALLOC:-true}
+else
+  export MONGODB_PREALLOC=${MONGODB_PREALLOC:-false}
+fi
 export MONGODB_SMALLFILES=${MONGODB_SMALLFILES:-true}
 export MONGODB_QUIET=${MONGODB_QUIET:-true}
 
--- a/2.6/test/run
+++ b/2.6/test/run
@@ -95,10 +95,10 @@ function test_mongo() {
     echo "  Testing MongoDB"
     if [ -v ADMIN_PASS ]; then
         echo "  Testing Admin user privileges"
-        mongo_admin_cmd "db=db.getSiblingDB('${DB}');db.removeUser('${USER}');"
-        mongo_admin_cmd "db=db.getSiblingDB('${DB}');db.addUser({user:'${USER}',pwd:'${PASS}',roles:['readWrite','userAdmin','dbAdmin']});"
+        mongo_admin_cmd "db=db.getSiblingDB('${DB}');db.dropUser('${USER}');"
+        mongo_admin_cmd "db=db.getSiblingDB('${DB}');db.createUser({user:'${USER}',pwd:'${PASS}',roles:['readWrite','userAdmin','dbAdmin']});"
         mongo_admin_cmd "db=db.getSiblingDB('${DB}');db.testData.insert({x:0});"
-        mongo_cmd "db.addUser({user:'test_user2',pwd:'test_password2',roles:['readWrite']});"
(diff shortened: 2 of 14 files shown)


omron93:
[test-openshift]

omron93:
[test][test-openshift]

omron93:
Fixed problem, that 2.6 has only one storage engine (so mmapv1 options are not separated). Please review...
[test-openshift]

2.6/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -10,7 +10,11 @@ set -o pipefail
 export MONGODB_DATADIR=/var/lib/mongodb/data
 export CONTAINER_PORT=27017
 # Configuration settings.
-export MONGODB_NOPREALLOC=${MONGODB_NOPREALLOC:-true}
+if [[ ${MONGODB_NOPREALLOC:-true} == "true" ]]; then
+  export MONGODB_PREALLOC="false"
+else
+  export MONGODB_PREALLOC="true"
+fi
```
bparees: you should probably surround this block w/ a check to see if MONGODB_PREALLOC is set and if so, respect that value directly.

And then update the image doc to deprecate MONGODB_NOPREALLOC in preference to MONGODB_PREALLOC

2.6/test/run
```diff
@@ -135,16 +136,16 @@ function test_config_option() {
...
 }
 
 function run_configuration_tests() {
     echo "  Testing image configuration settings"
-    test_config_option MONGODB_NOPREALLOC noprealloc true
-    test_config_option MONGODB_SMALLFILES smallfiles true
-    test_config_option MONGODB_QUIET quiet true
+    test_config_option MONGODB_NOPREALLOC true preallocDataFiles false
+    test_config_option MONGODB_SMALLFILES true smallFiles true
+    test_config_option MONGODB_QUIET true quiet true
```
bparees: and then add a test for the new env variable here.

bparees (changes requested):

2.6/root/usr/share/container-scripts/mongodb/common.sh
omron93: @bparees Thanks.

Changed a way how the default is computed. `MONGODB_PREALLOC` is always preferred. If it is not set, default is get according to `MONGODB_NOPREALLOC`.

omron93:
Please review.

2.6/test/run
```diff
@@ -135,15 +135,15 @@ function test_config_option() {
...
 
     docker stop $(get_cid ${name})
 }
 
 function run_configuration_tests() {
     echo "  Testing image configuration settings"
-    test_config_option MONGODB_NOPREALLOC noprealloc true
-    test_config_option MONGODB_SMALLFILES smallfiles true
+    test_config_option MONGODB_PREALLOC preallocDataFiles false
+    test_config_option MONGODB_SMALLFILES smallFiles true
```
bparees: should keep the test for NOPREALLOC also since it is technically still supported.

bparees (changes requested):

2.6/test/run
omron93: @bparees Thanks, testing also `MONGODB_NOPREALLOC` now.

bparees:
lgtm

omron93:
@bparees While inspecting differences among test scripts, I noticed that 2.6 uses deprecated user management functions. (`addUser` renamed to `createUser` and `removeUser` to `dropUser`) -> this is really small change (and no functionality is changed), so I've added it to this PR.

omron93:
[test-openshift]

omron93:
[test][test-openshift]

omron93:
Squashed. [test-openshift]

omron93:
Tests are passing. Merging.

@bparees Thanks for review.

Issue #32, opened 2015-05-26
Make the _USER and _PASSWORD optional when _ADMIN_PASSWORD is set


mfojtik:
When the `_ADMIN_PASSWORD` is set (required now in MongoDB) we should make the the _USER and _PASSWORD optional as the user already gets the `admin` account created. When the admin account is created, then we can't force users to also create regular user account as they can use the admin account to create more users.
Also to keep this backward compatible, we can make _USER and _PASSWORD optional for the case that users want to do full bootstrapping.

This was discussed with database guys (@praiskup, @hhorak) already and it makes sense to me as well.

CC @bparees @jhadvig @mnagy @soltysh

soltysh:
:+1: makes sense, if you have an admin user, there's no point in forcing you to create one at start, you can but are not obligated to do so.

mnagy:
What about the database, can it be optional as well?

Also, I think this could also be applicable to MySQL and PostgreSQL images as well.

mfojtik:
@mnagy can you please look at this?

bparees:
@mnagy I suppose DB name can be optional also..if you specify no DB name, we do not initialize any DB for you (which is also going to mean we don't create any admin/user accounts, right?)

@mfojtik what do you think?

omron93:
> I suppose DB name can be optional also..if you specify no DB name, we do not initialize any DB for you (which is also going to mean we don't create any admin/user accounts, right?)

@bparees For example in MongoDB admin user is stored in different database than DB user. So you don't have to create _USER in _DATABASE to have usable admin user.

mnagy:
I think I'll have to look into this post-3.0 for now.

PR #232, opened 2017-03-28
Make only admin password mandatory.

omron93:
User have to only specify `MONGODB_ADMIN_PASSWORD` and optionally provide settings for user with 'readWrite' role.

Resolves: #32

### DIFF
--- a/2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
@@ -43,7 +43,7 @@ function initiate() {
 
   info "Creating MongoDB users ..."
   mongo_create_admin
-  mongo_create_user "-u admin -p ${MONGODB_ADMIN_PASSWORD}"
+  [[ -v CREATE_USER ]] && mongo_create_user "-u admin -p ${MONGODB_ADMIN_PASSWORD}"
 
   info "Successfully initialized replica set"
 }
--- a/2.6/test/run
+++ b/2.6/test/run
@@ -45,6 +45,42 @@ function mongo_admin_cmd() {
     docker run --rm $IMAGE_NAME mongo admin --host $CONTAINER_IP -u admin -p "$ADMIN_PASS" --eval "${@}"
 }
 
+# Make sure the invocation of docker run fails.
+function assert_container_creation_fails() {
+
+  # Time the docker run command. It should fail. If it doesn't fail,
+  # container will keep running so we kill it with SIGKILL to make sure
+  # timeout returns a non-zero value.
+  set +e
+  timeout -s 9 --preserve-status 10s docker run --rm "$@" $IMAGE_NAME
+  ret=$?
+  set -e
+
+  # Timeout will exit with a high number.
+  if [ $ret -gt 30 ]; then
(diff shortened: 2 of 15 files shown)


omron93:
[test-openshift]

omron93:
[test]

omron93:
Changes applied to all versions. [test-openshift]

@bparees Please review.

2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
```diff
@@ -43,7 +43,7 @@ function initiate() {
 
   info "Creating MongoDB users ..."
   mongo_create_admin
-  mongo_create_user "-u admin -p ${MONGODB_ADMIN_PASSWORD}"
+  [[ -v CREATE_USER ]] && mongo_create_user "-u admin -p ${MONGODB_ADMIN_PASSWORD}"
```
bparees, line 46: if the admin password is required, why are you not always creating the admin user here?

2.6/test/run
```diff
@@ -45,6 +45,42 @@ function mongo_admin_cmd() {
     docker run --rm $IMAGE_NAME mongo admin --host $CONTAINER_IP -u admin -p "$ADMIN_PASS" --eval "${@}"
 }
 
+# Make sure the invocation of docker run fails.
+function assert_container_creation_fails() {
+
+  # Time the docker run command. It should fail. If it doesn't fail,
+  # postgresql will keep running so we kill it with SIGKILL to make sure
```
bparees: postgresql?

bparees (changes requested):
what's the motivation for this change?

edit: nm, i see the issue discussion.

2.6/test/run
omron93: Sorry, it was copied from postgresql-container. Fixed.

2.6/root/usr/share/container-scripts/mongodb/init-replset.sh
omron93: `mongo_create_admin` line creates admin user.

`mongo_create_user` creates "normal" user (passed string contains parameters for `mongo` shell - needs to know how to login)

omron93:
> what's the motivation for this change?
edit: nm, i see the issue discussion.

@bparees  As noted there, other databases also do not require both users [1][2].

In mongodb admin user is not created by default, so it has to be created (otherwise auth have to be disabled and server restarted to allow admin changes). Normal user might not be necessary, also number of env vars needed for running container (for example for testing) is lower :-) 

[1] https://github.com/sclorg/postgresql-container/blob/master/9.5/root/usr/share/container-scripts/postgresql/common.sh#L34
[2] https://github.com/sclorg/mariadb-container/blob/master/10.1/root/usr/share/container-scripts/mysql/validate-variables.sh#L1

[test-openshift]

bparees:
lgtm

omron93:
Squashed and rebased.

omron93:
[test-openshift]

omron93:
[test][test-openshift]

omron93:
LGTM from @bparees and tests are passing. So merging.

Issue #99, opened 2015-10-29
Test DATADIR permissions and fail cleanly


markllama:
When running this image in Kubernetes 1.1.0, the USER clause and the ownership/permissions imposed on the volume _/var/lib/mongodb/data_ by kubernetes are incompatible. If we check the permissions before starting the daemon we can provide a clear error message and avoid waiting/thrashing.
### Process and File Ownership/Permissions

```
process: uid=184(mongodb) gid=998(mongodb) groups=998(mongodb)
directory: drwxr-x---. 2 root root 6 Oct 29 13:49 /var/lib/mongodb/data
```

This causes the intialization to fail when creating a lock file.  The script proceeds to wait for the DB to become available and eventually times out.  The actual error is buried in the logs.
### Log Example

```
...
Thu Oct 29 13:57:59.288 [initandlisten] exception in initAndListen std::exception: boost::filesystem::status: Permission denied: "/var/lib/mongodb/data/mongod.lock", terminating
Thu Oct 29 13:57:59.288 dbexit: 
Thu Oct 29 13:57:59.288 [initandlisten] shutdown: going to close listening sockets...
...
Thu Oct 29 13:57:59.289 dbexit: really exiting now
 => Waiting for MongoDB service startup ...
=> Waiting for MongoDB service startup  ...
=> Waiting for MongoDB service startup  ...
... 
=> Waiting for MongoDB service startup  ...
=> Giving up: Failed to start MongoDB service!
```

**NOTE:** In OpenShift this is not observed because OpenShift imposes a high numbered UID and GID 0.  The effect is that the mongodb process can write to the volume even though it has _root:root_ ownership.

Upstream changes to Kubernetes, allowing Security Contexts to impose non-root groups on both processes and filesystems will allow this container to run as expected.  It still would be nice to check and report if the DATADIR is not writable.

rhcarvalho:
FTR, #190 added a similar check to the keyfile creation flow. Unfortunately, that only applies the case where we do replication (and not the standalone use case).

PR #233, opened 2017-03-29
Test DATADIR permissions and fail cleanly.

omron93:
Test permissions of `MONGODB_DATADIR` and if it is not writable (or SELinux deny it), print error message and exit cleanly.

Also removes creating of SCL rpm package default directory (`/var/opt/rh/rh-mongodb26/lib/mongodb`,...) and changing of permissions. These directories are created by rpm, so it is not necessary to create them. Also they are not used anywhere, so working with them in Dockerfiles is redundant.

Resolves: #99

### DIFF
--- a/2.6/Dockerfile
+++ b/2.6/Dockerfile
@@ -27,9 +27,8 @@ RUN yum install -y centos-release-scl && \
     yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all && \
-    mkdir -p /var/lib/mongodb/data && chown -R mongodb.0 /var/lib/mongodb/ /var/opt/rh/rh-mongodb26/lib/mongodb && \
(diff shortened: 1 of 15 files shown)


omron93:
[test-openshift]
@bparees Please review.

bparees:
lgtm

omron93:
Thanks. One more thing I would like to ask:

I've realized, that if system doesn't have SELinux support (no SELinux labels stored in file system) the command `stat -c '%A owned by %u:%g, SELinux: %C' "${MONGODB_DATADIR}"` fails.

For example output for `/var` is `drwxr-xr-x owned by 0:0, SELinux: stat: failed to get security context of '/var': No data available` and command exits with $?=1.

It would not cause problems because next command is `exit 1`, but @bparees is this a problem?

(I've added info about SELinux context, because SELinux causes problems)

bparees:
> It would not cause problems because next command is exit 1, but @bparees is this a problem?

i think it's ok.  certainly for openshift we require selinux anyway.

omron93:
Thanks. So merging this PR.

PR #234, opened 2017-04-03
Set cacheSizeGB for 3.2 image + add cgroup-limits script into images.

### LINKED ISSUE #204: Set cacheSizeGB for 3.2 image
According to the docs:

> If you run mongod in a container (e.g. lxc, cgroups, Docker, etc.) that does not have access to all of the RAM available in a system, you must set storage.wiredTiger.engineConfig.cacheSizeGB to a value less than the amount of RAM available in the container. The exact amount depends on the other processes running in the container.

(https://docs.mongodb.com/manual/faq/storage/#to-what-size-should-i-set-the-wiredtiger-internal-cache)

We currently do not set that configuration option.

omron93:
- add `cgroup-limits` scripts to check amount of available memory to all supported images.
- add function to set WT cache size to default in configuration file for images which supports wiredTiger
- set WT cache size for MongoDB 3.2, where wiredTiger is a default storage engine.

@bparees Please take a look.

Fixes #204

### DIFF
--- a/2.6/Dockerfile.rhel7
+++ b/2.6/Dockerfile.rhel7
@@ -33,16 +33,18 @@ RUN yum install -y yum-utils && \
     INSTALL_PKGS="bind-utils gettext iproute rsync tar v8314 rh-mongodb26-mongodb rh-mongodb26" && \
     yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
-    yum clean all && \
-    mkdir -p /var/lib/mongodb/data && chown -R mongodb.0 /var/lib/mongodb/ && \
-    # Loosen permission bits to avoid problems running container with arbitrary UID
-    chmod -R g+rwx /var/lib/mongodb
-
-VOLUME ["/var/lib/mongodb/data"]
+    yum clean all
 
 ADD root /
 
 # Container setup
-RUN touch /etc/mongod.conf && chown mongodb:0 /etc/mongod.conf && /usr/libexec/fix-permissions /etc/mongod.conf
+RUN touch /etc/mongod.conf && \
+    mkdir -p ${HOME}/data && \
+    chown -R mongodb:0 /etc/mongod.conf ${CONTAINER_SCRIPTS_PATH}/mongodb.conf.template ${HOME} && \
+    # Set g+rw(x) permission for group 0 - to avoid problems running container with arbitrary UID
+    /usr/libexec/fix-permissions "/etc/mongod.conf ${CONTAINER_SCRIPTS_PATH}/mongodb.conf.template \
+    ${HOME}"
+
+VOLUME ["/var/lib/mongodb/data"]
 
 USER 184
--- a/2.6/Dockerfile.rhel7
+++ b/2.6/Dockerfile.rhel7
@@ -40,10 +40,9 @@ ADD root /
 # Container setup
 RUN touch /etc/mongod.conf && \
     mkdir -p ${HOME}/data && \
-    chown -R mongodb:0 /etc/mongod.conf ${CONTAINER_SCRIPTS_PATH}/mongodb.conf.template ${HOME} && \
-    # Set g+rw(x) permission for group 0 - to avoid problems running container with arbitrary UID
-    /usr/libexec/fix-permissions "/etc/mongod.conf ${CONTAINER_SCRIPTS_PATH}/mongodb.conf.template \
-    ${HOME}"
+    # Set owner 'mongodb:0' and 'g+rw(x)' permission - to avoid problems running container with arbitrary UID
+    /usr/libexec/fix-permissions /etc/mongod.conf ${CONTAINER_SCRIPTS_PATH}/mongodb.conf.template \
+    ${HOME}
 
 VOLUME ["/var/lib/mongodb/data"]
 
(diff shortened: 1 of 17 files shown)


omron93:
Rebased. [test][test-openshift]

3.0-upg/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -181,6 +181,21 @@ function setup_default_datadir() {
     echo >&2 "DETAILS: directory permissions: $(stat -c '%A owned by %u:%g, SELinux: %C' "${MONGODB_DATADIR}")"
     exit 1
   fi
+
+# setup_wiredtiger_cache checks amount of available RAM (it has to use cgroups in container)
+# and if there are any memory restrictions set storage.wiredTiger.engineConfig.cacheSizeGB
+# in MONGODB_CONFIG_PATH to upstream default size
+# it is inteded to update mongodb.conf.template, with custom config file it might create conflict
```
bparees: s/inteded/intended/

3.2/root/usr/bin/run-mongod
```diff
@@ -30,6 +30,7 @@ trap 'cleanup' SIGINT SIGTERM
 if [ ! -s $MONGODB_CONFIG_PATH ]; then
   # Generate config file for MongoDB
   envsubst < ${CONTAINER_SCRIPTS_PATH}/mongodb.conf.template > $MONGODB_CONFIG_PATH
+  setup_wiredtiger_cache
```
bparees: since you say it's unsafe to do this on anything other than the template file, it would be better to have it operate on the template file... run this step before the envsubst instead of after, and have it operate on the template file (perhaps passed as an argument to the function).

i know technically this is safe as is since we're inside the `if [ ! -s $file ]` block, but i think my suggestion would make it more robust against future refactoring.

3.2/test/run
```diff
@@ -341,6 +341,39 @@ run_doc_test() {
...
+    ADMIN_PASS=${admin_password}
+
+    # minimum is 1G
+    create_container "${name}_200M" ${DOCKER_ARGS} -m 200M
+    test_connection ${name}_200M
+    mongo_admin_cmd "if (db.serverStatus()['wiredTiger']['cache']['maximum bytes configured'] == Math.pow(2,30)){quit(0)}; quit(1)"
+    # if greater that 1G, use 60% of (RAM - 1G)
+    create_container "${name}_5G" ${DOCKER_ARGS} -m 5G
+    test_connection ${name}_5G
+    mongo_admin_cmd "if (db.serverStatus()['wiredTiger']['cache']['maximum bytes configured'] == 2*Math.pow(2,30)){quit(0)}; quit(1)"
```
bparees: this seems to be saying you expect the cache size to be 2 gigs, but the comment above implies it should be 60% of 4gigs(?), which is more than 2 gigs.  is it being rounded off?

omron93: You are, right. Thanks.

omron93: >  is it being rounded off?

Yes, it is - but this wasn't the case. In MongoDB<3.4 WT cache size is integer. So in `setup_wiredtiger_cache` it is rounded down.

3.2/root/usr/bin/run-mongod
omron93: I was modifying template file, but it didn't work - template file is owned by root. Also it is `/usr/share/container-scripts/mongodb/`.

Ok, I modified Dockerfiles to change permissions of template file.

3.0-upg/root/usr/share/container-scripts/mongodb/common.sh
omron93: Thanks.

omron93:
[test-openshift]

2.6/Dockerfile.rhel7
```diff
@@ -33,16 +33,18 @@ RUN yum install -y yum-utils && \
...
 
 ADD root /
 
 # Container setup
-RUN touch /etc/mongod.conf && chown mongodb:0 /etc/mongod.conf && /usr/libexec/fix-permissions /etc/mongod.conf
+RUN touch /etc/mongod.conf && \
+    mkdir -p ${HOME}/data && \
+    chown -R mongodb:0 /etc/mongod.conf ${CONTAINER_SCRIPTS_PATH}/mongodb.conf.template ${HOME} && \
+    # Set g+rw(x) permission for group 0 - to avoid problems running container with arbitrary UID
+    /usr/libexec/fix-permissions "/etc/mongod.conf ${CONTAINER_SCRIPTS_PATH}/mongodb.conf.template \
```
bparees: why not just a straight chmod/chgrp?

omron93:
[test][test-openshift]

2.6/Dockerfile.rhel7
omron93: I thought `fix-permissions` is a standard way how to change permissions :-)

I updated it to also change the owner to `mongodb` (similar to how it works in postgresql-container) and removed chown in `Dockerfile`.

omron93:
@bparees All comments are fixed.

[test-openshift]

omron93:
Rebased. [test-openshift]

2.6/Dockerfile.rhel7
bparees: fix-permissions was originally written to modify source files to ensure they would be writable by the container user so they could make hot deploy changes w/o building a new image.

i don't even recall how it ended up in the mongodb image.  I guess doing this is ok, i'd just worry that at some point the fix-permissions script behavior may diverge from the behavior needed here.  but i'm not going to stand in the way.

bparees:
lgtm

omron93:
Thanks. So merging.

Issue #235, opened 2017-04-05
erroneous error message when specifying admin password and a database name


elmiko:
while testing out this image i am running into a problem where the error output from the container does not match the behavior i am seeing.

## steps to reproduce:

1. create a new container with the following command

```
oc new-app centos/mongodb-32-centos7 \
-e MONGODB_ADMIN_PASSWORD=admin \
-e MONGODB_DATABASE=foo \
--name mongodb
```

this results in a pod that will fail with the following error log:

```
You must specify the following environment variables:
  MONGODB_ADMIN_PASSWORD
Optionally you can provide settings for user with 'readWrite' role:
  MONGODB_USER
  MONGODB_PASSWORD
  MONGODB_DATABASE
MongoDB settings:
  MONGODB_QUIET (default: true)
```

i am positive that the environment variable is set, but i am informed that the true error is that you cannot specify the admin password **and** a database name.

this is the result of running the new-app command with the `-o yaml` option:

```
apiVersion: v1
items:
- apiVersion: v1
  kind: ImageStream
  metadata:
    annotations:
      openshift.io/generated-by: OpenShiftNewApp
    creationTimestamp: null
    labels:
      app: mongodb
    name: mongodb
  spec:
    tags:
    - annotations:
        openshift.io/imported-from: centos/mongodb-32-centos7
      from:
        kind: DockerImage
        name: centos/mongodb-32-centos7
      generation: null
      importPolicy: {}
      name: latest
      referencePolicy:
        type: ""
  status:
    dockerImageRepository: ""
- apiVersion: v1
  kind: DeploymentConfig
  metadata:
    annotations:
      openshift.io/generated-by: OpenShiftNewApp
    creationTimestamp: null
    labels:
      app: mongodb
    name: mongodb
  spec:
    replicas: 1
    selector:
      app: mongodb
      deploymentconfig: mongodb
    strategy:
      resources: {}
    template:
      metadata:
        annotations:
          openshift.io/generated-by: OpenShiftNewApp
        creationTimestamp: null
        labels:
          app: mongodb
          deploymentconfig: mongodb
      spec:
        containers:
        - env:
          - name: MONGODB_ADMIN_PASSWORD
            value: admin
          - name: MONGODB_DATABASE
            value: ophicleide
          image: centos/mongodb-32-centos7
          name: mongodb
          ports:
          - containerPort: 27017
            protocol: TCP
          resources: {}
          volumeMounts:
[...]

PR #236, opened 2017-04-06
Move checkinkg of variables and usage into common.sh

omron93:
Move checkinkg of variables and usage into common.sh + prints more detailed error message

Adds validation of database name - https://docs.mongodb.com/manual/reference/limits/

Using similar approach as postgresql-container (https://github.com/sclorg/postgresql-container/blob/master/9.5/root/usr/share/container-scripts/postgresql/common.sh#L58)

Resolves: #235

### DIFF
--- a/2.6/root/usr/share/container-scripts/mongodb/common.sh
+++ b/2.6/root/usr/share/container-scripts/mongodb/common.sh
@@ -183,6 +183,65 @@ function setup_default_datadir() {
   fi
 }
 
+# check_env_vars checks environmental variables
+# if variables to create non-admin user are provided, sets CREATE_USER=1
+# if REPLICATION variable is set, checks also replication variables
+function check_env_vars() {
+  local readonly database_regex='^[^/\. "$]*$'
+
+  [[ -v MONGODB_ADMIN_PASSWORD ]] || usage "MONGODB_ADMIN_PASSWORD has to be set."
+
+  if [[ -v MONGODB_USER || -v MONGODB_PASSWORD || -v MONGODB_DATABASE ]]; then
+    [[ -v MONGODB_USER && -v MONGODB_PASSWORD && -v MONGODB_DATABASE ]] || usage "You have to set all or none of variables: MONGODB_USER, MONGODB_PASSWORD, MONGODB_DATABASE"
+
+    [[ "${MONGODB_DATABASE}" =~ $database_regex ]] || usage "Database name must match regex: $database_regex"
+    [ ${#MONGODB_DATABASE} -le 63 ] || usage "Database name too long (maximum 63 characters)"
+
+    export CREATE_USER=1
+  fi
+
+  if [[ -v REPLICATION ]]; then
+    [[ -v MONGODB_KEYFILE_VALUE ]] || usage "MONGODB_KEYFILE_VALUE has to be set"
+    [[ -v MONGODB_REPLICA_NAME ]] ||  usage "MONGODB_REPLICA_NAME has to be set"
+  fi
+}
+
+# usage prints info about required enviromental variables
+# if $1 is passed, prints error message containing $1
+# if REPLICATION variable is set, prints also info about replication variables
+function usage() {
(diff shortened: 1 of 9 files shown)


2.6/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -183,6 +183,65 @@ function setup_default_datadir() {
...
+
+    [[ "${MONGODB_DATABASE}" =~ $database_regex ]] || usage "Database name must match regexp: $database_regexp"
+    [ ${#MONGODB_DATABASE} -le 63 ] || usage "Database name too long (maximum 63 characters)"
+
+    CREATE_USER=1
+  fi
+
+  if [[ -v REPLICATION ]]; then
+    [[ -v MONGODB_KEYFILE_VALUE ]] || usage "MONGODB_KEYFILE_VALUE has to be set"
+    [[ -v MONGODB_REPLICA_NAME ]] ||  usage "MONGODB_REPLICA_NAME has to be set"
```
bparees: i think this would be better as a single error so users don't enable replication, get the first error, set that variable, and then get the second error.

2.6/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -183,6 +183,65 @@ function setup_default_datadir() {
...
+# if REPLICATION variable is set, prints also info about replication variables
+function usage() {
+  if [ $# == 1 ]; then
+    echo >&2 "error: $1"
+  fi
+
+  echo "
+You must specify the following environment variables:
+  MONGODB_ADMIN_PASSWORD
+Optionally you can provide settings for user with 'readWrite' role:
```
bparees: s/user/a user/

2.6/root/usr/share/container-scripts/mongodb/common.sh
omron93: Ok, checked together.

2.6/root/usr/share/container-scripts/mongodb/common.sh
omron93: Thanks.

bparees:
lgtm

omron93:
[test-openshift]

2.6/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -183,6 +183,64 @@ function setup_default_datadir() {
...
+# if REPLICATION variable is set, prints also info about replication variables
+function usage() {
+  if [ $# == 1 ]; then
+    echo >&2 "error: $1"
+  fi
+
+  echo "
+You must specify the following environment variables:
+  MONGODB_ADMIN_PASSWORD
+Optionally you can provide settings for a user with 'readWrite' role:
```
elmiko, line 219: is there a way to easily specify that all of these are needed in this message?

elmiko (commented):
overall lgtm, i have one suggestion

2.6/root/usr/share/container-scripts/mongodb/common.sh
omron93: Do you have any suggestion?

elmiko: sure, how about something like this:

```
Optionally you can provide settings for a user with 'readWrite' role:
(Note you MUST specify all three of these settings)
...
```

omron93:
Rebased and added suggested changes.
[test-openshift]

@bparees @elmiko Please take a look again.

bparees:
lgtm but seems to have some test issues

elmiko:
aside from the test issues, lgtm. thanks for incorporating my suggestion =)

omron93:
Test failure was done by some networking issue (I saw this error for the first time, so I think it isn't important. If it occurs more often I'll look deeper into it). Rerunning this test fixed this.

Merging this PR. @bparees @elmiko Thanks for review and suggestions.

PR #237, opened 2017-04-07
Fix creation of user in run-mongod-replication

omron93:
In `run-mongod-replication` `CREATE_USER` was not exported, so it can't be used in `init-replset.sh`.

[test-openshift]

Fixes: https://github.com/openshift/origin/issues/13659#issuecomment-292263840


omron93:
[test-openshift]

PR #239, opened 2017-04-13
Add source-to-image support

omron93:
First implementation of s2i support to allow extending this image.

I'll add documentation on Thursday.

@hhorak @praiskup @pkubatrh Please take a look and comment.


omron93:
[test-openshift]

bparees:
where's the assemble script?

3.2/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -253,6 +174,26 @@ within the container or visit https://github.com/sclorgk/mongodb-container/."
...
+  shopt -s nullglob
+  for filename in $(echo $custom_dir/*.sh $default_dir/*.sh | xargs -r basename -a | sort | uniq); do
+    # Custom file is prefered
+    if [ -f $custom_dir/$filename ]; then
+      source $custom_dir/$filename
+    else
+      source $default_dir/$filename
+    fi
+  done
+  shopt -s nullglob
```
pkubatrh: This should be `shopt -u nullglob`

3.2/Dockerfile
```diff
@@ -27,6 +28,9 @@ RUN yum install -y centos-release-scl && \
     rpm -V $INSTALL_PKGS && \
     yum clean all
 
+# Copy the S2I scripts from the specific language image to $STI_SCRIPTS_PATH.
+COPY s2i/bin/ $STI_SCRIPTS_PATH
```
pkubatrh: The `s2i/bin/` tree is missing from git

pkubatrh (changes requested):
Missing s2i/bin tree (probably just not added to the commit?) makes the tests fail at build stage.

3.2/root/usr/share/container-scripts/mongodb/common.sh
omron93: Thanks, fixed.

3.2/Dockerfile
omron93: You are right. Files are added.

3.2/root/usr/share/container-scripts/mongodb/README.md
```diff
@@ -88,3 +91,38 @@ the values stored in the variables and the actual passwords. Whenever a database
...
+---------------------------------
+This image can be extended using [source-to-image](https://github.com/openshift/source-to-image).
+
+For example to build customized MongoDB database image `my-mongodb-centos7` with configuration in `~/image-configuration/` run:
+
+```
+$ s2i build ~/image-configuration/ centos/mongodb-32-centos7 my-mongodb-centos7
+```
+
+The directory passed to `s2i build` can contain these directories:
```
hhorak: I'd be more specific here: "...should contain one or more of the following directories:"

3.2/root/usr/share/container-scripts/mongodb/init/50-update-users.sh
```diff
@@ -0,0 +1,25 @@
+# update_users creates default users (see usage)
+# if users are already created, updates passwords to match
+# environmental variables
```
hhorak: s/environmental/environment/

3.2/root/usr/share/container-scripts/mongodb/README.md
```diff
@@ -88,3 +91,38 @@ the values stored in the variables and the actual passwords. Whenever a database
...
+- `mongodb-cfg/`
+  - when running `run-mongod` or `run-mongod-replication` commands contained `mongod.conf` file is used for `mongod` configuration
+    - `envsubst` command is run on this file to still allow customization of the image using environmental variables
```
hhorak: s/environmental/environment/

3.2/root/usr/share/container-scripts/mongodb/pre-init/10-check-env-vars.sh
```diff
@@ -0,0 +1,23 @@
+# check_env_vars checks environmental variables
```
hhorak: s/environmental/environment/

3.2/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -253,6 +174,26 @@ within the container or visit https://github.com/sclorgk/mongodb-container/."
...
+# process_files process extending files in $1 and $2 directories
+# - source all *.sh files
+#   (if there are files with same name source only file from $1)
+function process_files() {
+  local custom_dir default_dir
+  custom_dir=$1
+  default_dir=$2
+
+  shopt -s nullglob
+  for filename in $(echo $custom_dir/*.sh $default_dir/*.sh | xargs -r basename -a | sort | uniq); do
```
hhorak: I'd rather use `find` command instead of `echo $dir | xargs...` e.g. `find $custom_dir $default_dir -maxdepth 1 -type f -name '*.sh' -printf "%f\n" | sort -u` should do similar thing I hope. It will also allow removing the `shopt -s nullglob` lines.

3.2/s2i/bin/assemble
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+shopt -s dotglob
+echo "---> Installing application source ..."
+ls -A /tmp/a/* &>/dev/null && mv /tmp/src/* ./
```
hhorak: I'm not sure whether I'm missing something or it is a typo -- shouldn't `ls -A /tmp/a/` be rather `ls -A /tmp/src/`?

3.2/root/usr/share/container-scripts/mongodb/README.md
```diff
@@ -88,3 +91,38 @@ the values stored in the variables and the actual passwords. Whenever a database
...
+    - custom configuration file does not affect name of replica set - it has to be set in `MONGODB_REPLICA_NAME` environment variable
+    - it is not possible to configure SSL using custom configuration file
+
+- `mongodb-pre-init/`
+  - contained shell scripts (`*.sh`) are sourced before `mongod` server is started
+
+- `mongodb-init/`
+  - contained shell scripts (`*.sh`) are sourced when `mongod` server is started
+    - in this phase `mongod` server don't have enabled authentication when running `run-mongod` command (for `run-mongod-replication` command configured users are already created and authentication is enabled)
+
```
hhorak: I think we should also include a simple example and typical use case for every directory. The use cases might be:

* changing a config option based on the environment variable (`mongodb-cfg`)
* checking that admin password is longer than 16 bytes (additional restrictions) (`mongodb-pre-init`)
* storing some initial documents (constants) into the database (`mongodb-init`)

These examples may be used for both, as a documentation and by the testing script.

hhorak: When sourcing the scripts, some variables are defined and sometimes users may need them in their extension scripts. For example `REPLICATION` to make things differently for replication use case and non-replication scenarios. I think we should be explicit about which variables users may use this way and consider all the other variables (even if available and set) to be an implementation detail only, so users don't depend on them. I'd start with most necessary variables only and add more when needed later.

hhorak (commented):
I've provided few in-line comments, but generally this looks very nice IMO. We should gather broader feedback now.

3.2/s2i/bin/usage
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+man -l /help.1
```
hhorak: The `/help.1` is meant to be formatted for `atomic` tool that shows the help file using `groff` tool. Using just pure `groff` would also limit dependencies, since `man` requires more deps. So, `man -l /help.1` may be replaced by `groff -man -ETascii /help.1`.

hhorak: `groff-base` should also be installed in the `Dockerfile`, so we make sure it is available.

3.2/root/usr/share/container-scripts/mongodb/common.sh
```diff
@@ -253,6 +174,26 @@ within the container or visit https://github.com/sclorgk/mongodb-container/."
   exit 1
 }
 
+# process_files process extending files in $1 and $2 directories
+# - source all *.sh files
+#   (if there are files with same name source only file from $1)
+function process_files() {
```
hhorak: `process_files` does not describe the function specifically enough, I'd rather use something like `process_extending_files` as a function name.

3.2/root/usr/share/container-scripts/mongodb/common.sh
hhorak: To really support spaces in filename (if some crazy man does it), we might do something like this:
`find "$custom_dir" "$default_dir" -maxdepth 1 -type f -name '*.sh' -printf "%f\n" | sort -u | while read filename ; do ...`

3.2/s2i/bin/usage
pkubatrh: +1 for using groff instead of man. However we should use `groff -t -man -ETascii /help.1` to properly format any tables located in the help file.

3.2/root/usr/share/container-scripts/mongodb/README.md
omron93: Thanks.

3.2/root/usr/share/container-scripts/mongodb/init/50-update-users.sh
omron93: Thanks.

3.2/root/usr/share/container-scripts/mongodb/README.md
omron93: Thanks.

3.2/root/usr/share/container-scripts/mongodb/pre-init/10-check-env-vars.sh
omron93: Thanks.

3.2/s2i/bin/assemble
omron93: You are right. It is typo - fixed.

3.2/s2i/bin/usage
omron93: Thanks for suggestion. Now should be done this way.

3.2/root/usr/share/container-scripts/mongodb/common.sh
omron93: It is better. Renamed.

3.2/root/usr/share/container-scripts/mongodb/common.sh
omron93: When running for example `process_extending_files ${APP_DATA}/mongodb-pre-init/ ${CONTAINER_SCRIPTS_PATH}/pre-init/` the user provided directory doesn't have to exist.

With `shopt -s nullglob` `echo $custom_dir/*.sh` prints nothing in this case. But `find "$custom_dir"` returns 1 and prints error message.

@hhorak Would be using `find` with checking that directories exists better?

Difference is that in current implementation

3.2/root/usr/share/container-scripts/mongodb/README.md
omron93: OK. I'll add example app configuration. And it should be also used in image tests.

omron93: Added. Please review.

3.2/root/usr/share/container-scripts/mongodb/common.sh
hhorak: @omron93 I found this during mariadb PoC as well, this is how I solved it:
https://github.com/sclorg/mariadb-container/pull/32/files#diff-fe50a7af34bff8099d244f9c0cd56273R177
(I can't say that I'm very proud of running similar find command twice, but it works)

omron93:
Rebased.

Changes:
- added example and using it in tests
- fixed `process_extending_files` to allow env vars modification
  - in old implementation while loop was run in subshell so no env vars chages were possible
- `mongodb` user added to `root` group
- updated README and example README

@bparees @pkubatrh @hhorak Please take a look again.

3.2/s2i/bin/assemble
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+shopt -s dotglob
+echo "---> Installing application source ..."
+ls -A /tmp/src/* &>/dev/null && mv /tmp/src/* ./
```
bparees: what's the point of the ls?

omron93: To not to exit with mv error in case that no extending files are present.

omron93:
[test-openshift]

3.2/s2i/bin/assemble
pkubatrh: Would use `mv /tmp/src/* ./ 2>/dev/null || true` which is a bit more readable.

pkubatrh (approved):

3.2/s2i/bin/assemble
omron93: @bparees Are files in `/tmp/src/` always owned by current images user? (does `s2i build` always set this?)
If not, `mv /tmp/src/* ./ 2>/dev/null` could hide useful error message. But it is true that `|| true` is more readable.

Anyway, thanks for suggestions.

bparees: we always chmod the files to 666 as we layer them into the container, so permissions should not be a problem.

omron93: Changed to what Petr suggested.

omron93:
[test-openshift]

alikhajeh1:
Any update on this? It's needed for SNI according to https://github.com/sclorg/mongodb-container/issues/240

hhorak:
I think the SNI is good enough reason to not wait with this any more, let's merge it after rebasing.

bparees:
i haven't studied this PR lately but i just want to confirm that in addition to making it easy to provide config via an s2i build, this PR is also going to make it easier(possible) to inject a custom mongo config file into the container without building a new image?  (@hhorak is familiar w/ the problems w/ the current image on this front)

Namely by letting users customize the $MONGODB_CONFIG_PATH value (which is currently hardcoded into common.sh)?

hhorak:
> i haven't studied this PR lately but i just want to confirm that in addition to making it easy to provide config via an s2i build, this PR is also going to make it easier(possible) to inject a custom mongo config file into the container without building a new image? (@hhorak is familiar w/ the problems w/ the current image on this front)

This PR specifically documents using `-v` option, so users will be able to either mount scripts and config file together via `-v ./image-configuration/:/opt/app-root/src/` or only the config file directory by `-v ./image-configuration/:/opt/app-root/src/mongodb-src` -- the later case should work also well with ConfigMaps, since we mount whole directory, but we should proof it by working example first and document ConfigMaps support properly.

> Namely by letting users customize the $MONGODB_CONFIG_PATH value (which is currently hardcoded into common.sh)?

The original version of this PR did not allow to change `$MONGODB_CONFIG_PATH`. Although it might be also another way to change just the config file location (basically not change `$MONGODB_CONFIG_PATH` when defined by user), I'm not sure whether it would bring something more than the ability to mount config file to `/opt/app-root/src/mongodb-src` directly.

3.2/root/usr/share/container-scripts/mongodb/README.md
```diff
@@ -88,3 +91,45 @@ the values stored in the variables and the actual passwords. Whenever a database
...
+During `s2i build` all provided files are copied into `/opt/app-root/src` directory in the new image. If some configuration files are present in destination directory, files with the same name are overwritten. Also only one file with the same name can be used for customization and user provided files are preferred over default files in `/usr/share/container-scripts/mongodb/`- so it is possible to overwrite them.
+
+Same configuration directory structure can be used to customize the image every time the image is started using `docker run`. The directory have to be mounted into `/opt/app-root/src/` in the image (`-v ./image-configuration/:/opt/app-root/src/`). This overwrites customization built into the image.
+
+Mounting custom configuration file
+---------------------------------
+
+It is allowed to use custom configuration file for `mongod` server. To use it in container it has to be mounted into `/etc/mongod.conf`. For example to use configuration file stored in `/home/user` directory use this option for `docker run` command: `-v /home/user/mongod.conf:/etc/mongod.conf:Z`.
```
bparees: @hhorak this is either wrong, or isn't what you described re: configmap mounting (and won't work for mounting configmaps since as discussed, we can't mount a single file, we need to mount the entire /etc dir)

hhorak: @bparees you're referring to a different part of README.md than what I meant -- see my comment above, users will be able to mount whole directory into `/opt/app-root/src/mongodb-cfg` (so the config file will appear at `/opt/app-root/src/mongodb-cfg/mongodb.conf`.

This section, which you refer to, should be probably removed from the documentation (or just explicitly marked as obsolete way of changing config file and removed later).

hhorak: Btw. I asked @pvalena to look on real testing the ConfigMaps with this PR, not sure whether he already has some progress -- @pvalena, please, share your findings here once you have something.

pvalena: @hhorak, yes I am testing this PR [rebased](https://github.com/omron93/mongodb-container/pull/2) on master. I have already uncovered one issue: on several places `mongodb.conf` was used, opposed to `mongod.conf` (without `b`). Also .template file was named `mongodb.conf.template`. I have added commit to my PR.

pvalena:
I have tested current state of PRs (75ce8fd8f39556d5acd1de28305b99f47c0d977b):
 - build and run with docker on RHEL 7
 - mount and load custom mongod.conf (with `mode: preferSSL`) and mongodb.pem
 - I am now trying to proceed with ConfigMaps

```
$ docker run -p 27017:27017 --rm -it -e MONGODB_ADMIN_PASSWORD=password -v "`pwd`/mongodb-cfg:/opt/app-root/src/mongodb-cfg:Z" nmongo
[ . . . ]
2017-07-17T12:19:44.402+0000 I NETWORK  [initandlisten] waiting for connections on port 27017 ssl
2017-07-17T12:19:44.402+0000 I NETWORK  [HostnameCanonicalizationWorker] Starting hostname canonicalization worker
2017-07-17T12:21:29.132+0000 I NETWORK  [initandlisten] connection accepted from 172.17.0.1:55304 #1 (1 connection now open)
2017-07-17T12:58:04.933+0000 I NETWORK  [conn1] end connection 172.17.0.1:55304 (0 connections now open)
```

omron93:
Rebased.

I'll apply these changes to 2.6 and 3.0-pg versions too.
@hhorak @bparees Please review.

omron93:
[test-openshift]

omron93:
[test-openshift]

omron93:
Images from 744cb94 are accessible in
~~~
docker.io/mskalick/mongodb-26-centos7:pr239
docker.io/mskalick/mongodb-30-upg-centos7:pr239
docker.io/mskalick/mongodb-32-centos7:pr239
~~~

@pkubatrh @torsava Please review this PR.

pvalena:
ConfigMaps support this

2.6/s2i/bin/run
```diff
@@ -0,0 +1 @@
+/bin/run-mongod
```
bparees, line 1: should this take/respect args so you can start it w/ replication?

pvalena:
I can confirm via testing that creating ConfigMaps  in OpenShift and then mounting `mongod.conf` and `mongodb.pem` from ConfigMaps work as expected.

I have built container from this PR (744cb94eebe811e4a6d4f7d11a1058f553d63eb7), pushed it to local OpenShift cluster(registry) and successfully deployed the container with custom configuration, which enabled ssl (similar to https://github.com/sclorg/mongodb-container/pull/239#issuecomment-315753497).

Note that pushing a custom container to the registry required a 'magic' workaround for some unknown reason. I will investigate that further later on.

2.6/s2i/bin/run
omron93: @bparees What do you mean by this?

In majority of s2i based images `run` script is used as default image CMD - it is not a mongodb case (it has different default CMD). So I've added this link mainly for "compatibility" reasons and expected that it won't be used. Is the assumption wrong, is something using specially `/usr/libexec/run` ?

Or is this comment related to https://github.com/sclorg/mongodb-container/issues/246 ?

dongboyan77:
Hi, I test with image `docker.io/mskalick/mongodb-30-upg-centos7:pr239` on ocp 3.6 cluster, find 2 issues:
 1. after changing user name and password via env vars, still can login with old user name and password in new pod,
 2. petset-replicas, after removing all db pods, new pod cannot be running again, the expect result is all db pods are running again

`# oc logs -f mongodb-0
=> sourcing 10-check-env-vars.sh ...
=> sourcing 30-set-config-file.sh ...
=> sourcing 35-setup-default-datadir.sh ...
=> sourcing 40-setup-keyfile.sh ...
=> [Tue Jul 25 09:29:37] Waiting for local MongoDB to accept connections  ...
note: noprealloc may hurt performance in many applications
2017-07-25T09:29:38.317+0000 I CONTROL  [initandlisten] MongoDB starting : pid=26 port=27017 dbpath=/var/lib/mongodb/data 64-bit host=mongodb-0
2017-07-25T09:29:38.317+0000 I CONTROL  [initandlisten] db version v3.0.11
2017-07-25T09:29:38.317+0000 I CONTROL  [initandlisten] git version: 48f8b49dc30cc2485c6c1f3db31b723258fcbf39
2017-07-25T09:29:38.317+0000 I CONTROL  [initandlisten] build info: Linux c1bg.rdu2.centos.org 2.6.32-573.22.1.el6.x86_64 #1 SMP Wed Mar 23 03:35:39 UTC 2016 x86_64 BOOST_LIB_VERSION=1_53
2017-07-25T09:29:38.317+0000 I CONTROL  [initandlisten] allocator: tcmalloc
2017-07-25T09:29:38.317+0000 I CONTROL  [initandlisten] options: { config: "/etc/mongod.conf", net: { port: 27017 }, replication: { oplogSizeMB: 64, replSet: "rs0" }, security: { keyFile: "/var/lib/mongodb/keyfile" }, storage: { dbPath: "/var/lib/mongodb/data", mmapv1: { preallocDataFiles: false, smallFiles: true } }, systemLog: { quiet: true } }
2017-07-25T09:29:38.322+0000 W -        [initandlisten] Detected unclean shutdown - /var/lib/mongodb/data/mongod.lock is not empty.
2017-07-25T09:29:38.332+0000 I STORAGE  [initandlisten] ************** 
old lock file: /var/lib/mongodb/data/mongod.lock.  probably means unclean shutdown,
but there are no journal files to recover.
this is likely human error or filesystem corruption.
please make sure that your journal directory is mounted.
found 2 dbs.
see: http://dochub.mongodb.org/core/repair for more information
*************
2017-07-25T09:29:38.334+0000 I STORAGE  [initandlisten] exception in initAndListen: 12596 old lock file, terminating
2017-07-25T09:29:38.334+0000 I CONTROL  [initandlisten] dbexit:  rc: 100
`

omron93:
@dongboyan77 Thank you for the testing.

> after changing user name and password via env vars, still can login with old user name and password in new pod,

Are you changing only password in new pod?

If you are changing both `MONGODB_USER` and `MONGODB_PASSWORD` it is intentional that "old" users in database are preserved.

> petset-replicas, after removing all db pods, new pod cannot be running again, the expect result is all db pods are running again

Is this working with `centos/mongodb-30-upg-centos7` from Docker Hub?

How can I test it myself?

2.6/s2i/bin/run
bparees: yes it's related to #246, but I guess it's ok at least for now.  The entrypoint for the user-built image will be "container-entrypoint" still and the default command will be this run script, the user can change the command to "/bin/run-mongod-replication" if they want to use replication.

dongboyan77:
@omron93 ok, I change both MONGODB_USER and MONGODB_PASSWORD , so it's normal.
and test with centos/mongodb-30-upg-centos7 , the same issue.

I use this template https://github.com/sclorg/mongodb-container/blob/master/examples/petset/mongodb-petset-persistent.yaml

omron93:
Removed section about mounting custom configuration file.
[test-openshift]

> and test with centos/mongodb-30-upg-centos7 , the same issue.

OK. Also basic scaling was working for me with this template.

Then it is unrelated to this PR, so I've created new issue https://github.com/sclorg/mongodb-container/issues/249
@dongboyan77 Please take a look.

PR #243, opened 2017-05-23
Update Dockerfiles

omron93:
Use COPY instruction instead of ADD
(COPY is preffered for simple usage - see https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy)

Provide same labels for all base image variants (CentOS, RHEL, Fedora)
    - add name and version labels (recommended labels in Project Atomic Container best practices)
    - remove architecture label for RHEL based images (this label is set in RHEL image - no need to define it again)
 
(don't use separate instruction for each label)

@hhorak @praiskup @pkubatrh Please take a look and merge.

### DIFF
--- a/2.6/Dockerfile
+++ b/2.6/Dockerfile
@@ -27,7 +27,7 @@ RUN yum install -y centos-release-scl && \
     rpm -V $INSTALL_PKGS && \
     yum clean all
 
-ADD root /
+COPY root /
 
 
 # Container setup
--- a/2.6/Dockerfile
+++ b/2.6/Dockerfile
@@ -2,7 +2,10 @@ FROM centos:centos7
 LABEL MAINTAINER SoftwareCollections.org <sclorg@redhat.com>
 
 ENV SUMMARY="MongoDB NoSQL database server" \
-    DESCRIPTION="MongoDB is a scalable, high-performance, open source NoSQL database."
+    DESCRIPTION="MongoDB (from humongous) is a free and open-source \
+cross-platform document-oriented database program. Classified as a NoSQL \
+database program, MongoDB uses JSON-like documents with schemas. This \
+container image contains programs to run mongod server."
 
 LABEL summary="$SUMMARY" \
       description="$DESCRIPTION" \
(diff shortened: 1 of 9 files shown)


2.6/Dockerfile
```diff
@@ -1,10 +1,19 @@
 FROM centos:centos7
-MAINTAINER SoftwareCollections.org <sclorg@redhat.com>
+LABEL MAINTAINER SoftwareCollections.org <sclorg@redhat.com>
```
pkubatrh: I would put this under the same LABEL command as the other labels. Also use the key="value" format that is used for existing labels.

2.6/Dockerfile
```diff
@@ -1,10 +1,19 @@
...
 
-LABEL io.k8s.description="MongoDB is a scalable, high-performance, open source NoSQL database." \
+ENV SUMMARY="MongoDB NoSQL database server" \
+    DESCRIPTION="MongoDB is a scalable, high-performance, open source NoSQL database."
```
pkubatrh: The description could be a bit more verbose

2.6/Dockerfile
omron93: Thanks for review.

I saw this in Fedora Container Guidelines (https://fedoraproject.org/wiki/Container:Guidelines), so I thought that this is right way so I simply changed it in this way. Now I see that in same guidelines in different example there is also your proposed syntax :-)

@hhorak How to deal with specifying maintainers in sclorg/ images? Are there any internal guidelines about similar requirements?

2.6/Dockerfile
omron93: More verbose description added.

2.6/Dockerfile
hhorak: I don't think we have guidelines written anywhere, the practice is that we use it for Fedora and CentOS images (the mailing list), but not for RHEL images, but I don't see this written somewhere either. A quick look-up showed that lowercase `maintainer` seems preferred (https://github.com/projectatomic/ContainerApplicationGenericLabels/blob/master/vendor/redhat/labels.md and also mentioned once in the Fedora Guidelines)
and it would make sense to me to handle the `maintainer` label the same as other labels..

omron93: OK. Updated so. (other images will be too in near future)

hhorak:
Looks good to me, merging.

PR #247, opened 2017-07-13
fixed README.md for version 3.2 according to new rules

ficap:


hhorak:
One suggestion I have -- I think we should use RHEL-based image and also include link to the container catalog (it's possible to use direct links to the particular image), same as it is done in this MySQL's pull-request: https://github.com/sclorg/mysql-container/pull/188

Otherwise it looks good, please, create the same changes for 2.6, and 3.0-upg versions as well.

omron93:
@ficap Thank you. Merging.

PR #248, opened 2017-07-19
Remove the created containers during test properly

trepik:
After running tests on containers, not all of them were cleaned up properly. It was in the container creation tests and the containers that should have been stopped and removed were left running.
Changing the signal sent to containers by timeout to SIGTERM solved the issue.


omron93:
LGTM.

Thank you very much.

[test]

PR #250, opened 2017-07-27
tests: Use symlinks to point to files outside of versioned directories

pkubatrh:
This is needed for downstream sync as anything outside the versioned directories is unavailable (but can be pulled in if symlinked properly).

### DIFF
--- a/2.6/test/run
+++ b/2.6/test/run
@@ -492,7 +492,7 @@ function run_s2i_test() {
     ln -s mongod-mmapv1.conf ${test_dir}/test-image/mongodb-cfg/mongod.conf
 
     echo "  Testing s2i build"
-    s2i build file://${test_dir}/test-image/ ${IMAGE_NAME}-testapp
+    s2i build file://${test_dir}/test-image/ ${IMAGE_NAME} ${IMAGE_NAME}-testapp
 
     local container_name=s2i_config_build
     IMAGE_NAME=${IMAGE_NAME}-testapp _s2i_test_image "s2i_config_build" ""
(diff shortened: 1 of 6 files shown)


hhorak:
It looks like the test fails only in the end of the clean section, maybe this change made in the last commit might be related:
```
-    docker network ls | grep -q ${network_name} && docker network rm ${network_name}
+    docker network ls | grep -qv ${network_name} || docker network rm ${network_name}
```
IMO the combination of `docker network ls | grep -qv ${network_name}` can never end with exit code other than 0, since `docker network ls` returns the header and the grep will match at least the header.. The previous version makes better sense to me, although we might ignore errors at this point, so something like this might work:
```
set +e
docker network ls | grep -q ${network_name} && docker network rm ${network_name}
```
But it's possible I'm totally wrong and we see a totally different issue here.

hhorak:
@omron93 any ideas?

pkubatrh:
The fail seems to be caused by me removing a bit more from the code than I was supposed to from the s2i phase. I find it weird that it did not fail on rhel though. There must have been some leftover images from earlier tests

pkubatrh:
Tests are passing now

omron93:
```
-    docker network ls | grep -q ${network_name} && docker network rm ${network_name}
```

Problem is that if `${network_name}` is not there, `docker network ls | grep -q ${network_name} ` ends with `1` so second part is not run and whole `cleanup()` ends with `1` too.

@hhorak But you are right, that new solution isn't working right.
I've created https://github.com/sclorg/mongodb-container/pull/251 to fix it. Please take a look.

omron93:
LGTM, @pkubatrh thanks. Merging.

PR #251, opened 2017-07-28
Fix cleanup in tests

omron93:
Checking for presence of network used for replication isn't right
- `docker network rm` is never run now

Discovered in https://github.com/sclorg/mongodb-container/pull/250

@hhorak @pkubatrh What about this solution?


hhorak:
Makes sense to me, merging.

PR #253, opened 2017-08-08
Use s2i-core instead of s2i-base

pkubatrh:
Change to `Dockerfile.fedora` is not part of this PR as it seems the Fedora version of s2i-core has yet to be built.


hhorak:
Looks good to me, merging.

PR #255, opened 2017-08-15
MongoDB 3.4 + Sharding/ConfigServer/Security Functionality

jbornemann:
Lots of content here, I know; but the gist is that @ecwpz91 and I have added features to enable setting up a fully sharded, secured MongoDB cluster on OpenShift (or elsewhere?). The container can now be run to accommodate deploying a shard router, shard configuration server, or replica set member. In addition, members of replica sets have features to secure communications over TLS - including automatic certificate, and CSR generation.

@hhorak I saw the previous comment about using Red Hat builds for 3.4. Let me know how I can help get a 3.4 build up


jbornemann:
@rpitonak @hhorak @pkubatrh

omron93:
@jbornemann @ecwpz91 Thank you very much for these improvements. And sorry for replying too late.

Certainly this PR contains a lot of useful features and improvements.

Some points to this PR:
- we try to have all supported versions of this image as similar as possible to simplify maintenance
(for example support for source-to-image was added to this image, so this PR differs much)
- we uses rpms for installing SW. We have control over what we ship - we build mongo from sources and we can quickly react for possible urgent fixes. Also we pack the SW to comply with guidelines to better fit into RHEL/Fedora ecosystem

Also this PR is really big and it is very hard to review it. Would it be possible to split this PR into smaller ones? Can I somehow help?

jbornemann:
@omron93 thanks for taking a look!

Revisiting a bit, it looks like there is some coordination that would need to occur between this PR and your PR https://github.com/sclorg/mongodb-container/pull/256 

I saw recently that rh-mongodb-34 is in beta? 

It might be an easier PR to review if done commit by commit; I was careful to isolate features to commits.

Just a quick question - it looks like we are moving towards a "latest" folder instead of copying functionality version release to version release. I think this is a good step forward. Any thoughts on keeping a single folder, and tagging mongo releases instead? The latest folder will probably solve a lot of the code friction issues; but it seems like we could collapse the previous folders into it as well - we know the latest SHAs in each, and could tag them. Besides new features, it seems the Dockerfile build (specifying the rh-mongo image) is really the only thing that changes between mongo releases - Having a single dynamic Dockerfile that builds a mongo release knowing the tag name might make back-porting easier if it comes to that.

omron93:
> I saw recently that rh-mongodb-34 is in beta?

Right.

> It might be an easier PR to review if done commit by commit; I was careful to isolate features to commits.

That is true. But discussion about all features have to take place in one PR and will be mixed together.

> Any thoughts on keeping a single folder, and tagging mongo releases instead?

We are releasing builds/rebuilds of almost all versions (only 2.4 is EOL). And we need deliver fixes/updates in image scripts into all our build images. So I think tagging is not right approach (maybe I don't understand you proposal right).

>  Having a single dynamic Dockerfile that builds a mongo release knowing the tag name might make back-porting easier if it comes to that.

That is right. I plan to do similar PR as https://github.com/sclorg/mariadb-container/pull/44 for mongodb too.

We try to keep source accessible - we want to have sources easily available for each version (also we want to have it buildable using `docker build`). So we are keeping sources for each version in own directory - originally set up by OpenShift team developing docker images. Now we are considering using some kind of generator all versions from "one source", but it's still WIP. Overall we need same source structure in all sclorg/ images.

jbornemann:
@omron93 I have https://github.com/sclorg/mongodb-container/pull/264 and https://github.com/sclorg/mongodb-container/pull/263 split out along with others. It would make sense to review these two first, as the others build off of these. 

I may need some help shepherding some of this stuff - I won't have a whole lot of time this week to devote to this unfortunately. I didn't integrate the s2i features into the new code organization here. I wasn't quite sure the intent of using s2i in this context. Feel free to make changes, or add commits.

PR #256, opened 2017-08-28
Add rh-mongodb34

omron93:
Add Mongodb 3.4 from RH Software Collections.

This PR also add `latest` directory which always contains latest version. In this directory development of images is done and git history is preserved there after adding new image versions. To allow this please **DON'T SQUASH this PR** using github UI (at least two commits are required), thanks.

@hhorak What to do to make available rh-mongodb34* packages built in CBS?


pkubatrh:
It would be better to use 2.4 as the base for `latest`, since that is the version that will have most of the history (except some of the recent changes)

omron93:
It would be better to use 2.4 as the base for latest, since that is the version that will have most of the history (except some of the recent changes)

On the other hand a lot of changed since 3.2 was added (and changes was aiming to it). And I thing it is more that half of the code was somehow rewritten/restructured.
That is true that 2.4 contains more history, but latest changes in 3.2 after 2.4 deprecation are more useful in future (as mentioned abore the code is rewritten).

@pkubatrh So how to do to have changes from 2.4 and to keep also changes made to 3.2 after 2.4 deprecation?

pkubatrh:
>So how to do to have changes from 2.4 and to keep also changes made to 3.2 after 2.4 deprecation?

I guess you could apply all of the recent changes that went into `3.2` after `2.4` got deprecated on top of `2.4` as part of the PR but that would get real messy and most likely not worth the hassle...

I have nothing against keeping the current state (`latest` based on `3.2`) if you think the recent history is more valuable.

hhorak:
@omron93 as for the packages in cbs, I expect them to be on bulidlogs (testing repo today or tomorrow)

As for the question whether to use 2.4 or 3.2 as a base, my view is that I think we should try to keep history if possible, but if that would require a lot of work like applying patches one by one, I'd say let's just use 3.2.. the history is handy, but one can always figure out the necessary information by some git dancing..

omron93:
@pkubatrh @hhorak Thanks for reviews. So keeping `3.2` as a base for `latest`. 

I finally managed to rebase. So anything to do before merging?

omron93:
[test]

omron93:
[test]

omron93:
[test]

omron93:
[test]

omron93:
[test]

omron93:
Rebased.

One change: "workaround" for https://jira.mongodb.org/browse/SERVER-28072 has to be used for all versions because of shared tests. (I planned to do this later, but now I have to add changes to older versions to this PR)

pkubatrh:
tests seem to be failing due to:
```
Error: Invalid port number "//172.18.0.2,172.18.0.3/admin?replicaSet=rs0" in connection string "mongodb://172.18.0.2,172.18.0.3/admin?replicaSet=rs0/test" at src/mongo/shell/mongo.js:135
```

latest/Dockerfile.rhel7
```diff
@@ -37,7 +37,7 @@ CMD ["run-mongod"]
 RUN yum repolist > /dev/null && \
     yum install -y yum-utils && \
     yum-config-manager --disable \* &> /dev/null && \
-    yum-config-manager --enable rhel-server-rhscl-7-rpms && \
+    yum-config-manager --enable rhel-server-rhscl-beta-7-rpms && \
```
pkubatrh: rhel-server-rhscl-beta-7-rpms -> rhel-server-rhscl-7-beta-rpms

latest/Dockerfile.rhel7
```diff
@@ -13,7 +13,7 @@ LABEL summary="$SUMMARY" \
       io.openshift.expose-services="27017:mongodb" \
       io.openshift.tags="database,mongodb,rh-mongodb34" \
       com.redhat.component="rh-mongodb34-docker" \
-      name="rhscl/mongodb-34-rhel7" \
+      name="rhscl-beta/mongodb-34-rhel7" \
```
pkubatrh: We are using the regular `rhscl/` namespace this time around so no change needed here.

latest/Dockerfile.rhel7
omron93: Thanks, fixed.

latest/Dockerfile.rhel7
omron93: Thanks, fixed.

hhorak:
Tests pass, so I guess we can merge it.

PR #258, opened 2017-08-30
modify the tests to make use of new library in container-common-scripts

trepik:
The get_version function is just stupid one should probably find a better way to implement it

otherwise all tests passed for me ... localy

### DIFF
--- a/2.4/test/run
+++ b/2.4/test/run
@@ -11,85 +11,22 @@ shopt -s nullglob
 
 IMAGE_NAME=${IMAGE_NAME-openshift/mongodb-24-centos7-candidate}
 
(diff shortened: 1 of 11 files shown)


2.4/test/run
```diff
@@ -11,85 +11,22 @@ shopt -s nullglob
 
 IMAGE_NAME=${IMAGE_NAME-openshift/mongodb-24-centos7-candidate}
 
-CIDFILE_DIR=$(mktemp --suffix=mongodb_test_cidfiles -d)
+source ../common/test-lib.sh
+source ../test/lib
```
pkubatrh, line 15: Both of these need to become symlinks instead, so that we can make sure the tests do not fail even when run in isolation.

pkubatrh:
Will be continuing the PR in #259

PR #259, opened 2017-09-26
modify the tests to make use of new library in container-common-scripts

pkubatrh:
Continues #258 

I removed changes for `2.4` as the version is EOL already.
Modified the code so that the test can be run even when isolated inside the versioned directories (similarily to https://github.com/sclorg/postgresql-container/pull/186) and rebased to latest `common` to be able to use the common test library.


test/run
```diff
@@ -0,0 +1,132 @@
...
+. test/lib
+
+TEST_LIST="\
+ctest_container_creation
+ctest_configuration
+ctest_general
+ctest_change_password
+ctest_mount_config
+ctest_doc_content
+ctest_WT_cache
```
omron93: This should/was run only for 3.2 (and newer)

test/run
```diff
@@ -0,0 +1,132 @@
...
+    local user='user'
+    local password='password'
+    local admin_password='adminPassword'
+
+    local volume_dir
+    volume_dir=`mktemp -d --tmpdir mongodb-testdata.XXXXX`
+    chmod a+rwx ${volume_dir}
+    config_file=$volume_dir/mongod.conf
+    echo "dbpath=/var/lib/mongodb/dbpath
+unixSocketPrefix = /var/lib/mongodb" > $config_file
```
omron93: Please add also

    smallfiles = true
    noprealloc = true

into `config_file`. This is for `2.6`-`3.0-upg` which don't use WiredTiger storage engine. For `3.2` it is not problem, maybe only some warning about having options for old storage engine is printed.

test/run
```diff
@@ -0,0 +1,132 @@
...
+    mongo_cmd "db.testData.find().forEach(printjson);"
+    mongo_cmd "db.testData.count();"
+    mongo_cmd "db.testData.drop();"
+    mongo_cmd "db.dropDatabase();"
+    echo "  Success!"
+}
+
+function ctest_configuration() {
+    echo "  Testing image configuration settings"
+    test_config_option MONGODB_QUIET true "quiet: true"
```
omron93, line 66: Also

    test_config_option MONGODB_PREALLOC false "preallocDataFiles: false"
    test_config_option MONGODB_NOPREALLOC true "preallocDataFiles: false"
    test_config_option MONGODB_SMALLFILES true "smallFiles: true"

was tested for `2.6`-`3.0-upg`

test/run
omron93: Also this is why CI failed...

test/ctest_local_replication
```diff
@@ -0,0 +1,67 @@
+#
+# Local replication test for MongoDB image.
+#
+# Always use sourced from a specific container testfile
+
+function ctest_local_replication() {
```
omron93, line 6: `cleanup` function also contained 

```bash
local network_name="mongodb-replset-$$"
! docker network ls | grep -q ${network_name} || docker network rm ${network_name}
```

to remove network created for this test.

So I suggest to add something like this into this function
```bash
function cleanup() {
  ct_cleanup()
  local network_name="mongodb-replset-$$"
  ! docker network ls | grep -q ${network_name} || docker network rm ${network_name}
}
trap cleanup EXIT SIGINT
```

and to the end of this function add `ct_enable_cleanup`

omron93 (changes requested):
@pkubatrh Thanks for working on this. I hope tests will be more readable and easier to maintain.

I found some some differences which was in tests.

pkubatrh:
Right, I have blindly copied in the tests for the 3.2 version of mongodb. Pointed out differences should be added now, thanks!

test/run
```diff
@@ -0,0 +1,152 @@
...
+    chmod a+rwx ${volume_dir}
+    config_file=$volume_dir/mongod.conf
+    config_content="dbpath=/var/lib/mongodb/dbpath
+unixSocketPrefix = /var/lib/mongodb"
+    if test $VERSION_NUM -le 30; then
+    config_content="$config_content
+smallfiles = true
+noprealloc = true"
+    fi
+    echo $config_content > $config_file
```
omron93: `"` is missing (`echo "$config_content"`). Without it all content is in one line.

pkubatrh: Fixed, thanks!

test/run
```diff
@@ -0,0 +1,158 @@
...
+}
+
+function ctest_s2i() {
+    echo "  Testing s2i usage"
+    s2i usage ${S2I_ARGS} ${IMAGE_NAME} &>/dev/null
+
+    # Set configuration file (differs for mmapv1 storage engine)
+    mv ${TEST_DIR}/test-image/mongodb-cfg/mongod.conf ${TEST_DIR}/test-image/mongodb-cfg/mongod.conf.backup
+
+    if test $VERSION_NUM -ge 32; then
```
omron93, line 134: @pkubatrh Thanks for fixing this problem.

Tests reports: `test/run: line 137: test_dir: unbound variable`

pkubatrh:
rhel tests are green now

omron93:
[test]

omron93:
[test] again. CentOS CI should be working now. (there are some changes for users, I'll send email)

omron93 (approved):

pkubatrh:
Thanks for the review! Both tests are passing so merging.

PR #264, opened 2017-10-09
Configuration Server Run-Mode functionality

jbornemann:
This PR accomplishes two things:

1.) Setting up init folder structure to a more domain-driven design - this will help with segregating/sharing code across future run modes. (coming in other PRs). I've looked through both new mongodb/db structure, and init/pre-init structure to make sure nothing was missed, and it seems to be a mostly organizational change. The exception seems to be s2i - which I was not sure of the intent of this, so I didn't attempt to incorporate it. I might need someone with the backstory to add the commit adding that functionality. p.s why extend the image with s2i as opposed to just a plain docker build? I guess I would expect the tool to be used more exclusively in quick turn-around type uses; as opposed to extension of a middleware image - where I would expect iterations to be much more well defined, and steady. 

2.) Added run-configsvr

Just a note, https://github.com/jbornemann/mongodb-on-os I plan on submitting to container-quickstarts as an example of how all of these run modes fit together on OpenShift in a sharded Mongo environment.


jbornemann:
@omron93 this is probably a good first PR to start on, in terms of splitting up https://github.com/sclorg/mongodb-container/pull/255

latest/root/usr/share/container-scripts/mongodb/environment
```diff
@@ -0,0 +1,78 @@
...
+#----------------------------------------------------
+
+# PYTHON is the full path of the python command, explicitly set to guard against
+# incautious use of shebang or interpreter/program invokation.
+readonly PYTHON=/usr/bin/python
+
+# MONGO is the full path of the interactive JavaScript shell interface to
+# MongoDB, explicitly set to guard against incautious use of shebang or
+# interpreter/program invokation.
+readonly MONGO=/usr/bin/mongo
```
omron93, line 13: These images uses mongodb from RHSCL, so this is not right. And in general can't be used because each SCL has binaries in own directory. So `scl enable COLLECTION` is used to set up paths,...

latest/root/usr/share/container-scripts/mongodb/environment
```diff
@@ -0,0 +1,78 @@
...
+# explicitly set to guard against incautious use of shebang or
+# interpreter/program invokation.
+readonly MONGOD=/usr/bin/mongod
+
+#----------------------------------------------------
+# Directory PATHs
+#----------------------------------------------------
+
+# MONGODB_DATADIR is the location of the database files live for mongo runtime
+export MONGODB_DATADIR=$HOME/data
```
omron93, line 25: This is default path for Upstream MongoDB.

1. This is really nonstandard location according to FHS
2. Our images uses `/var/lib/mongodb/data` for long time and I think it won't be easily to change this even we have strong reasons for this. Users have this directory in their configurations, templates, ...

Also `MONGODB_LOG_PATH` and `MONGODB_PID_PATH` uses nonstandard location according to FHS.

latest/root/usr/share/container-scripts/mongodb/generate_container_user
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+# Set current user in nss_wrapper
+USER_ID=$(id -u)
+GROUP_ID=$(id -g)
+
+if [ x"$USER_ID" != x"0" -a x"$USER_ID" != x"1001" ]; then
```
omron93, line 7: Why is nsswrapper needed?

latest/Dockerfile.rhel7
```diff
@@ -30,7 +30,7 @@ ENV MONGODB_VERSION=3.4 \
 EXPOSE 27017
 
 ENTRYPOINT ["container-entrypoint"]
-CMD ["run-mongod"]
+CMD ["run-standalone"]
```
omron93, line 33: I remember that we had to keep `run-mongod-replset` as symlink to `run-mongod-replication` after renaming. Because users have this command already in the templates, ...
So not sure if we don't have to do the same for `run-monod` if we decide to rename it.
@jbornemann Any background story/rasoning behind your naming scheme? (`run-standalone`, `run-configsrv`, `run-replset`,...)

We chose `run-mongod`, `run-mongod-replication` and I can imagine `run-mongod-configsrv`, `run-mongos`,...

mariadb-container, mysql-container and postgresql-container uses same pattern `run-DAEMON[-ROLE]`. So `run-mysqld`, `run-mysqld-master` and `run-mysqld-slave`.

latest/root/usr/bin/run-configsvr
```diff
@@ -0,0 +1,46 @@
...
+log_info "Setting up config file"
+if [[ ! -s "$MONGODB_CONFIG_PATH" ]]; then
+	envsubst < "${CONTAINER_SCRIPTS_PATH:-}/mongodb.conf.template" > "$MONGODB_CONFIG_PATH"
+fi
+
+log_info "Settings up keys"
+setup_keyfile
+
+log_info "Starting mongo configuration server"
+$MONGOD --config "$MONGODB_CONFIG_PATH" --configsvr --keyFile $MONGODB_KEYFILE_PATH --replSet "${MONGODB_REPLICA_NAME:-}" &
```
omron93, line 37: `--configsvr` option is useless in our containers anyway. 

1. we change database location anyway
2. port does not matter in containers (no conflicting)

omron93 (commented):
Thanks for this PR. There are some new features which are good. But also there is a lot of changes (naming, structuring,...) which was done probably by writing code independently. It would be good to have reason for these changes - useless rewriting of code can bring only hidden bugs ;-)

Also it would be great to catch latest changes about s2i support. So how it works:
(https://github.com/sclorg/mongodb-container/tree/master/latest#extending-image)
1. user can submit own config file - not influencing this PR
2. running mongod (standalone) consists of these phases:
a) sourcing pre-init scripts - in new container, nothing is running,... (so checking variables, setting cache size, config file,...)
b) starting mongod in background with disable authentication
c) sourcing init scripts - accessing mongod without auth  (creating users,...)
d) stoping mongod and execing it with auth enabled
3. running mongod (replicaset) consists of these phases:
a) sourcing pre-init scripts - same as 2.a)
b) starting mongod in background with enabled auth and with --replSet set 
c) setting up replicaset - initialiting mongodb config or adding server into mongodb replset
d) sourcing init scripts - replset is running and working (creating users,...)

Scripts used in 2.a)c) and 3.a)d) have specified location. So using `s2i build` or docker mount user can provide own scripts executed in this phase (or overwrite default one). Also our scripts can be divided into smaller peaces - which makes maintained easier, can share scripts among 2. and 3. and so on

So as I said it would be great to use same pattern (if you are not strongly against it).

----------------------------------------------------
Also I have idea about config server. Basically it is replicaset - with same initialization process as `run-mongod-replication`. So IMPOV code coud be shared. What about having `run-mongod-configsvr` (or whatever name):
```bash
#!/bin/bash

set -o errexit
set -o nounset
set -o pipefail

export mongo_common_args=" --configsvr"

export MONGODB_REPLICA_NAME=${MONGODB_REPLICA_NAME:-rsConfigsvr}
export MONGODB_SERVICE_NAME=${MONGODB_SERVICE_NAME:-mongodb-configsvr}

exec run-mongod-replication
```
And rest of the code share. Ideas?

----------------------------------------------------
You added function about removing from replicaset - `rs_remove()`. I can't find discussion about this in old PRs now :-( But result was not remove dead members from mongodb replicaset config and to leave this to administrator. If someone would use it, I'm not against it.

----------------------------------------------------
@jbornemann I am ready to prepare PR for whatever I suggest, feel free to ask. (I don't want duplicate work)

latest/root/usr/share/container-scripts/mongodb/environment
jbornemann: Ah okay, good catch. This was put in pre-RPM install.

latest/root/usr/share/container-scripts/mongodb/environment
jbornemann: Agreed. This should be changed.

latest/root/usr/share/container-scripts/mongodb/generate_container_user
jbornemann: I'm going to delegate to @msurbey for this one. I imagine it is to circumvent umask issues, without introducing additional SCC; but he put this in, so he might have a better answer.

latest/Dockerfile.rhel7
jbornemann: Reasoning was to keep intent clear - almost all of the modes are running mongod, but *how* are they running mongod? Keeping consistent with the mariadb scheme is probably a good idea.

latest/root/usr/bin/run-configsvr
jbornemann: Not sure I understand all of this comment. For number 2, correct, we don't have to worry about ports conflicting. I explicitly set the default 27017 in the configuration for this reason - because config server assumes running on the same host, and runs on a different port by default (27019)

jbornemann:
> There are some new features which are good. But also there is a lot of changes (naming, structuring,...) which was done probably by writing code independently. It would be good to have reason for these changes - useless rewriting of code can bring only hidden bugs ;-)

The intent was to structure the code in a way such that functions are shard, or segregated based on run-mode - to easily find functions, and reduce duplication. With the introduction of new run-modes, code organization problems are uncovered that weren't previously there. 

> s2i

No strong feelings about having vs not having it. Personally, if I were a user wanting to extend the image; I would just extend it via Dockerfile, add scripts, and change the entrypoint - but that is just personal taste. To me, s2i is more of a tool for deploying applications on an existing image within a mostly pre-defined lifecycle (assemble, save-artifacts, run) - although, yes, you can extend it. 

> Also I have idea about config server

Sure, sounds good to me. Those two things seem to be likely to change together, so probably does make sense to consolidate. 

> rs_remove()

I delegate to @msurbey again for this one. No strong feelings.

omron93:
[test]

jbornemann:
@omron93 just FYI, if it suites your fancy - feel free to help move these PRs forward. I likely won't have a whole lot of dedicated time to this.

ghost:

hhorak:

PR #275, opened 2018-01-12
Use shared functions from common submodule also in replication tests

omron93:
Replication tests are currently broken. I hope this will fix them.


omron93:
[test-openshift]

omron93:
[test-openshift]

omron93:
[test-openshift]

pkubatrh:
Since the tests have a `print_logs` function it would be a good idea to call it before returning a failure.

The function is also something that would be nice to have in common scripts as I expect all of the repositories would benefit from it.

omron93:
> Since the tests have a print_logs function it would be a good idea to call it before returning a failure.

@pkubatrh What is the "most elegant" way how to do it?
There was set `trap print_logs ERR` but `set -e` ends it before logs are printed. On the other hand it's useless and annoying to print logs if all tests pass.

pkubatrh:
Calling it explicitly right before `return 1` would work.

Best thing would be to actually call it even when all tests pass - the project should get deleted on a pass, so nothing is going to get printed out.

pkubatrh:
>the project should get deleted on a pass, so nothing is going to get printed out.

Although it seems that the project instance is still available for a few moments after it gets flagged for removal... So maybe also add a smal sleep at the end of the test suite to sync up.

omron93:
[test-openshift]

Now includes changes from https://github.com/sclorg/container-common-scripts/pull/51 . Will remove definition o functions after the PR is merged.

omron93:
[test-openshift]

omron93:
[test-openshift]

(local functions are removed)

omron93:
[test-openshift]

omron93:
[test-openshift]

omron93:
[test-openshift]

omron93:
[test-openshift]

omron93:
[test-openshift]

omron93:
[test-openshift]

omron93:
[test-openshift]

omron93:
Tests are finally passing. So merging.

PR #278, opened 2018-02-28
Support mongos, configsvr, and shardsvr

jkupferer:
This is work based off of jbornemann's work documented here:

https://github.com/redhat-cop/containers-quickstarts/tree/master/mongodb

and:

https://github.com/jbornemann/mongodb-container/tree/working_branch/3.2

The purpose of these updates to support sharding with mongodb using mongos with a configsvr. I can provide more documentation as needed, but before I do I want to be sure there is interest in these features being incorporated into this mongodb image.

Thanks in advance

### DIFF
--- a/latest/root/usr/share/container-scripts/mongodb/init-configsvr.sh
+++ b/latest/root/usr/share/container-scripts/mongodb/init-configsvr.sh
@@ -21,11 +21,7 @@ readonly MEMBER_HOST="$(hostname -f)"
 function initiate() {
   local host="$1"
 
-  local config="{_id: '${MONGODB_REPLICA_NAME}', members: [{_id: 0, host: '${host}'}]}"
-
-  if [[ "$MONGODB_MODE" == 'configsvr' ]]; then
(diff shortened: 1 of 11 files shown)


latest/root/usr/share/container-scripts/mongodb/init-configsvr.sh
```diff
@@ -0,0 +1,73 @@
...
+# Arguments:
+# - $1: host address[:port]
+#
+# Uses the following global variables:
+# - MONGODB_REPLICA_NAME
+# - MONGODB_ADMIN_PASSWORD
+function initiate() {
+  local host="$1"
+
+  local config="{_id: '${MONGODB_REPLICA_NAME}', configsvr: true, members: [{_id: 0, host: '${host}'}]}"
```
omron93, line 24: Is this file needed? I see that `latest/root/usr/share/container-scripts/mongodb/init-replset.sh` can also set `configsvr: true`...

omron93 (commented):
I'm very sorry that it took me so long to get into this PR.
@jkupferer This PR look really good. So it would be definitely great to have it merged. Documentation is welcomed - is your offer still valid?

ghost:

hhorak:

PR #281, opened 2018-03-12
Rename -init hook to -start hook

omron93:
SSIA

@praiskup @pkubatrh @ficap Did I miss something?


praiskup:
It seems to match the postgresql-container case, so looks OK to me.

omron93:
@praiskup We decided that -init is run before -start? I can't remember...

2.6/root/usr/share/container-scripts/mongodb/README.md
```diff
@@ -179,6 +179,11 @@ started
 these scripts are skipped if `run-mongod-replication` is run with already
 initialized data directory
 
+##### `mongodb-start/`
+
+same sematics as `mongodb-init/`, except that these scripts are always source
```
praiskup: sourced

praiskup:
I don't know whether we made some final decision, @ficap did the research, .. but
if `-init` is supposed to run only once, and `-start` always -- it would be nice to have
consistent feeling in `-start` hook (users can expect that the database is initialized).

Btw., I would prefer double-quotes around all the shell variables in the added code
at least.

2.6/root/usr/share/container-scripts/mongodb/README.md
omron93: thanks

omron93:
> I don't know whether we made some final decision, @ficap did the research, .. but
if -init is supposed to run only once, and -start always -- it would be nice to have
consistent feeling in -start hook (users can expect that the database is initialized).

+1

> Btw., I would prefer double-quotes around all the shell variables in the added code
at least.

Where it's missing?

omron93:
Rebased.

[test-openshift]

omron93:
[test-openshift]

omron93:
Rebased to include fix for Fedora images.
[test-openshift]

omron93:
[test-openshift]

omron93:
Tests are passing. No big changes after last review (rebases mainly). So merging.

PR #300, opened 2018-10-24
Fix Fedora image

omron93:
Remove release label - in Fedora it's updated automatically by build system.

Select python binary on runtime - Fedora images don't ship python
command by default

Remove arch label - 'Optional: if omitted, it will be built for
all supported Fedora Architectures'

@pkubatrh Please take a look and merge.


omron93:
[test-openshift]

pkubatrh:
lgtm thanks!

Issue #203, opened 2016-10-14
how to make mongod accept only ssl connections...?


SimonVuong:
i was able to use my on mongod.conf to specify ssl connections only

```
# mongodb.conf

port = 27017
pidfilepath = /var/lib/mongodb/mongodb.pid

# Set this value to designate a directory for the mongod instance to store its data.
# Default: /var/lib/mongodb/data
dbpath = /var/lib/mongodb/data

# Disable data file preallocation. Default: true
# only for mounted data directory from older MongoDB server
noprealloc = true

# Set MongoDB to use a smaller default data file size. Default: true
# only for mounted data directory from older MongoDB server
smallfiles = true

# Runs MongoDB in a quiet mode that attempts to limit the amount of output.
# Default: true
quiet = true

# Disable the HTTP interface (Defaults to localhost:28017).
nohttpinterface = true

#force connections to use TLS
sslMode = requireSSL

#specify the certificate
sslPEMKeyFile = /etc/mongodb.pem
```

but the problem is that the first mongo connection used to specify the mongo users fails because it doesn't try to connect with ssl.

From run-mongod

```
if [ "$(mongo admin --quiet --eval "$js_command")" == "1" ]; then
  echo "=> Admin user is already created. Resetting password ..."
  mongo_reset_admin
else
  mongo_create_admin
fi
```

any ideas on how we should account for this WITHOUT modifying common.sh and run-mongod??

madvimer:
Any updates on this. we have a requirement to implement ssl. Only option i see right now is to make changes to common.sh and run-mongod. Can this be achieved as asked ?

omron93:
@madvimer I hope this issue could be solved by using s2i to extend the image. So waiting to https://github.com/sclorg/mongodb-container/pull/287 to be merged. You can test the code from that PR if that works for you.

@UncleAlbie Can you estimate when you will be ready to review the PR?

panovotn:
@omron93 I'll review #287 by tomorrow afternoon.

omron93:
#287 is merged and changes are available in `docker.io/centos/mongodb-36-centos7` image already.

You can solve this issue by having folders like this:
```
myapp/
└── mongodb-ssl
    ├── ca.pem
    └── mongodb.pem
```

[...]

PR #301, opened 2018-10-24
Add requirements on SSL certificate into README

omron93:
Discussed in #203 

> I understand that eval was removed and replaced with "-host localhost" option. MongoDB docs specify that with SSL enabled it becomes mandatory to specify hostname( localhost will not work). Is my understanding correct on this or am i missing something

Thanks, you are right. I've missed it first.

I'm thinking about it and I see no other option than having to specify `localhost` in SAN of SSL certificate.
Basically in the container 'localhost' and 'hostname` can be resolved. And the hostname often changes with every container start. Also IMHO using real server/certificate domain for "internal" connection isn't good idea (although it could be possible if set up properly). This would be also complicated in openshift because connection ability from outside of the cluster needs to be configured explicitly.

--------------------------------------------------------------------------------
So until better solution is found, it would be better to have this notice in README.

@pkubatrh Please review and merge. I would be great if it would be possible to have it merged before next rebuild.

### DIFF
--- a/3.2/root/usr/share/container-scripts/mongodb/README.md
+++ b/3.2/root/usr/share/container-scripts/mongodb/README.md
@@ -156,6 +156,8 @@ when running `run-mongod` or `run-mongod-replication` commands contained
 
 SSL/TLS certificates used to configure MongoDB server SSL/TLS support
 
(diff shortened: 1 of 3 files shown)


pkubatrh:
lgtm, thanks!

PR #305, opened 2019-04-05
Added count_endpoint to determine the endpoints.

lehmeyer:
When the first replica set is deleted, the replica set always initializes itself again, even if a set already exists. By querying the number of endpoints, the node adds itself to an existing replica set.


ghost:

hhorak:

PR #307, opened 2019-05-15
template: do not point to a specific image version in description

pkubatrh:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1709251


pkubatrh:
Affects docs only, merging.

PR #314, opened 2019-08-23
changes stemming from testing this on OpenShift v4.1

kstephen314159:
The included changes fall into two buckets:
1.  Technical changes such as getting rid of alpha / beta annotations and such stuff from the days when petsets were still new to k8s. There was one situation where it wasn't sufficient to remove the annotation, but I had to find the equivalent "GA" version that was introduced as an attribute
2.  Documentation / re-organization changes where I renamed the "examples/petset" directory to "examples/statefulset - which then meant I had to edit a lot of places where that file path was referenced.


ghost:

hhorak:

PR #327, opened 2022-04-19
setfulset version update apps/v1

cybshark:
update version beta to apps/v1


hhorak:

